PR #2, opened 2015-01-02
Fixed README

gabipurcaru:
I've renamed README to README.md (since it's actually 99% in markdown syntax) and also added markdown-style links.


martanne:
Thanks for your contribution! Do you happen to know whether it is possible to instruct github to interpret the README as Markdown without changing the file extension?

In any case I will probably apply your changes later on.

gabipurcaru:
I don't think so; see this for more details: https://github.com/github/markup/issues/3. A proposed workaround is to symlink README.md to README

twe4ked:
If you want to keep the references at the bottom you can use the following syntax:

```
It all started when I was recently reading the excellent [Project Oberon][0],

[0]: http://www.inf.ethz.ch/personal/wirth/ProjectOberon/
```

If you [view the rendered README](https://github.com/gabipurcaru/vis/blob/f581025ded87f05ee733e898024632cf6560aeb6/README.md) there are a few places that need a bit of work. A few things under the _vis a vim like frontend_ section stick out to me. Tables are supported in [GitHub Flavoured Markdown](https://help.github.com/articles/github-flavored-markdown/) which might work well for the _Movements_ section and would still be readable in plain text. If that seems to hard to maintain perhaps they could be code blocks.

On a side note, you can also use the `.markdown` extension which Gruber [recommends](http://daringfireball.net/linked/2014/01/08/markdown-extension).

martanne:
I've now updated the README and converted it to Markdown (could probably still be improved).

PR #5, opened 2015-01-02
Changes to support compilation on OSX

fiveminuteargument:
Should be totally safe, but only able to test on OSX.


connerbrooks:
:+1:

martanne:
Thanks! I pushed a change which should also support a couple of other *nix systems.

Hopefully I didn't screw up the OS X related part.

fiveminuteargument:
Great - your way looks much nicer! Thanks for the feedback.

PR #6, opened 2015-01-03
Add python syntax support

matiaslina:
Nothing special here. Just a few rules :)

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -1115,6 +1115,45 @@ static Syntax syntaxes[] = {{
 		&colors[COLOR_SYNTAX6],
 	}}
 },{
+	.name = "python",
+	.file = "\\.py$",
+	.rules = {{
+		"(#.*$|#$)",
+		&colors[COLOR_COMMENT],
+	},{
+		"(\"\"\".*\"\"\")",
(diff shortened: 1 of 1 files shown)


martanne:
Thanks, I added them. However the python syntax with its symmetric multiline comment markers really shows the limitations of the simple regex based approach.

matiaslina:
> really shows the limitations of the simple regex based approach.

Could you elaborate this?

martanne:
Since """ is used both as the start and end marker of a multiline comment you can't easily distinguish the two without further context. If you scroll through a file until the beginning of a comment is no longer visible, the corresponding end marker will now be interpreted as the start of another comment, thus wrongly coloring the whole code.

matiaslina:
Yes, I have seen that glitch in some C code too with the comments. Thanks for explain. 

I'll dig into it to see how we can improve the situation. This will happen to any multiline regexp around there, except for complicated regexp like the C comments.

martanne:
I would not spend much time on it right now, it will require quite some design work.

As for the patches, I don't care. You can submit them here, send them by email directly to me or the suckless list (if you don't mind an occasional flame war ...)

PR #7, opened 2015-01-03
Relative line jumps

LemonBoy:


martanne:
Thanks! I implemented it in a slightly different way. Please report back whether this works for you.

PR #11, opened 2015-01-04
Fix openline command on the first line

mbohuslavek:
When on the first line, openline command wouldn't move the cursor to the
newly created line above the current line.


martanne:
Thanks, I applied it with MOVE_LINE_BEGIN instead of MOVE_LINE_START which correctly handles leading white spaces.

mbohuslavek:
Yes, I missed that. Thank you.

PR #13, opened 2015-01-05
define underscores to be part of a word

MatzeB:
Accept _ as part of a word. Was this left out on purpose before?


martanne:
No it wasn't on purpose, thanks.

PR #33, opened 2015-01-21
Cindent

MatzeB:
Putting this up for an early review (and I also fear that I won't find the time to work on this any more in the next weeks). This imitates vims cindent mode without having any options. What it does do:
- Add a hook to reindent the current line on CTRL+F
- Add the = command to reindent a visual block
- If the previous line ended in { indent
- Lines starting with # or which just contain spaces (but are not newly created) are not indentet
- Lines beginning with } are indented like the line that contained the matching {
- Lines ending in : (labels, case labels) are indentet like the line with the opening { of the current block (this is usually 1 indentation level less)
- Comment rules: If the previous line begins with "/*" or " *" then let the current line begin with " *" (for newly created lines)

This already does a fairly good job in many situations but it lacks some things that vim can do:
- Indent if the previous line started with "if", "while", "do" (for the cases when no {} block was used)
- Indent when the previous line did not end in ;, { or }  -  for broken long lines
- Maybe add a mode to indent unclosed ( ) braces to the level where the brace opened.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 include config.mk
 
-SRC = editor.c window.c text.c text-motions.c text-objects.c register.c buffer.c ring-buffer.c
+SRC = editor.c window.c text.c text-motions.c text-objects.c register.c buffer.c ring-buffer.c indent.c
 HDR := ${SRC:.c=.h} macro.h syntax.h util.h config.def.h
 SRC += vis.c
 OBJ = ${SRC:.c=.o}
--- a/config.def.h
+++ b/config.def.h
@@ -557,6 +557,7 @@ static KeyBinding vis_mode_insert[] = {
 	{ { CONTROL('X'), CONTROL('Y') }, wslide,       { .i = +1                } },
 	{ { NONE('\t')              }, insert_tab,      { NULL                   } },
 	{ { KEY(END)                }, movement,        { .i = MOVE_LINE_END     } },
+	{ { CONTROL('F'),           }, reindent_line,   { NULL                   } },
 	{ /* empty last element, array terminator */                               },
 };
 
--- a/config.def.h
+++ b/config.def.h
@@ -218,6 +218,7 @@ static KeyBinding vis_operators[] = {
 	{ { NONE('p')               }, put,           { .i = +1              } },
 	{ { NONE('P')               }, put,           { .i = -1              } },
 	{ { NONE('>')               }, operator,      { .i = OP_SHIFT_RIGHT  } },
+	{ { NONE('=')               }, operator,      { .i = OP_REINDENT     } },
 	{ { NONE('<')               }, operator,      { .i = OP_SHIFT_LEFT   } },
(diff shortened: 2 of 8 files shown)


matiaslina:
> Putting this up for an early review

will try it :) use cindent every day.

I don't think that `text_bracket_match_`  on  text-motions.c it's a good name for that function. Maybe you can change it for something more explicit, like `text_bracket_match_find`. (just an opinion)

> (and I also fear that I won't find the time to work on this any more in the next weeks). 

If you have some problems just tell, I can work on it on my free time. ;)

martanne:
I'm currently also busy with upcoming exams, thus a few comments regarding the implementation just from looking at the diff.

While I personally agree that tabs should be used for indention purposes, simply inserting a `\t` character breaks the expand tab functionality and therefore files which use spaces. See also the `expand_tab` function from vis.c.

`op_reindent` probably won't work correctly as is. The problem is that once you modify lines by adding indention you "invalidate" the range passed through `OperatorContext`. The end of range is now off by the amount of bytes you inserted for indention purposes. This is why the shifting operators work backwards i.e. they start at the end of the range and work towards the start which remains constant and isn't modified by the insertions in between. The same technique is probably not applicable here, since code should be indented based on the preceding context.

I don't like the introduction of `text_bracket_match_`, if I'm not mistaken it should be possible to achieve what you want with something like:

    Filerange block = text_object_curly_bracket(txt, pos);
    if (text_range_valid(&block))
       /* do something based on block.start */

Instead of testing for individual characters one could probably use  `text_find_char_{next,prev}`  (yes these are terribly misnamed). They should be changed to accept a NUL terminated string. Furthermore new versions which stop once a line break (`\r` or `\n`) is encountered should be introduced. These could then be used to check whether lines start with if/while/do etc:

    size_t begin = text_line_begin(text, pos);
    size_t match = text_line_string_next(text, begin, "while");
    if (match == text_line_start(text, begin))
        /* do something */

In order to detect case labels line based regex are probably necessary, not sure about the performance consequences.

On a more general level I'm not sure whether implementing the reindention operator internally in the editor is such a good idea. Piping the relevant text through an external program is also worth considering.

rychipman:
I would second Marc's aversion to making this a part of editor core functionality. This kind of functionality is undeniably useful and ubiquitous across editors, but merging it into the editor core is how feature bloat happens.

In my opinion, the lean, well-organized codebase (as well as the scrutiny and conservatism with which pull requests & feature requests are handled) is vis' killer feature. I would much rather all of these sorts of features be added as 'plugin-style' external programs using an rpc api.

sshbio:
So let's write some `cindent.lua` (and perhaps `lispindent.lua`) instead?

mcepl:
@MatzeB would it be possible to rebase this PR on the top of the current master, please? I would love to merge this into my fork https://git.sr.ht/~mcepl/vis , but this is too complicated for me to do it right. Thank you.

PR #37, opened 2015-03-30
[WIP] Add additional key mappings for window switching (^w+h, ^w+l, ^w^h, ^w^l)

twe4ked:
I can't actually get the `^w^h` mapping to work, all the others I added work fine and I can't see anything obvious in `config.def.h` that might be conflicting. I've never written any C before so unfortunately I'm not sure where to go from here to work it out. There doesn't seem to be any horizontal/vertical window movement at the moment so I just used the next/prev to match the j/k movements.

Love the premise of this project. Keep it up :)


twe4ked:
Did the `^w^h` mapping work for you? I marked it as WIP because I didn't know what was going on and I wanted feedback. :)

martanne:
No it didn't work. The short answer is that terminal input is a mess,  `^h` gets translated to backspace. Should be fixed by d95f7600.

twe4ked:
Ah of course. I've pushed a few small feature to branches on [my fork](https://github.com/twe4ked/vis), are you interested in more noob PRs? I'm just doing it for fun, your code is so easy to read.

martanne:
Thanks! In general I'm always interested in what other people do with the code. A few lines motivating the changes are always nice. As an example, what is your use case for the `cquit` command? I assume you use the editor in something like `vis file && $CMD`?

I don't think displaying the syntax in the status bar is really useful, it probably only wastes space. However what seems useful is that `:set syntax` would display the currently used syntax definition (if any) or else a list of known names.

A version flag will be provided once there is an official release.

twe4ked:
> As an example, what is your use case for the cquit command?

I generally use it in Vim to abort a Git commit.

> I don't think displaying the syntax in the status bar is really useful

Fair enough. Your suggestion sounds good.

> A version flag will be provided once there is an official release.

Makes sense. I thought it would be a few line change but ended up running into problems.

PR #40, opened 2015-04-09
Display info when setting syntax

twe4ked:
As discussed in #37.


martanne:
Thanks,  what I had in mind was more like 5123bbc4 i.e. that `:set syntax` would display the currently active syntax definition.

twe4ked:
Ah, I misunderstood.

PR #41, opened 2015-04-12
Add javascript syntax highlighting

trengrj:
This commit adds javascript and json syntax highlighting. I know it would complicate things but maybe splitting out syntax highlighting from config.def.h would be a good idea.


martanne:
Yes I tend to agree, but I'm not yet sure what the best way to do so is. Suggestions welcome ...

PR #46, opened 2015-04-23
Use #pragma once rather than include guards

torstehu:


martanne:
Thanks, however I prefer the include guards. #pragma is not standard (albeit admittedly supported by most compilers). Compile time shouldn't be an issue for a small project like vis and furthermore modern compilers have build in support to detect include guards.

The only advantage #pragma has is that is slightly less code and there is no danger of forgetting to update the include guards when a file is renamed (as was recently forgotten in view.h).

torstehu:
I understand, thanks for the quick and thorough response.

Issue #42, opened 2015-04-14
Cursor position issues


rychipman:
When changing from insert to normal mode, the cursor doesn't move back one column as it does in vi and vim. Not sure if this is an oversight or a conscious decision.

Also, the cursor's column position gets decreased when it enters a row with fewer columns than the previous row. So instead of the cursor displaying in column 5 for any rows with >= 5 columns and displaying in the last column for any shorter columns, entering into a row with, say, 3 columns will actually change the cursor's position.

martanne:
> When changing from insert to normal mode, the cursor doesn't move back one column as it does in vi and vim. Not sure if this is an oversight or a conscious decision.

I guess a combination of both. For which reasons would you prefer the vim behavior?

The other issue should hopefully be fixed (well at least for the simple non-wrapped case). And yes the cursor movement code is an ugly mess.

rychipman:
At least for me, it introduces some confusion about what exists at the end of a line. The cursor can move one column past the last character on the line, but that character is not a space. In fact, it is the newline character (which you can verify by deleting it). The exception to this is the end of the file, where that extra column can be accessed by the cursor but does not seem to represent a character at all.

rychipman:
Another somewhat-related issue (which I plan to address at some point) is that a redo leaves the cursor at the *start* position of the redone action, not at the *end* where it ought to be.

martanne:
In the latest commits I introduced support for multiple cursors/selections.

This will further complicate cursor positioning. For now it does not move the cursor upon undo/redo. In general this is all brand new and quite rough around the edges. For example selections are not merged if they overlap, thus causing sometimes unexpected outcomes. Also a few commands are probably not yet converted to deal with multiple cursors. I have a couple of ideas for further improvements but currently not enough time to pursue them.
 
For now the (somewhat arbitrary, suggestions welcome) recognized keybindings are, in normal mode:

    CTRL-J   create a new cursor on line below
    CTRL-K   create a new cursor on line above
    CTRL-P   remove least recently added cursor
    CTRL-N   select word the cursor is currently over, switch to visual mode
    CTRL-A   try to align all cursor on the same column
    ESC      if a selection is active, clear it.
             Otherwise dispose all but the primiary cursor.

In visual mode:

    CTRL-N   create new cursor and select next word matching current selection
    CTRL-X   clear (skip) current selection, but select next matching word
    CTRL-P   remove least recently added cursor
    CTRL-O   creates a new cursor at the start of every line covered by the current selection

[...]

PR #50, opened 2015-06-11
Undo tree

rychipman:
Realized I completely forgot to put in a pull request after finishing this back in May.

This adds history-tree functionality to the editor. Undo and redo work as before, and there are new :earlier and :later commands (shortcuts g- and g+, respectively) which move chronologically through changes.

### DIFF
--- a/text.c
+++ b/text.c
@@ -99,6 +99,7 @@ struct Action {
 	Action *earlier;        /* the previous Action, chronologically */
 	Action *later;          /* the next Action, chronologically */
 	time_t time;            /* when the first change of this action was performed */
+	size_t seq;             /* a unique, strictly increasing identifier */
 };
 
 typedef struct {
@@ -342,6 +343,13 @@ static Action *action_alloc(Text *txt) {
 		return NULL;
 	new->time = time(NULL);
 	txt->current_action = new;
+
+	/* set sequence number */
+	if (!txt->last_action)
+		new->seq = 0;
+	else
+		new->seq = txt->last_action->seq + 1;
+
 	/* set earlier, later pointers */
 	if (txt->last_action)
 		txt->last_action->later = new;
--- a/text.c
+++ b/text.c
@@ -633,13 +633,13 @@ static size_t history_traverse_to(Text *txt, Action *a) {
 		return pos;
 	bool changed = history_change_branch(a);
 	if (!changed) {
-		if (a->time == txt->history->time) {
+		if (a->seq == txt->history->seq) {
 			return txt->lines.pos;
-		} else if (a->time > txt->history->time) {
+		} else if (a->seq > txt->history->seq) {
 			while (txt->history != a)
 				pos = text_redo(txt);
 			return pos;
-		} else if (a->time < txt->history->time) {
+		} else if (a->seq < txt->history->seq) {
 			while (txt->history != a)
 				pos = text_undo(txt);
 			return pos;
--- a/README.md
+++ b/README.md
@@ -135,17 +135,22 @@ As illustrated above, each change to the text is recorded by an old and
 a new span. An action consists of multiple changes which logically belong
 to each other and should thus also be reverted together. For example
 a search and replace operation is one action with possibly many changes
-all over the text. Actions are saved in a tree structure.
+all over the text.
 
-A state of the text can be marked by means of a snapshotting operation.
-Snapshotting saves a new node to the history tree and creates a fresh Action
-to which future changes will be appended until the next snapshot.
-The undo/redo functionality operates on such marked states and switches
-between them by traversing adjacent nodes in the tree.
+The text states can be marked by means of a snapshotting operation.
+Snapshotting saves a new node to the history graph and creates a fresh
+Action to which future changes will be appended until the next snapshot.
 
-Undo and Redo move the state of the document up and down on a single branch
-of the history tree, but other branches can be accessed with the Earlier
-and Later commands, which iterate through snapshotted states chronologically.
+Actions make up the nodes of a connected digraph, each representing a state
+of the file at some time during the current editing session. The edges of the
+digraph represent state transitions that are supported by the editor. The edges
+are implemented as four Action pointers (`prev`, `next`, `earlier`, and `later`).
+
+The editor operations that execute the four aforementioned transitions
+are `undo`, `redo`,`earlier`, and `later`, respectively. Undo and
+redo behave in the traditional manner, changing the state one Action
+at a time. Earlier and later, however, traverse the states in chronological
+order, which may occasionally involve undoing and redoing many Actions at once.
 
 Properties
 ----------


greduan:
+1 (although I dunno if it actually works)

rychipman:
It does work! (I think)
If you wanted to grab the branch and test it out, I'd really appreciate knowing if everything works as you'd expect!

martanne:
Hi thanks for your work and sorry for the late response. I was really busy with university related projects. Here a few remarks from a quick glance over the changes. I will do a more careful review once these are addressed:

 - rebase against current master (one of the reasons I don't particularly
   like the github workflow is that it allows to create pull requests
   which do not merge cleanly, you wouldn't send outdated patches to a mailing
   list, right?).

 - clean up the history, squash commits into logical units (e.g core undo tree
   changes, hook up of :-commands, debug dump functionality etc). The resulting
   pull request should have 3-5 commits, not 29. Avoid unrelated changes, there
   seem to be changes for improved cursor positioning (as mentioned in #42)
   part of this series. Having a clean history also allows usage of `git bisect` etc.

 - clean up the code itself. Avoid unrelated white space changes, spelling fixes
   in comments etc. These are welcome too, but should be applied separately such
   that they do not distract from the main changes at hand. In general match the
   surrounding coding style, e.g. use `Action *a` not `Action* a` when declaring
   a pointer variable. The indention of comments alongside the struct members is
   wrong (use spaces not tabs). The convention of `*_free` function within vis
   is that they can be called with a `NULL` argument. This doesn't seem the case
   for `history_free` (might never happen in the code, but still).

 - documentation: this touches core editor functionality and requires more than a
   couple of half sentences in the README. Explain the idea behind the used data
   structures. Strictly speaking it is not a tree, more like a DAG formed around
   a double linked list, hence it contains cycles and is not really a DAG either.

   Explain what happens when a branch in the "tree" is switched. Possibly add an
   example.

I was somewhat surprised that you do not store the root of the "tree" in the Text struct and instead prefer to follow the chain of `prev` pointers? Also the sibling pointers `alt_prev`, `alt_next` are only used for display purposes, right? `history_free` could also work based on the chronicle order which should contain all nodes. Therefore these should only be introduced in the patch for
the dump functionality.

How did you test the changes? To my astonishment some people are actually using vis in a "production environment". Therefore introducing bugs which might cause data loss has to be avoided if possible. I need to be reasonable confident that it works before merging it into master.

Once it is rebased I will test it myself.

Thanks again!

rychipman:
Thanks Marc! I am fixing these things up now, and will start a new PR from a nicely rebased branch once they are all done.

rychipman:
Hi Marc, I've rebased everything against master, and reorganized my commits a bit. I'm not sure if you want the graphviz dump functionality merged into master -- if not, I can pull it out into a different branch.

I still have to fix up a few of the items you listed in your last comment, but this should at least make things more readable.

I'll push new commits to this branch to address any concerns you want fixed before merging so that they're easy for you to find and review, and then before it gets merged I can rebase them all to their appropriate commits. Sound good?

rychipman:
Editor code should be all set for the most part; please let me know if I missed anything that still needs to be cleaned up.

As far as a couple of the design decisions you mentioned: the `alt_{prev,next}` pointers are indeed just used for display purposes. If you'd prefer, I could move them into their own commit and implement `history_free` without them.

A more detailed writeup for the README will be coming tomorrow.

martanne:
Thanks, this is progressing nicely!

> Editor code should be all set for the most part; please let me know if I missed anything that still needs to be cleaned up.

The addition of the `prev` pointer to the `Change` struct also seems unrelated. It is only used for positioning of the cursor after a redo. Doing this correctly is actually rather tricky.

> As far as a couple of the design decisions you mentioned: the alt_{prev,next} pointers are indeed just used for display purposes. If you'd prefer, I could move them into their own commit and implement history_free without them.

Yes that would be nice! Then the graphviz dump could be separated into its own branch. Also since `history_dump` is exported from text.h it should be prefixed with `text_`. I'm trying to think of a way to combine the tree structure with the [dump of the piece chain itself](https://github.com/martanne/vis/tree/text-dump). However it becomes rather messy.

> A more detailed writeup for the README will be coming tomorrow.

Looking forward to it. Still wondering why you chose to not also store the root of the tree explicitly?

text.c
```diff
@@ -331,30 +337,36 @@ static void span_swap(Text *txt, Span *old, Span *new) {
...
-
-static Action *action_pop(Action **stack) {
-	Action *action = *stack;
-	if (action)
-		*stack = action->next;
-	return action;
-}
-
 /* allocate a new action, empty the redo stack and push the new action onto
  * the undo stack. all further changes will be associated with this action. */
```
martanne: Comment is outdated.

text.c
```diff
@@ -557,38 +585,102 @@ bool text_insert(Text *txt, size_t pos, const char *data, size_t len) {
...
+	lineno_cache_invalidate(&txt->lines);
+	return pos;
+}
+
+bool history_change_branch(Action *a) {
+	bool changed = false;
+	while (a->prev) {
+		if (a->prev->next != a)
+			changed = true;
+		a->prev->next = a;
```
martanne: Assignment could be moved inside the if block, above `changed = true`.

text.c
```diff
@@ -867,19 +1007,21 @@ bool text_delete(Text *txt, size_t pos, size_t len) {
 /* preserve the current text content such that it can be restored by
  * means of undo/redo operations */
 void text_snapshot(Text *txt) {
+	if(txt->current_action)
```
martanne: Missing space between if and opening parenthesis.

text.c
rychipman: fixed by e427ebf

text.c
rychipman: fixed by a236864

text.c
rychipman: fixed by addfb3c

rychipman:
> Still wondering why you chose to not also store the root of the tree explicitly? 

It seemed a bit inelegant to add another field to the `Text` struct when it was only going to be used in some cases, and even then could be supplanted by

```
Action *a = txt->history;
while (a->prev) {
    a = a->prev;
}
// and now a == root
```

As far as I know, there is nowhere where performance would be improved (asymptotically, that is) by storing it; however, I'm perfectly happy to add it in if you think that's a preferable design decision -- this was based entirely on my own style preferences (I try to keep fields as orthogonal as possible).

rychipman:
> The addition of the prev pointer to the Change struct also seems unrelated. It is only used for positioning of the cursor after a redo.

I added it because undoing & redoing an action require traversing the linked list of changes for that action in opposite directions; so I made Change a doubly-linked list. The `prev` pointer gets used in `action_undo` and `action_redo`.

rychipman:
I have done some more rebasing to factor history-dump-specific code out of this branch. There is no longer a `history_free` function, and all the actions are freed in chronological order in `text_free`.

Sibling pointers (`alt_prev` and `alt_next`) are now only introduced in [this branch](https://github.com/rychipman/vis/commits/history-dump).

martanne:
Thanks. I haven't yet tested it thoroughly but so far it looks good (modulo a bug introduced in the latest changes). I've added some testing infrastructure on [a separate branch](https://github.com/martanne/vis/blob/test/test/text.c). Do you intend to add some tests related to history branch switching and chronological traversal? Otherwise I will do it when I find the time.

text.c
```diff
@@ -557,38 +560,102 @@ bool text_insert(Text *txt, size_t pos, const char *data, size_t len) {
...
+	pos = action_redo(txt, a);
+	txt->history = a;
+	lineno_cache_invalidate(&txt->lines);
+	return pos;
+}
+
+bool history_change_branch(Action *a) {
+	bool changed = false;
+	while (a->prev) {
+		if (a->prev->next != a)
```
martanne: You want to enclose the if block with curly braces, otherwise changed will always be true.

text.c
```diff
@@ -557,38 +560,102 @@ bool text_insert(Text *txt, size_t pos, const char *data, size_t len) {
...
+				pos = text_redo(txt);
+			return pos;
+		} else if (a->time < txt->history->time) {
+			while (txt->history != a)
+				pos = text_undo(txt);
+			return pos;
+		}
+	} else {
+		while (txt->history->prev && txt->history->prev->next == txt->history)
+			pos = text_undo(txt);
```
martanne: This assignment has no effect, since `pos` is overwritten on the line below.

text.c
rychipman: Fixed by e9b3056

text.c
rychipman: Fixed by 824fcb8

rychipman:
Fixed those little bugs.

I will write some tests for chronological traversal and submit a PR against `test`

martanne:
I merged your changes. And extended the `:earlier` and `:later` commands to accept time arguments (only very lightly tested). One thing I noticed is that our `time` attribute has only seconds granularity, which probably screws up the comparisons in `history_traverse_to` (because multiple actions will have the same time stamp). It is probably better to use a sequence number and only use the time stamp in `text_restore`.

The [text-dump branch](https://github.com/martanne/vis/tree/text-dump) now tries to dump both the tree structure and the piece chain content. Unfortunately I am not really a graphviz expert and as soon as there is something to undo, i.e. the current state is not a leaf, the layout gets completely messed up ...

rychipman:
I added a slightly more thorough explanation to the README, and I replaced `time` with `seq` in `history_traverse_to`, which should address that issue.

text.c
```diff
@@ -342,6 +343,13 @@ static Action *action_alloc(Text *txt) {
 		return NULL;
 	new->time = time(NULL);
 	txt->current_action = new;
+
+	/* set sequence number */
+	if (txt->last_action)
+		new->seq = 0;
+	else
+		new->seq = txt->last_action->seq + 1;
```
martanne: The if/else cases are inverted, hence you are dereferencing a NULL pointer.

martanne:
Thanks, I think this will work. Commit 2+3 should probably be squashed.

rychipman:
Fixed the backwards conditional and squashed the two commits together.

PR #51, opened 2015-06-23
fixed some typos in text.c comments

rychipman:
these were previously changed in my undo-tree PR, but I factored them out into a separate PR too keep them separate from the undo-tree changes


martanne:
Thanks, applied!

Issue #44, opened 2015-04-15
Change word and repeat with period


erf:
If you press c-w and enter some text, than press escape, and try to repeat this action with period, it only adds the last word, and does not change the current.

martanne:
Thanks for the report! This is an interesting issue. 

It seems like all changes following a command which switches to insert mode should be repeated. In vis the affected commands are change (`c`, `C`, `Y`), append (`a`, `A`, `I`) and open (`o`, `O`). To fix this the changes in insert/replace mode should probably be recorded in an implicit macro which can then be replayed.

For example in vis, in insert mode typing `^wFOO` and then repeating it with a `.` will only insert `FOO` but not delete a word whereas in vim it will do both (delete followed by insert).

However certain movements should reset this "repetition buffer". As an example if in insert mode `foo<RIGHT-ARROW>bar` is typed and this is then repeated only `bar` should be inserted.

erf:
I'm reading 'Practical Vim' now, and he mentions that his favorite action in vim is to use the dot operator to repeat an action, and mentions that Vim records what you do in insert mode, and i also find that a really good feature. You should also include (s, S) and (r, R) and any other insert action. You would probably just need to fill a register for that specific insert, i checked that Vim does not record arrow movements, but a implicit macro for even more recordings would be pretty cool.

martanne:
With the latest commits, `c`, `C`, `a`, `A`, `i`, `I` should now more or less work as expected. Only *very* briefly tested though, further feedback would be appreciated ...

`o`, `O` and `r` aren't yet repeatable (due to the way input handling works, key handlers currently can't push new keys into the input queue which makes implementing them somewhat tricky).

The arrow keys are deliberately not treated specially i.e. they do not reset the repetition macro. 

Interaction with the undo/redo functionality could be improved (if a count is provided, the implementation switches repeatedly  in and out of insert mode which creates a document snapshot).

erf:
Great work! Seems to work as expected. Are you adding <pre>s</pre>, <pre>S</pre> as well?

martanne:
[`s` and `S` should already be recognized?](https://github.com/martanne/vis/blob/master/config.def.h#L210)

erf:
you are correct. but there is a missing functionality, that is, if you visually select some words and then press c or s then you can't repeat that operation. that is if you go to the next line and press dot, you should be able to repeat the operation, from the current position, in the same range as the previous visual selection

martanne:
[...]

PR #54, opened 2015-06-26
Optional vi-like cursor behavior when leaving insert mode.

boomlinde:
This addresses #42 in a non-intrusive way. When the `vicursor` option is set, the cursor will move one column to the left upon exiting insert mode. Since newlines are addressable and appear before the first column of the next line, the exception to this behavior is when insert mode is left while the cursor is on the first column of a line.

 I am not sure that putting this functionality in config.def.h is the right approach, but it works for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,11 @@ and their current support in vis.
        use syntax definition given (e.g. "c") or disable syntax
        highlighting if no such definition exists (e.g :set syntax off)
 
+     vicursor (yes|no)
+
+       whether cursor steps back a column after leaving insert mode,
+       as in nvi/elvis/vim
+
   Each command can be prefixed with a range made up of a start and
   an end position as in start,end. Valid position specifiers are:
 
--- a/config.def.h
+++ b/config.def.h
@@ -566,6 +566,11 @@ static KeyBinding vis_mode_insert[] = {
 static void vis_mode_insert_leave(Mode *old) {
 	/* make sure we can recover the current state after an editing operation */
 	text_snapshot(vis->win->file->text);
+
+	/* If vicursor is enabled, move the cursor back one column unless already on
+	 * the first one */
+	if (vis->vicursor && view_cursor_getpos(vis->win->view).col != 1)
+		movement(&(const Arg){ .i = MOVE_CHAR_PREV });
 }
 
 static void vis_mode_insert_idle(void) {
--- a/editor.h
+++ b/editor.h
@@ -235,6 +235,7 @@ struct Editor {
 	int tabwidth;                     /* how many spaces should be used to display a tab */
 	bool expandtab;                   /* whether typed tabs should be converted to spaces */
 	bool autoindent;                  /* whether indentation should be copied from previous line on newline */
+	bool vicursor;                    /* whether cursor behaves like in nvi/elvis/vim after leaving insert mode */
 	Map *cmds;                        /* ":"-commands, used for unique prefix queries */
 	Map *options;                     /* ":set"-options */
 	Buffer buffer_repeat;             /* holds data to repeat last insertion/replacement */
--- a/vis.c
+++ b/vis.c
@@ -1329,6 +1329,7 @@ static bool cmd_set(Filerange *range, enum CmdOpt cmdopt, const char *argv[]) {
 		OPTION_SYNTAX,
 		OPTION_NUMBER,
 		OPTION_NUMBER_RELATIVE,
+		OPTION_VICURSOR,
 	};
 
 	/* definitions have to be in the same order as the enum above */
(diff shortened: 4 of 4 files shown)


martanne:
Thanks! If possible I would like to avoid introducing options like `vicursor`. Instead if this is what people expect then we should probably make it the default and only behavior.

Having said that, why do you prefer it? What is the reason vi(m) did it that way? 

Intuitively leaving insert mode and then entering it again should not change the cursor position.

greduan:
No idea why they would do it that way, but it's muscle memory to expect it to go left one after leaving insert. :)

I'm fine with this setting being available, but I don't care if it isn't implemented either.

trengrj:
Since newlines are not addressable in vim, when you are appending to a line in vim the cursor is temporarily over an invalid character (and hence the move back after an insert / append). See http://stackoverflow.com/a/17975729.

The newline addressability decision makes the vim / vis interfaces diverge.

On a single empty line, in vis `a`  moves cursor to the next line (operates like `ji`), while in vim `a` operates like `i`.

The '$' operator in vis can actually move you left if you are on a newline.

If newline addressiablity is kept, `vicursor` isn't strictly needed. However I'm not sold on newline addressibility as it adds clutter to the line (an additional character), breaks the meaning of `a`, and will probably cause more issues with vim / vis differences in core editing commands.

rychipman:
The original impetus for raising issue #42 was that it did not match my expectations coming from vim (a muscle-memory thing, as @greduan said).

The more I think about it though, I don't need or even really particular care about that UX nuance.

As @trengrj mentioned, newline addressability does introduce other interface divergences, and those divergences are the ones that worry me a bit more. For example:

- `a` at the end of a line does the same thing as `i` (doesn't move the cursor forward) which is a bit uncomfortable.
- being able to delete a newline with `x`. except for the final newline of a file, which is addressable, but cannot be deleted.
- related to the previous item, `dd` on the last line of a file removing its contents but not the line itself.
- text editors in general (not just vim) have conditioned me to expect that end-of-line whitespace will **not** be a newline character.
- related to the previous item, it feels weird that the visual representation of a newline is the same as a space.

Those are just a handful of edge cases that I was able to find in about 5 minutes of playing around. Now it may be entirely possible to resolve these inconsistencies while maintaining newline addressability, but it does seem to be a common cause of a number of these UX 'bugs'.

martanne:
I agree that the newline addressability causes some differences compared to vim. The question then is whether they are harmful or not. I'm yet undecided and in general quite like it. It certainly is a controversial thing, even one of the first [comments after the public announcement of vis](http://lists.suckless.org/dev/1409/23498.html) was about it.

> On a single empty line, in vis a moves cursor to the next line (operates like ji), while in vim a operates like i

Is this a bad thing? `a` always does the same thing not depending on whether the line is empty or not. Why would you even use `a` when you expect the same thing to happen as for `i`? vis seems to give you more options in this specific case.

> The '$' operator in vis can actually move you left if you are on a newline.

True, but does it really matter? Why would you use `$` if you are already at the end of the line (on a newline)?

> a at the end of a line does the same thing as i (doesn't move the cursor forward) which is a bit uncomfortable.

`a` moves the cursor forward, the complain from @trengrj was that it even does it for empty lines / when the cursor is on a newline character.

> it feels weird that the visual representation of a newline is the same as a space.

I tend to agree, that is why I implemented [:set show newlines=1](https://github.com/martanne/vis/commit/87c7258ccc683965d2b6a5190f1e98d74c8d25d3). If you enable this setting you will see that you can in fact delete the last newline of a file. What you are probably refering to is the (illegal) position after the end of file where you would insert new text. I agree that in this case `x` should probably delete the last valid character of the file. Also `dd` should probably not place the cursor at this EOF position but at the start of the last line. 

It is the way it is because I wanted to avoid special cases. Another reason might be that I rarely edit files at the end and therefore it didn't bother me too much so far.

martanne:
Changing the cursor position when leaving insert mode would break the the dot command which is implemented by replaying an implicit macro which might repeatedly switch in and out of insert mode (see also #44).

I still like newline addressability and I'm thus closing this. If someone sends an updated patch which works with the repetition code, this can be reconsidered.

martanne:
Because this pull request is regularly referenced when discussing newline addressability related issues, here a quick update:

> The '$' operator in vis can actually move you left if you are on a newline.

This is no longer the case.

> On a single empty line, in vis a moves cursor to the next line (operates like ji), while in vim a operates like i

Due to popular demand (e.g. #287, #295) this was changed to match the vim behavior.

Repeatedly switching in and out of insert mode is still idempotent in the sense that the cursor position remains unchanged. This is the desired behavior for vis and there currently exist no plans to change it.

unrealapex:
> This is the desired behavior for vis and there currently exist no plans to change it.

I would love for this feature to be implemented. This behavior is the only reason feel the urge to switch to nvi.

Issue #56, opened 2015-07-03
incorrect display when file begins with a multibyte character


TieDyedDevil:
When the first character of a file is a multibyte character, vis displays a sequence of Unicode REPLACEMENT CHARACTER.

I believe that the cause is the use of an anonymous (i.e. NULL) mbstate_t* object in the call to mbrtowc() in view_draw(). While this might actually be a bug in the implementation of mbrtowc(), it is avoided by use of an explicitly initialized mbstate_t object.

Here's a short patch:
```diff
diff --git a/view.c b/view.c
index d49c417..80f7038 100644
--- a/view.c
+++ b/view.c
@@ -349,6 +349,8 @@ void view_draw(View *view) {
 	memset(match, 0, sizeof match);
 	/* default and current curses attributes to use */
 	int default_attrs = COLOR_PAIR(0) | A_NORMAL, attrs = default_attrs;
+  /* start from known multibyte state */
+  mbstate_t mbstate = { 0 };
 
 	while (rem > 0) {
 
@@ -403,7 +405,7 @@ void view_draw(View *view) {
 			}
 		}
 
-		size_t len = mbrtowc(&wchar, cur, rem, NULL);
+		size_t len = mbrtowc(&wchar, cur, rem, &mbstate);
 		if (len == (size_t)-1 && errno == EILSEQ) {
 			/* ok, we encountered an invalid multibyte sequence,
 			 * replace it with the Unicode Replacement Character
```

martanne:
Thanks, I didn't notice since it seems to work as is with [musl libc](http://www.musl-libc.org/).

PR #64, opened 2015-08-06
Add APL syntax highlighting.

TieDyedDevil:
Syntax highlighting compatible with GNU APL.
http://www.gnu.org/software/apl/

For sample *.apl files, see (for example)
https://github.com/TieDyedDevil/apl-pkg

... and some of the projects linked from:
http://www.gnu.org/software/apl/Community.html

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -1379,5 +1379,40 @@ static Syntax syntaxes[] = {{
 		&colors[COLOR_DATATYPE],
 	}}
 },{
+	.name = "apl",
+	.file = "\\.apl$",
+	.settings = (const char*[]){
+		"set number",
+		NULL
+	},
+	.rules = {{
+		"(⍝|#).*$",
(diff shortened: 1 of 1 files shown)


martanne:
Thanks. APL seems to nicely test some basic Unicode support. Did you notice any issues? How do you input these "special" symbols? Do you miss digraph support?

TieDyedDevil:
I wrote this for APL input:
https://github.com/TieDyedDevil/akt

It's used like this:
$ akt|vis ...

I haven't noticed any Unicode issues since #56.

Thanks. APL seems to nicely test some basic Unicode support. Did you notice
any issues? How do you input these "special" symbols? Do you miss digraph
support?

—
Reply to this email directly or view it on GitHub
<https://github.com/martanne/vis/pull/64#issuecomment-128620682>.

PR #68, opened 2015-08-19
Generate keydoc.txt reference from config.h.

TieDyedDevil:
I thought it'd be nice to not have to grovel through config.h for key binding info, so I wrote an awk script to extract something that can be used as a reference document.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ config.h
 *.o
 nano
 vis
+keydoc.txt
--- a/Makefile
+++ b/Makefile
@@ -11,12 +11,15 @@ vis: config.h config.mk *.c *.h
 	@echo ${CC} ${CFLAGS} *.c ${LDFLAGS} -o $@
 	@${CC} ${CFLAGS} *.c ${LDFLAGS} -o $@
 
+keydoc.txt: config.h
+	awk -f keydoc.awk < $^ > $@
+
 debug: clean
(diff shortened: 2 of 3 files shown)


martanne:
I agree that documentation is currently lacking. Your approach is a reasonable short term solution, however I would prefer

* a proper manual page which describes the default keybindings
* a `:help` command which displays all bindings currently available

The second point should be done when adding support for runtime key bindings. I would like to experiment with a solution where the editor would start with no bindings at all, before adding them (ideally also creating the different modes dynamically) via some sort of `:map` command. Not yet completely sure whether it is doable in a clean way.

The key bindings would be stored in a [map](https://github.com/martanne/vis/blob/master/map.h) where the indices are keys represented as strings (as provided by something like [libtermkey](http://www.leonerd.org.uk/code/libtermkey/overview.html)). The values would be either strings (if the key is bound to some other key) or a function pointer executing the corresponding action together with a human readable description. The `:help` command would then use `map_iterate` to print all bindings with their corresponding description.

There is also the question how performant such a map based lookup would be. While the current lookup code is as simple as it gets from an algorithmic viewpoint it has a sequential memory access pattern whereas the map (currently implemented as a tree structure) would be more scattered.

martanne:
I finally had some time to start working in the direction outlined in the last comment. The resulting `:help` command is far from perfect, but should at least be more informative than the generated keydoc.txt file.

The required changes were quite invasive and not yet thoroughly tested, feel free to send further cleanup patches ...

PR #74, opened 2015-10-06
Version

eworm-de:


martanne:
Sorry that I forgot your patch on the mailing list. I changed the default compression format to gzip. I hope this won't cause any problems for you? Thanks for maintaining the vis Arch Linux package!

I noticed that you specified the license as [custom](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=vis-git#n13) while it is actually a standard [ISC license](https://en.wikipedia.org/wiki/ISC_license) as for example recommended by [OpenBSD](http://www.openbsd.org/policy.html).

eworm-de:
The patch was/is just in preparation for a release... So this did not cause any trouble. Just discovered the code on github and decided to resend with a pull request.

No problem with the compression, I am fine with gzip. ;)

Probably the wrong license was caused by copy'n'paste... I will change that.

eworm-de:
Ah, `custom` was not completely wrong as the license file is not distributed with the Arch Linux licenses package. I updated to `custom:ISC` now.

PR #86, opened 2015-10-31
Add ledger lexer

clehner:
It's not as featureful as the last one, or the vim-ledger plugin, but it's got the basics.


martanne:
Thanks, I applied your patch!

PR #120, opened 2015-11-21
Fixed $VIS_THEME warning message

roboguy13:
If the `$VIS_THEME` environment variable doesn't exist, the existing warning incorrectly indicates that you should set `$VIS_PATH`. This changes the warning to suggest setting `$VIS_THEME`.


martanne:
Thanks for your contribution! However I think the current code is correct. In the [else case](https://github.com/martanne/vis/blob/master/ui-curses.c#L1006) `VIS_THEME` is not set, hence the default profile is used. This should always succeed unless vis is unable to locate the `lexers` directory (as is the case when running from the build directory without a previously done `make install`), hence the warning about setting `VIS_PATH`.

PR #122, opened 2015-11-23
include https as link in markdown

erf:
include https as link in markdown. maybe there is a more clever way of doing this..?

### DIFF
--- a/lexers/markdown.lua
+++ b/lexers/markdown.lua
@@ -48,7 +48,8 @@ local link = token('link', P('!')^-1 * l.delimited_range('[]') *
                             (S(' \t')^1 *
                              l.delimited_range('"', false, true))^-1 * ')' +
                             S(' \t')^0 * l.delimited_range('[]')) +
-                           P('http://') * (l.any - l.space)^1)
+                           P('http://') * (l.any - l.space)^1 +
+                           P('https://') * (l.any - l.space)^1)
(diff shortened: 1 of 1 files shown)


martanne:
Maybe @TieDyedDevil could take a general look at the markdown syntax definition as he is the de facto maintainer for the syntax highlighting definitions. As an example in vis's README.md the diagrams which are indented by 4 spaces are colored improperly (whether it makes sense at all to highlight them is another question) ...

erf:
👍

TieDyedDevil:
Ah. Sorry. I missed this the first time around.

Yes, this change should be fine for the purpose of treating an http link the same as an https link.

The only thing I'd ask is: Might it make sense to generalize this further to give all URIs the same treatment, regardless of method? IOW, would it make sense to match the method (e.g. http, https, ftp, webdav, ...) using a wildcard?

erf:
Why not all links on the format of []() ?

erf:
Hmm.. I see atom and neovim only highlights url's if they are of type `[]()`

apotheon:
That's because URIs are not syntactically significant to Markdown except as contents of Markdown link hints.  Did I miss discussion of the reason URIs should be highlit universally in Markdown?

erf:
I think only markdown links `[]()` should be highlighted, not URI's

martanne:
Since the 'correct' thing to do largely depends on personal taste, I chose to remain compatible with the upstream scintillua project.

PR #129, opened 2015-12-30
Mark that Lua is optional in README.md

erf:
I understand that since commit 13a95942b6e33033e4bbbe27cc8cafc98c6924a2 , LUA is optional with 
`make CONFIG_LUA=0` i thought it might be good to note this in the README.md. 

Also, maybe it is a good idea to check if lua is on the system and then build without LUA as default if not present.


martanne:
If one wants a customized build it is kind of expected to read through `config.mk`.

As for auto detection, I would like to minimize guess work and the current solution seems simple enough ...

PR #132, opened 2016-01-09
config.mk: Allow some variables to be set by the user

Pyrohh:
There is no `-ldl` on OpenBSD, for example, so compilation fails if that flag is used.

---

These were the only flags I needed to override when making a port for OpenBSD, but I realize it's inconsistent. Other porters for other systems might need to override additional flags; if you'd like, I can do that (`=` -> `?=`) for the other variables in `config.mk` where it makes sense.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -9,10 +9,10 @@ CONFIG_SELINUX=0
 # Customize below to fit your system
 
 PREFIX ?= /usr/local
-MANPREFIX = ${PREFIX}/share/man
+MANPREFIX ?= ${PREFIX}/share/man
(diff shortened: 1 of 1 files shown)


martanne:
This should be possible by doing:

    make LIBS=...

but I guess you want to be able to do:

    LIBS=... make

right? If we decide this is the way to go then it should probably be changed for all relevant make variables.

Thanks for your other patches, they have been merged!

Pyrohh:
> This should be possible by doing:

Heh, I had no idea. I just updated my port to do that, much simpler, thanks.

> Thanks for your other patches, they have been merged!

Thanks, that was very quick.

PR #135, opened 2016-01-09
gitignore: Fix pattern

Pyrohh:
From `man gitignore`:

```
o   A leading slash matches the beginning of the pathname. For example,
   "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
-config.h
+/config.h
 *.swo
 *.swp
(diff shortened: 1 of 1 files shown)


martanne:
Ok, I guess this should probably also be changed for `config.h` and `dependency`?

Pyrohh:
Done.

PR #138, opened 2016-01-10
config.mk tweaks

Pyrohh:
The LIBS change is required for building on OpenBSD.

As for the other changes:
- vis already builds fine with -Wstrict-prototypes there shouldn't be
  much harm in using it unconditionally.
- The size difference between -O2 and -Os is tiny on OpenBSD (233K vs
  218K).  Also, -Os disables 8 optimizations compared to -O2; given
  that, the potential performance benefits, as small as they may be,
  might be more important than saving 15K.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -12,7 +12,7 @@ PREFIX ?= /usr/local
 MANPREFIX = ${PREFIX}/share/man
 SHAREPREFIX = ${PREFIX}/share/vis
 
-LIBS = -lm -ldl -lc
+LIBS = -lc
 
 CFLAGS_TERMKEY = $(shell pkg-config --cflags termkey 2> /dev/null || echo "")
 LDFLAGS_TERMKEY = $(shell pkg-config --libs termkey 2> /dev/null || echo "-ltermkey")
@@ -22,7 +22,7 @@ LDFLAGS_CURSES = $(shell pkg-config --libs ncursesw 2> /dev/null || echo "-lncur
 
 ifeq (${CONFIG_LUA},1)
 	CFLAGS_LUA = $(shell pkg-config --cflags lua5.2 2> /dev/null || echo "-I/usr/include/lua5.2")
-	LDFLAGS_LUA = $(shell pkg-config --libs lua5.2 2> /dev/null || echo "-llua")
+	LDFLAGS_LUA = $(shell pkg-config --libs lua5.2 2> /dev/null || echo "-llua -lm")
 endif
 
 CFLAGS += -DCONFIG_LUA=${CONFIG_LUA}


martanne:
A couple of comments:

1. the build system is a mess
2. Adding libraries in the `if ${OS}` section which are already part of the default `LIBS=` (as you are doing for `-lm`) seems wrong
3. `-ldl` should probably not be part of the default `LIBS=` but added as a dependency of lua (if required). The lua pkg-config file should reference it if it is needed.
4. Why do we need the math library? Presumably because of Lua, hence 3. should apply.
5. I'm not a big fan of adding individual warning flags (except if they are *really* useful, what makes you think `-Wstrict-prototypes` falls into this category?). What could be useful is to add `-pedantic` or similar "global options" to the `make debug` target and review/fix the resulting warnings.
6. Did you observe any performance issues? In some cases smaller is faster because the cache behavior is better. Of course this can't be generalized and would need further analysis.
7. Are you on a 64bit system? Here a dynamically linked vis binary weights in at about 175K. As noted elsewhere vis is currently not optimized for size, its (relative) small size is more of a by product.

Pyrohh:
> 3. -ldl should probably not be part of the default LIBS= but added as a dependency of lua (if required). The lua pkg-config file should reference it if it is needed.

Okay, although `-ldl` doesn't exist on OpenBSD (it's part of the stdlib), so doing so unconditionally would fail the build (`pkg-config --libs lua52` on OpenBSD results in `-L/usr/local/lib -llua5.2 -lm`).

> 4. Why do we need the math library? Presumably because of Lua, hence 3. should apply.

Yes, that's right. I've pushed a new commit hoping to address that.

> I'm not a big fan of adding individual warning flags (except if they are really useful, what makes you think -Wstrict-prototypes falls into this category?). What could be useful is to add -pedantic or similar "global options" to the make debug target and review/fix the resulting warnings.

It helps avoid functions of the form `void fn();`, written with the intent to take no parameters, which isn't the reality of the sitatuion; see https://stackoverflow.com/questions/693788/c-void-arguments/693794#693794

I've seen it happen many times in various codebases, and if the aforementioned flag was used it could have been avoided.  Besides that it never seems to trigger false positives, in my experience at least.

> Did you observe any performance issues? In some cases smaller is faster because the cache behavior is better. Of course this can't be generalized and would need further analysis.
>
> Are you on a 64bit system? Here a dynamically linked vis binary weights in at about 175K. As noted elsewhere vis is currently not optimized for size, its (relative) small size is more of a by product.

No, no performance issues, and yes, I'm on a 64 bit system; OpenBSD's uses GCC 4.2.1, which may be the reason for the size difference, among other things.

And as you say, vis isn't optimized for size, so I thought that even a mild performance improvements would be worth a tiny increase in vis' size. That is of course speculative.

Pyrohh:
Ugh, this is all terribly messy. I can't help but feel it might be better if this complexity was kept out of the main repo, at least for OpenBSD. I don't even have to do that much in the current port to get vis to build properly:

```
MAKE_FLAGS =		LIBS=-lm \
			CFLAGS_LUA=-I${MODLUA_INCL_DIR} \
			LDFLAGS_LUA=-llua$(MODLUA_VERSION) \
```
which is equivalent to
```
$ gmake LIBS=-lm CFLAGS_LUA=-I/usr/local/include/lua-5.2 LDFLAGS_LUA=-llua5.22
```

Pyrohh:
Closing this for the reason mentioned above.

erf:
Could cmake make building for multiple platforms easier ? I know little of these system just curious ..

PR #146, opened 2016-01-13
Tab alignment

### LINKED ISSUE #144: Tab alignment
Tabbed text doesn't seem to align up.
```
hello    tabend
yo        tabend
```

erf:
Should fix #144 
Calculates remaining spaces for each expanded tab per cursor.

Note:
I was not able to call `cursor->col` so i made the method `view_cursor_col'.

### DIFF
--- a/view.c
+++ b/view.c
@@ -435,6 +435,10 @@ CursorPos view_cursor_getpos(View *view) {
 	return pos;
 }
 
+int view_cursor_col(Cursor *c) {
+  return c->col;
+}
+
 static void cursor_to(Cursor *c, size_t pos) {
 	Text *txt = c->view->text;
 	c->mark = text_mark_set(txt, pos);
--- a/view.h
+++ b/view.h
@@ -103,6 +103,8 @@ int view_colorcolumn_get(View*);
 
 /* get main cursor position in terms of screen coordinates */
 CursorPos view_cursor_getpos(View*);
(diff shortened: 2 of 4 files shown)


martanne:
Thanks! In the meantime I implemented it slightly differently. A couple general comments:

1. When you are contributing to an existing code base it is usually a good idea to stick to the existing coding style. As a general rule your code should look the same as the surrounding one. For example for vis we use tabs (i.e. `\t`) for indentation and spaces for alignment. Your code is indented by 4 spaces.
2. While it is fine to make as many commits as you like during development, before sending your changes upstream you should clean up your history. It should be grouped into logical commits. In general, and especially when dealing with simple changes as we are here, there is no point in keeping intermediate commits that introduce bugs which are then fixed by later patches.  For this particular pull request 1 patch would have been preferable.
3. The comments in the header files above the function prototypes are the closest thing to an API documentation that we have. They should clarify API usage if needed (yes they could be improved). Adding stuff like "i was not able to call c->col in vis.c" does not clarify anything.
4. You can't rely on `view_cursor_col` for cursor which are not currently in the visible area.

erf:
Thanks for constructive feedback!
I will enable option to display tabs and spaces next time, and look into git rebase.
Really impressed with the progress so far on an awesome editor.
I think i will have to learn some more c so i can contribute.

Issue #152, opened 2016-01-25
Star (*) does not match word


erf:
Star `*` does not match word. If you press `*` on top of word, it does not mark and move to similar word.

erf:
The bug was introduced in commit 495cc3d

TieDyedDevil:
It works for me, at least insofar as moving to the next word.

To the best of my knowledge, vis has never visually highlighted the word.


On Sun, Jan 24, 2016 at 04:04:06PM -0800, Erlend wrote:
>    Star * does not match word If you press * on top of word, it does not
>    mark and move to similar word
> 
>    —
>    Reply to this email directly or [1]view it on GitHub.
> 
> References
> 
>    1. https://github.com/martanne/vis/issues/152

-- 
"I think as musicians, we have more vision than we credit ourselves for. We should dive into it without reserve, and take the sort of risks it takes to be divisive and impactful. Things have yet to be done."
   Aaron Fowler Clark

"Do not seek to follow in the footsteps of the masters. Seek what they sought."
   Matsuo Basho

TieDyedDevil:
Sorry that I responded before you added this information...

I had updated to b9e8d20 (which I believe is current as I write this) to check the behavior on my machine.


On Sun, Jan 24, 2016 at 04:06:32PM -0800, Erlend wrote:
>    The bug was introduced in commit [1]495cc3d
> 
>    —
>    Reply to this email directly or [2]view it on GitHub.
> 
> References
> 
>    1. https://github.com/martanne/vis/commit/495cc3d53cdb7182c98827f8aa0ad2726167daaf
>    2. https://github.com/martanne/vis/issues/152#issuecomment-174355100

-- 
"I think as musicians, we have more vision than we credit ourselves for. We should dive into it without reserve, and take the sort of risks it takes to be divisive and impactful. Things have yet to be done."
   Aaron Fowler Clark

"Do not seek to follow in the footsteps of the masters. Seek what they sought."
   Matsuo Basho

erf:
I'm on OSX. Moving to the next word only works before 495cc3d . You are right about words not highlighting..

martanne:
Yes the used regex features are probably not POSIX compilant. I only tested it with musl/glibc. I guess it doesn't work with *BSD either?

What happens when you search for a word via `/\<word\>`  (on the / prompt)?

erf:
Searching for words with `/\<word\>` works fine.

martanne:
[...]

PR #154, opened 2016-01-31
regex match word fix for * and #

erf:
This fixes the problem #152 where i could not match words by `*`and `#` on OSX. 
Could you test if this works on linux as well?

### DIFF
--- a/vis-motions.c
+++ b/vis-motions.c
@@ -16,9 +16,9 @@ static bool search_word(Vis *vis, Text *txt, size_t pos) {
 	char *buf = text_bytes_alloc0(txt, word.start, text_range_size(&word));
 	if (!buf)
 		return false;
-	snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
+	snprintf(expr, sizeof(expr), "\\(%s\\)", buf);
 	free(buf);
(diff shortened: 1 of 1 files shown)


vis-motions.c
```diff
@@ -16,9 +16,9 @@ static bool search_word(Vis *vis, Text *txt, size_t pos) {
 	char *buf = text_bytes_alloc0(txt, word.start, text_range_size(&word));
 	if (!buf)
 		return false;
-	snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
+	snprintf(expr, sizeof(expr), "\\(%s\\)", buf);
 	free(buf);
-	return text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
+	return text_regex_compile(vis->search_pattern, expr, 0) == 0;
```
twe4ked: If this does need to change it should probably use `REG_BASIC`:

```
$ man regex
     REG_EXTENDED  Compile modern (``extended'') REs, rather than the obsolete (``basic'') REs that are the default.
     REG_BASIC     This is a synonym for 0, provided as a counterpart to REG_EXTENDED to improve readability.
```

erf: Thanks, i've updated it.

PR #159, opened 2016-01-31
using posix compliant word-boundaries

### LINKED ISSUE #152: Star (*) does not match word

erf:
Added a fix for POSIX compliant word boundaries. Should fix #152.
Ref (bottom of page):
http://www.regular-expressions.info/wordboundaries.html

### DIFF
--- a/vis-motions.c
+++ b/vis-motions.c
@@ -16,7 +16,7 @@ static bool search_word(Vis *vis, Text *txt, size_t pos) {
 	char *buf = text_bytes_alloc0(txt, word.start, text_range_size(&word));
 	if (!buf)
 		return false;
-	snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
+	snprintf(expr, sizeof(expr), "[[:<:]]%s[[:>:]]", buf);
 	free(buf);
 	return text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
 }


c00kiemon5ter:
uhm, this is not what it says. 

`[[:<:]]` breaks down to 

`[` which opens up a character list or range to be matched

`[:` which instead of the above, it marks a predefined character class

`<` the class name (in a correct case, this would be `alpha`, `digit`, `alnum`, etc)

`:]` which marks the end of the character class

`]` which ends the character list or range

..so, no, this does not implement word boundaries.
It tries to match one character from the `<` and `>` character class, that both do not exist.

erf:
Did you read the [link](http://www.regular-expressions.info/wordboundaries.html) ?
It states that: 
```
these tokens are word boundaries that have nothing to do with and cannot be used inside character classes
```

erf:
I have also tested that it works as expected.

c00kiemon5ter:
Yeah, I was just doing that, and it seems to work on GNU grep at least. 
It doesn't seem to work inside vim though. 

I've never come across that before and I don't see it in my POSIX doc. 
Neither in the bracket expression section, nor the RE grammar definition..

erf:
I believe `vim` has a custom implementation.
Yeah, hard to find docs - should send a thank you to the [author](http://www.just-great-software.com/img/JFG_14070612_32439.JPG) .
BTW - noticed words with unicode characters are not matched..

martanne:
I tried that before, it doesn't work on musl nor glibc ...

Maybe an ugly #ifdef is warranted here. Wondering whether it works on *BSD ...

erf:
Could you recreate word boundaries with other regex commands, with something like `(^|\W)foo(\W|$)` except not including the hyphens ..

erf:
`(?<=\W|^)foo(?=\W|$)` should do it, but **lookarounds** are not supported in POSIX regex it [seems](http://www.regular-expressions.info/refadv.html) ..

PR #163, opened 2016-02-10
ifdef APPLE hack for word match

erf:
Hack for  #152

### DIFF
--- a/vis-motions.c
+++ b/vis-motions.c
@@ -16,7 +16,11 @@ static bool search_word(Vis *vis, Text *txt, size_t pos) {
 	char *buf = text_bytes_alloc0(txt, word.start, text_range_size(&word));
 	if (!buf)
 		return false;
+	#ifdef __APPLE__
+	snprintf(expr, sizeof(expr), "[[:<:]]%s[[:>:]]", buf);
+	#else
 	snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
+	#endif
 	free(buf);
 	return text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
 }


martanne:
Thanks for spending time on this issue. Does the following also work?

```diff
diff --git a/vis-motions.c b/vis-motions.c
index 91224b5..2abae8a 100644
--- a/vis-motions.c
+++ b/vis-motions.c
@@ -17,8 +17,13 @@ static bool search_word(Vis *vis, Text *txt, size_t pos) {
 	if (!buf)
 		return false;
 	snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
+	bool ret = text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
+	if (!ret) {
+		snprintf(expr, sizeof(expr), "[[:<:]]%s[[:>:]]", buf);
+		ret = text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
+	}
 	free(buf);
-	return text_regex_compile(vis->search_pattern, expr, REG_EXTENDED) == 0;
+	return ret;
 }
 
 /** motion implementations */
```

Apologies if the stupid github interface mixes up tabs/spaces ...

erf:
That would be a better fix - but it does not work. The first `text_regex_compile` returns `0`, as there is no error, but the word is not found - so instead, i believe you need to check `vis->search_pattern->regex.rm_so` or `vis->search_pattern->regex.rm_eo`, to see if a actual word is found, if not search for the other. I tried to do this, but i got a `error: incomplete definition of type 'struct Regex'`.

martanne:
We can probably just switch the order around, first try `[[:<:]]%s[[:>:]]` which does not compile here on musl/glibc.

I will push it later ...

erf:
:+1:

PR #169, opened 2016-02-15
Open Vis given line number after filename

erf:
When opening a file, Vis now accepts a line number both before and after filename (like in vim).
Example: `vis somefile +100`


martanne:
Congratulation, you broke `vis +100 file1 +200 file2`

erf:
I seem to have. In Vim when passing multiple files and lineno as arguments, it creates buffers for all files but only opens one window (the first file) with given lineno. I like it better how Vis solves this. Would still prefer line number after the filename but it doesn't really matter and i guess it is silly to support both cases with multiple files. Close?

PR #171, opened 2016-02-15
Open Vis with file explorer if given folder as argument

### LINKED ISSUE #170: have `vis somefolder` start up with file explorer ( like with `:edit .`)
When you try to open a folder with Vis, instead of writing an error, open the "file explorer" like when you press `:edit somefolder`.

erf:
Should fix #170


erf:
You are right. Could we just break after the first directory ?

erf:
Update to break on EISDIR. Simplified.

PR #172, opened 2016-02-15
Make <C-c> an alias for <Escape> in normal mode

erf:
Make <C-c> an alias for <Escape> in normal mode to remove multiple cursors.


martanne:
I committed this patch (with some related cleanup). Thanks.

However why do you have all unrelated changes on the same branch/pull request?

Another reason why I don't particularly like github, it makes it too easy to do stupid things ...

erf:
No idea why all those PR got stacked. I made separate branches for them. 
Did an update regarding your second comment, see separate PR.

PR #173, opened 2016-02-16
Display ASCII-127 (DEL) character as ^?

rgburke:
Vis currently displays the ASCII-127 (DEL) character as `^M-?`, whereas vim (and other editors) display this character as `^?`.

This commit updates vis to display this character as `^?`.


martanne:
Thanks! There also seems to be a problem with NUL bytes.

Issue #161, opened 2016-02-02
Layout independent keybindings


xomachine:
There is a problem for all non-latin keyboard layout users. Current keybindings work only with latin letters. Every attempt to edit any non-latin text via vis makes user to perform a lot of layout switching instead of actually editing. It would be great if vis's keybindings will be independent from the current layout.

As temporary workaround the bash script below can be used to modify config.h and make it friendly to any non-latin layout (for layouts other than russian the "nonlatin" variable should be filled by analogues of symbols from "latin" variable, and " symbols should be escaped). 
Script usage:

- copy the script body to the new file [scriptname].sh in the "vis" source root folder
- type following command before compilation

``` bash
$ sh [scriptname].sh < config.def.h > config.h
```
[Script body](http://pastebin.com/EjGbQEjL)

martanne:
I guess in vim you use the `:set langmap` feature? Which remaps keys if not in insert/replace mode, correct?

In vis this could be achieved by translating the keys in the [getkey](https://github.com/martanne/vis/blob/66a56c827efcac36e07d6ac34bd05681ab8364e9/vis.c#L747) function.

Something like

```C
if (!vis->mode->input) {
      const char *mapped = map_get(vis->langmap, key);
      if (mapped)
            return mapped;
}
```

Where `vis->langmap` would be a [map](https://github.com/martanne/vis/blob/master/map.h#L9) populated with your translation table.

xomachine:
Actually I don't use vim.(It has lack of minimalism, that is the reason why I use vis)  But I've heard about this feature. Using the `:set langmap` does not allow to remap punctuation signs used in command syntax. For example in russian layout  ";" key should be mapped as "ж" letter. Another way to fix shortcuts in vim is using `:set keymap` command, but I have no idea how it works.

The best solution, in my opinion, would be handling keycodes instead of letters in all modes other than insert/replace and command line. As far as I know it is hard to achieve due to absence of necessary functionality in libtermkey.

martanne:
> Actually I don't use vim.(It has lack of minimalism, that is the reason why I use vis) 

Good attitude, wondering when vis will become too bloated ;-)

Unless I'm missing something working with letters should work fine. What is wrong with the patch below? Add all your desired mappings to config.def.h.

```diff
diff --git a/config.def.h b/config.def.h
index 368e80c..4946ba2 100644
--- a/config.def.h
+++ b/config.def.h
@@ -3,6 +3,11 @@
[...]

PR #174, opened 2016-02-17
langmap command implementation

xomachine:
Refered to  #161
I've decided to implement some analogue of vim's `:set langmap` command.
The implementation is based on [cmd_map](https://github.com/martanne/vis/blob/master/vis-cmds.c#L1033)'s code. The code shared between [cmd_map](https://github.com/xomachine/vis/blob/master/vis-cmds.c#L1110) and new [cmd_langmap](https://github.com/xomachine/vis/blob/master/vis-cmds.c#L1071) functions was separated in [bind_key](https://github.com/xomachine/vis/blob/master/vis-cmds.c#L1036) function.

Any criticism is appreciated.

Usage:
`:langmap <latin keys> <non-latin keys>`

Example:
`:langmap qwertyuiop[] йцукенгшщзхъ`

PS: I've tried to check my code for memory leaks using valgrind, but encountered  a leak in original [cmd_map](https://github.com/martanne/vis/blob/master/vis-cmds.c#L1033) function. Valgrind complains on [this calloc call](https://github.com/martanne/vis/blob/master/vis-cmds.c#L1044).

### DIFF
--- a/vis-cmds.c
+++ b/vis-cmds.c
@@ -89,6 +89,8 @@ static bool cmd_help(Vis*, Filerange*, enum CmdOpt, const char *argv[]);
 /* change runtime key bindings */
 static bool cmd_map(Vis*, Filerange*, enum CmdOpt, const char *argv[]);
 static bool cmd_unmap(Vis*, Filerange*, enum CmdOpt, const char *argv[]);
+/* set language specific key bindings */
+static bool cmd_langmap(Vis*, Filerange*, enum CmdOpt, const char *argv[]);
 
 /* command recognized at the ':'-prompt. commands are found using a unique
  * prefix match. that is if a command should be available under an abbreviation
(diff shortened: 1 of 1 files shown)


martanne:
Just some quick comments from a first glance (I will take a closer look later on).

I still think basing this on my earlier patch (in #161) would be preferable since the mappings should be in effect for all modes which do not take input directly. For example your patch does not modify operator pending mode hence motions/text object after an operator will not be affected. Also your implementation is likely trying to be too clever with the shifting and related stuff. Why don't you just use a loop and possibly the [ISUTF8 macro](https://github.com/martanne/vis/blob/master/util.h#L9).

> I've tried to check my code for memory leaks using valgrind, but encountered a leak in original cmd_map function

Yes this is to be expected at the moment. The "problem" is that the default mappings are located in the text/data section of the ELF file. Hence not all mappings can simply be freed, their origin would have to be tracked. I didn't bother to do this yet. It shouldn't be a big problem unless it also affects window local mappings which would occur more often (for instance via the prompt window).

In general patches to fix memory leaks are welcome.

martanne:
> Seems like ISUTF8 macro a bit broken. It returns true in all cases except non-leading UTF8 character.

That is intended, use it to move from the start of one UTF-8 character to the next. Something like (not tested):

```C
while (*s) {
   start = s;
   do s++; while (!ISUTF(s));
   end = s;
   /* your character goes from start to end */
}
```

> Well, I've added operator-pending and visual-line modes

Yes, but this wastes memory and just feels wrong. Also you do not need to add it to visual-line mode as it is a child mode of character wise visual mode, hence its bindings apply.

xomachine:
> this wastes memory and just feels wrong

That's true. I'll close PR for a while until a patch from #161 is not applied and reopen it (or open a new one) as soon as adapt [cmd_langmap](https://github.com/xomachine/vis/blob/master/vis-cmds.c#L1071) function to fill `vis->keymap`. The [cmd_map](https://github.com/martanne/vis/blob/master/vis-cmds.c#L1033) function will be restored.

PR #175, opened 2016-02-17
Display NUL bytes correctly

rgburke:
Marc pointed out in #173 that there is an issue with displaying NUL bytes.

This commit updates vis to specify the NUL byte column width as 2, so that vis treats NUL bytes like other non-printable characters.

Also fixed some typos in the Cell struct comments.


martanne:
Thanks, these bug fixes and code cleanups are appreciated!

In general the display related code is probably the ugliest part of vis. In the future I would like to investigate ways to move more of it into Lua.

PR #178, opened 2016-02-18
langmap command implementation

xomachine:
Refered to #161 and #174

It's realization of `:langmap` command adapted for [vis->keymap](https://github.com/xomachine/vis/blob/master/vis-core.h#L158) filling with short description in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,7 @@ Operators can be forced to work line wise by specifying `V`.
     :unmap        remove a global key mapping
     :map-window   add a window local key mapping
     :unmap-window remove a window local key mapping
+    :langmap      set key equivalents for layout specific key mappings
     :!            filter range through external command
     :|            pipe range to external command and display output in a new window
     :set          set the options below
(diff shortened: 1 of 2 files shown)


martanne:
Thanks, I still think the code could be simplified by avoiding the whole shifting business. Does the following also work?

```C
static bool cmd_langmap(Vis *vis, Filerange *range, enum CmdOpt opt, const char *argv[]) {
	const char *latin = argv[1];
	const char *nonlatin = argv[2];
	bool mapped = true;

	if (!latin || !nonlatin) {
		vis_info_show(vis, "usage: langmap <latin keys> <non-latin keys>");
		return false;
	}

	while (*latin && *nonlatin) {
		size_t i = 0, j = 0;
		char latin_key[8], nonlatin_key[8];
		do {
			if (i < sizeof(latin_key)-1)
				latin_key[i++] = *latin;
			latin++;
		} while (!ISUTF8(*latin));
		do {
			if (j < sizeof(nonlatin_key)-1)
				nonlatin_key[j++] = *nonlatin;
			nonlatin++;
		} while (!ISUTF8(*nonlatin));
		latin_key[i] = '\0';
		nonlatin_key[j] = '\0';
		mapped &= vis_keymap_add(vis, nonlatin_key, strdup(latin_key));
	}

	return mapped;
}
```

Also why did you chose the format `langmap <latin keys> <non-latin keys>` instead of the other way around `langmap <non-latin keys> <latin keys>` which seems more natural and similar to what vim does?

xomachine:
Yes, it does. 

Well, the `langmap <latin keys> <non-latin keys>` format seemed more natural  to me before I read your comment.

martanne:
It seems more natural since it is similar to `:map` where the left-hand side (1st argumnet) gets mapped to the right-hand side (2nd argument).

Thanks, merged!

PR #183, opened 2016-02-18
added operator tests

erf:
Added `operator` tests. 

Added new issues based on failed tests.


erf:
Removed visual mode from tests

.gitignore
```diff
@@ -4,3 +4,4 @@
 *.o
 /vis
 /dependency
+.DS_Store
```
twe4ked: Generally a good idea to put systems specific ignores in your global git ignore (`~/.gitignore`).

erf: Thanks, i will delete that gitignore line.

PR #184, opened 2016-02-19
update lexer for Arch Linux PKGBUILD

eworm-de:


eworm-de:
Do you want it this way or are there any upstream sources you sync with?

martanne:
The [Scintillua mercurial repository](http://foicica.com/hg/scintillua/) is the upstream source. I plan do periodically sync with them i.e. send our changes upstream and merge theirs back. But you can also send them directly upstream if you want ...

PR #185, opened 2016-02-19
colorcolumn enhancement

rgburke:
Currently the `colorcolumn` value must be less than or equal to the view width to be highlighted. This means lines which wrap do not have their `colorcolumn` highlighted.

This change updates vis to work with a `colorcolumn` greater than the view width, allowing lines which wrap to have their `colorcolumn` highlighted. This is consistent with vim's treatment of the `colorcolumn`.

### DIFF
--- a/view.c
+++ b/view.c
@@ -633,11 +633,27 @@ void view_update(View *view) {
 
 	view_syntax_color(view);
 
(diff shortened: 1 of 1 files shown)


martanne:
Thanks, merged!

PR #188, opened 2016-02-21
vim operator tests

erf:
Added vim operator tests. The test branch need to be updated for the `swap` test to run ok.


martanne:
Ok, I have merged the tests, without really taking a closer look. Some general comments.

You can use white spaces and C style comments in your tests (they are piped through the C pre-processor). This might improve readability, illustrating what you are attempting to do. See the [prompt test](https://github.com/martanne/vis/blob/test/test/vim/prompt/history.keys) for an example. If you need an actual space as test input use `<Space>`.

In general writing good tests is non-trivial. The current ones do not really have good coverage. As an example for yank/put I would assume to at least test:

* the default register
* a named register `"a`
* appending to a named register `"A`
* black hole register `"_`
* yank register `"0`
* testing each put variant `p`, `P`, `gP`, `gp` with a count
* repetition with the dot command `.` (this will likely reveal bugs ...)
* visual mode selection replacement
* ...

erf:
Thanks for the feedback. Will look into improving the tests.

PR #191, opened 2016-02-21
Test yank

erf:
I tried to write a more complex test for yank, but there are a couple of issues:
- i seem to have to write double quotes like `'\"'` in the `test.keys`
- the test always fails when i start with an empty `test.in`
- paste to empty line behaviour is different than in vim ( see `paste-to-empty-line test`), causing tests to fail


martanne:
Yes the testing infrastructure is a bit shaky ... A part of writing tests is to improve it along the way.

> i seem to have to write double quotes like '\"' in the test.keys

This should now work without escaping, as long as you only have one double quote per line? Not sure if using the C pre-processor is such a good idea after all ...

> the test always fails when i start with an empty test.in

I couldn't reproduce this, will take another look later on.

> paste to empty line behaviour is different than in vim

Yes for now tests should not rely on this behavior.

Because I'm constantely rebasing the test branch (whether that is a good workflow is a topic for another discussion), you will have to do the same. I believe you can just rebase your changes and force push your updates and Github should cope with it. I.e. there is no need to close and reopen new pull requests.

Issue #199, opened 2016-03-07
Visual bug while editing of the text containing multibyte characters


xomachine:
<img src="http://i.imgur.com/Tef7Imm.gif">
As you can see from the [gif](http://imgur.com/Tef7Imm), inserting symbols to the middle of text causes removing symbols at the end. Removing of inserted symbols restores previously deleted tail.

Versions affected: 

- vis v0.1.r155.gfbdc71b, compiled Feb 22 2016 20:07:05
- vis v0.1.r166.g73dda18, compiled Mar  7 2016 22:17:27

~~Now I cann't provide information for reproduction nor information about newer versions for a while. If I dig something out, I'll post it here.~~

I figured out that bug occurs only when text contains multibyte UTF characters and after ~2185'th twobyte character (characters other than twobyte were deleted in this test). Text area for the test: 91x48 (excluding borders and statusbar)

Using [this link](https://yadi.sk/d/UVypPMIypyvWF) you can download a sample text for reproducing the bug.

xomachine:
Fixed in 5cec5b27e258bc8ef61db9adc14707c80a3c7d8f

PR #201, opened 2016-03-08
Fix for visual bug with view size when text contains multibyte characters.

xomachine:
Workaround for #199

4-bytes character used as longest according to [RFC3629](https://tools.ietf.org/html/rfc3629#page-4)


martanne:
Thanks, I merged it. Eventually the display related code will need an overhaul.

PR #203, opened 2016-03-21
Configure

eworm-de:
Source and object code has to be specified before linker options.


eworm-de:
Looks like I missed some cases that did not break build...
Will add some more commits soon.

eworm-de:
Ok, should be fine now...

PR #204, opened 2016-03-21
configure: fix install prefix for man and share directories

eworm-de:


configure
```diff
@@ -420,8 +426,8 @@ srcdir = $srcdir
 prefix = $prefix
 exec_prefix = $exec_prefix
 bindir = $bindir
-MANPREFIX = $prefix/share/man
-SHAREPREFIX = $prefix/share/vis
+mandir = $mandir
+sharedir = $sharedir
```
martanne: But with this change `$MANPREFIX` and `$SHAREPREFIX` as used by the Makefile will once again be unset.

eworm-de: Ah, right... Just tested half way. :-p
Updated.

PR #208, opened 2016-03-23
visrc: Move filetype detection to separate function

joshaw:
I hadn't noticed the `return;` line when a filetype match is found so couldn't get my additions to this function to work. Moving the detection to a separate function keeps that isolated and indicates that other functions can be called from the same `vis.events.win_open` hook.

### DIFF
--- a/visrc.lua
+++ b/visrc.lua
@@ -67,8 +67,7 @@ vis:textobject_new("ii", function(win, pos)
 	return pos, pos
 end)
 
-vis.events.win_open = function(win)
-
+set_filetype = function(win)
 	local files = {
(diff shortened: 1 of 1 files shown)


martanne:
I will apply this later. File type detection in general could be improved quite a bit. At the moment it is solely based on the file extension which does not work for a number of files (e.g. shell scripts). By now the Lua API actually supports reading file content. Hence we could start detecting the file type based on shebang lines etc.

Also I'm not sure whether having this static data local inside a function (instead of assigned to a global variable) is such a good idea. The array will be repeatedly created and garbage collected for every window which is created.

martanne:
Thanks, I merged your changes with some slight tweaks. In the future please wrap your commit messages at a reasonable line length (for example by using `fmt(1)` or vis built in `=` operator).

PR #209, opened 2016-03-23
standalone build

eworm-de:


eworm-de:
Added a second commit that fixes dependencies for standalone build. (At least for me, please verify.)
Fixes #210

martanne:
I would like to avoid a `config.def.mk` file if possible. The necessary variables should instead be added to `Makefile` where they could be conditionally initialized (i.e. only if not already set) as it was done in the "old" build system

```
PREFIX ?= /usr/local
MANPREFIX ?= ${PREFIX}/share/man
SHAREPREFIX ?= ${PREFIX}/share/vis
```

Also currently we have the problem that in configure we use `$prefix` (lowercase) but in the Makefile we use `$PREFIX` thus the binary is installed in the wrong place (i.e. `$PREFIX is always empty). We should pick one variant and stick to it ...

GNUmakefile
```diff
@@ -87,7 +87,7 @@ dependency/sources/extract-libtermkey: dependency/sources/$(LIBTERMKEY).tar.gz
 	tar xzf $< -C $(dir $<)
 	touch $@
 
-dependency/sources/build-libtermkey: dependency/sources/extract-libtermkey
+dependency/sources/build-libtermkey: dependency/sources/extract-libtermkey dependency/sources/install-libncurses
```
martanne: This makes sense because libtermkey depends on curses. However the other change seems wrong, why should the lua build depend on curses?

eworm-de: Because the build instructions make lua  link against ncurses...
Updating patches, stay tuned.

eworm-de: Does 8d93f53e make sense?

martanne: Yes, thanks.

eworm-de:
Removed the first commit and opened #211.

martanne:
Thanks, merged (although I messed up the merge commit when rebasing something unrelated). In the GNU world they seem to prefer lower case variables, but I don't really care.

martanne:
Actually the above change to let libtermkey depend on the curses build only makes sense for the `standalone` target. In the case of `make local` (as for example used by the coverity static analyzer build) we want to use the system curses install.

A `touch dependency/sources/install-libncurses` at the right place should fix this, right?

eworm-de:
Ah, I did not know about `make local`. Probably a touch is sufficient...
Though this could break if you run `make local` followed by `make standalone`. :-p What level of perfection do we expect?

martanne:
> Though this could break if you run make local followed by make standalone

This is already broken because parts of the libraries would be linked against musl and parts against the system libc. Basically you have to rebuild from scratch when switching between the two.

martanne:
It is not pretty but `make local` should now be fixed. Also the two targets `local` and `standalone` can now be used in any order, the relevant libraries will be rebuilt from scratch.

PR #212, opened 2016-03-24
make sure build directory exists

eworm-de:


eworm-de:
Oh, this does not work... Let me think about it.

martanne:
The build directory should be created by the [dependency/sources](https://github.com/martanne/vis/blob/62ca03b94f09ddedc042bd5ac7bfa79215d7c380/GNUmakefile#L37) target, but apparently it does not work for you?

As for the `&&` no particular reason, feel free to clean it up.

GNUmakefile
```diff
@@ -62,7 +65,7 @@ dependency/sources/ncurses-%: | dependency/sources
 	mv $@.part $@
 	[ -z $(LIBNCURSES_SHA1) ] || (echo '$(LIBNCURSES_SHA1)  $@' | sha1sum -c)
 
-dependency/build/libncurses-extract: dependency/sources/$(LIBNCURSES).tar.gz
+dependency/build/libncurses-extract: dependency/build /dependencysources/$(LIBNCURSES).tar.gz
```
martanne: typo, misplaced `/`

eworm-de:
It works if I clone the git repository and build from a clean tree.

When I package this I do specify the source files in [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=vis-standalone) and place things where needed. Make does not download and dependency/sources target is not run.

Latest commit should be fine. ;)

martanne:
Thanks, hopefully this concludes the build system saga. I think it is ready for a release now ...

eworm-de:
Perhaps one last change... Just discovered that things break if wget is not installed in build chroot.
Want to commit b827a592?

martanne:
The dependencies for the build directory should actually be order only. Does your stuff still work?

PR #213, opened 2016-03-24
handle patch as usual source

eworm-de:


eworm-de:
Oh damn... It's not my day. Will have a look.

eworm-de:
Not sure whether or not you noticed that I updated the commit.
Should work now.

PR #217, opened 2016-03-29
Readme

eworm-de:


martanne:
Thanks, other distribution packages I'm aware of:

* [Void Linux](https://github.com/voidlinux/void-packages/tree/master/srcpkgs/vis) not sure whether that is the best link
* [Alpine Linux](https://pkgs.alpinelinux.org/packages?name=vis) currently outdated
* [pkgsrc-wip](http://pkgsrc.se/wip/vis-editor) currently outdated

Not sure whether it makes sense to link to outdated distribution packages ...

PR #227, opened 2016-04-07
Compile translation units to object code

matze:
Adding dependencies on object code rather than C translation units has two benefits: 1) if a translation has changed, only that needs to be recompiled instead of all and 2) parallel make (using -j) is possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ SRC = array.c buffer.c libutf.c main.c map.c register.c ring-buffer.c \
 	ui-curses.c view.c vis.c vis-lua.c vis-modes.c vis-motions.c \
 	vis-operators.c vis-prompt.c vis-text-objects.c
 
+OBJ = $(patsubst %.c,%.o,$(SRC))
+
 # conditionally initialized, this is needed for standalone build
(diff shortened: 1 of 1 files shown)


martanne:
Thanks for your contribution! Are build times an issue for your?

I prefer the current behavior because:

* the Makefile is intended to be POSIX compatible or at least work with both GNU and BSD make, I'm not sure `patsubst` is supported by the latter?
* it is a kind of sanity check regarding feature bloat. The build takes about 7 seconds on my ~8 year old non-SSD laptop. That seems fast enough for me. On modern machines it should be considerably faster.
* there is no `*.o` clutter in the source directory
* it might allow the compiler to perform some whole program optimizations

Some of these things were also discussed in [an old issue](https://github.com/martanne/vis/issues/9#issuecomment-68842228).

matze:
> Are build times an issue for you?

Yes. 7 seconds is right between immediate testing and I-can-have-a-coffee. In fact, it's pretty annoying if modifying only minor parts (i.e. printf-debugging). But I see, that you're reluctant to these kind of changes, so …

PR #228, opened 2016-04-07
Improve help command.

xihh87:
By sorting the keybindings and making explicit the space we make the `:help` command more useful.


vis-cmds.c
```diff
@@ -470,7 +470,7 @@ static bool print_keybinding(const char *key, void *value, void *data) {
 	const char *desc = binding->alias;
 	if (!desc && binding->action)
 		desc = binding->action->help;
-	return text_appendf(txt, "  %-15s\t%s\n", key, desc ? desc : "");
+	return text_appendf(txt, "  %-15s\t%s\n", isspace(key[0]) ? "<Space>" : key, desc ? desc : "");
```
martanne: Thanks, this should be `key[0] == ' ' ? ...`

I can amend it while pulling, if you want.

xihh87: Ammended on my branch, thanks.

PR #240, opened 2016-04-08
Use default instead of black, white in dark-16 theme

larryhynes:
This is probably a matter of taste - if not a source of anguish! - so feel free to ignore this PR.

I've tested `back:default` and `fore:default`, instead of `back:black` and `fore:white` in both `xterm` and `urxvt` and it 'just works' or, at least, works the way that I would expect it to. (I'm not sure how much of it 'working' is dependent on my local set-up `.Xresources`, etc., though.)

Here's the default `dark-16.lua`...

screen shot 2016-04-08 at 16 49 30

And my local changes, as proposed here...

screen shot 2016-04-08 at 16 49 52

### DIFF
--- a/lexers/themes/dark-16.lua
+++ b/lexers/themes/dark-16.lua
@@ -1,8 +1,8 @@
 -- Eight-color scheme
 local lexers = vis.lexers
 -- dark
-lexers.STYLE_DEFAULT = 'back:black,fore:white'
-lexers.STYLE_NOTHING = 'back:black'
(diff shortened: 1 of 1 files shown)


martanne:
Not sure about this. IIRC one idea behind the dark theme was also to be dark in a terminal which by default has a white background ...

larryhynes:
Yeah, no problem, I think I was scratching a personal itch with this!

PR #241, opened 2016-04-08
Add section on visrc.lua to README.md

larryhynes:


martanne:
Thanks! This is most welcome, however it should probably also mention `$VIS_PATH` and maybe list all directories considered according to their priority as is done in the [manual page](https://github.com/martanne/vis/blob/master/vis.1).

larryhynes:
No worries - thanks yourself!! :D Have a look at the revised changes and see what you think.

larryhynes:
Hmmm. I'm wondering if this section should include mention of `config.def.h`?

PR #253, opened 2016-04-11
Statusline additional information

joshaw:
I've found having the total line count and the percentage position in the file useful while navigating around and since the code is protected by the `UI_OPTION_LARGE_FILE` setting, it should not be too costly.

My apologies for any C blunders, I'm not much of a programmer but have tried my best to be consistent.

### DIFF
--- a/text.c
+++ b/text.c
@@ -1448,10 +1448,6 @@ size_t text_size(Text *txt) {
 	return txt->size;
 }
 
-size_t text_lines(Text *txt) {
-	return lines_count(txt, 0, text_size(txt));
-}
-
 /* count the number of new lines '\n' in range [pos, pos+len) */
 static size_t lines_count(Text *txt, size_t pos, size_t len) {
 	size_t lines = 0;
--- a/text.c
+++ b/text.c
@@ -1448,6 +1448,10 @@ size_t text_size(Text *txt) {
 	return txt->size;
 }
 
+size_t text_lines(Text *txt) {
+	return lines_count(txt, 0, text_size(txt));
+}
+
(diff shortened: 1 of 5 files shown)


martanne:
I'm not sure about this. I would prefer if the default settings would not clutter the status bar further. This is probably another thing which at some point should be made configurable via Lua. Also the line based `%` indication might be confusing, because the `50%` motion will act upon bytes not lines.

joshaw:
I do understand entirely. 

I realised that the `%` operator would be inconsistent so just changed the indicator to use the byte offset. This way, also, the line number is not used. Perhaps, just the percentage could be shown and I'll remove the total line number - that way its not adding as much, doesn't need to calculate any line numbers but does give indication of position?

martanne:
With the latest changes to the Lua API the following should now work:

```lua
        local win = vis.win;
        local mode = ""
        if vis.mode == vis.MODE_INSERT then
                mode = "--INSERT--"
        elseif vis.mode == vis.MODE_REPLACE then
                mode = "--REPLACE--"
        elseif vis.mode == vis.MODE_VISUAL then
                mode = "--VISUAL--"
        elseif vis.mode == vis.MODE_VISUAL_LINE then
                mode = "--VISUAL LINE--"
        end
        print(mode)
        print(win.file.name)
        print(win.cursor.number..'/'..#win.cursors)
        print(win.cursor.line..', '..win.cursor.col)
        print(math.ceil(win.cursor.pos/win.file.size*100)..'%')
        print(win.file.newlines)
```

what is currently missing is a way to put the information into the status bar ...

PR #255, opened 2016-04-11
Add block motions from vim, [{, ]}, [( and ])

bobrippling:


text-motions.c
```diff
@@ -586,25 +586,37 @@ size_t text_function_end_prev(Text *txt, size_t pos) {
...
+static size_t text_block_direction(Text *txt, size_t pos, int direction) {
 	int nest = 0;
-	char c;
+	char c, target, nest_target;
 	bool instring = false;
 	Iterator it = text_iterator_get(txt, pos);
-	while (text_iterator_byte_prev(&it, &c)) {
+
+	target = (direction > 0 ? '}' : '{');
+	nest_target = (direction > 0 ? '{' : '}');
```
kopischke: Considering that `vis` is not trying to be compatible with Vim’s bugs and limitations, does it really make send
sense to add hard wired motions that only apply to languages with a C-like syntax? I think this kind of, for lack of a better word, semantic motions would be better if they tied in with the lexing system, which would also deliver non-naive handling of blocks to be ignored (comments, here doc, strings with all kinds of quotes, interpolation, misc. and sundry).

martanne:
Thanks for your contribution! It generally looks good, but I believe the implementation in `text-motions.c` could be simplified by reusing the code used for text objects. As an example the block start motion would become:

```
size_t text_block_start(Text *txt, size_t pos) {
        Filerange r = text_object_curly_bracket(txt, pos);
        return text_range_valid(&r) ? r.start : pos;
}
```

text-motions.c
martanne: While I agree that the current approach has the drawbacks you mention, it works for the common case, is reasonably simple and already exists. It is currently used to:

* highlight matching symbols at the cursor location
* implement the `%` motion when on a special symbol
* implement text objects for blocks enclosed by `{`, `(`, `[` etc.

It will also work in builds without Lua support (whether we want to support these is another question).

This pull requests simply hooks up a few more motions to already existing code. Adding more "semantic" versions of these motions and text objects is a worthwhile, but separate and not completely trivial, goal which should not block this.

Also I might be slightly biased, because these days I primarily work in C like languages.

bobrippling:
Thanks for the prompts - I wasn't too sure if the existing bracket matching code supported this, I'll see if I can make those changes this evening.

text-motions.c
bobrippling: Additionally, I find even when writing plain text, being able to jump to the start/end of a `()` pair is pretty handy and I don't have to context switch, or put the editor into a C-language mode.

(although admittedly you could turn on the `(` block matching for all syntaxes)

kopischke: I don’t have any issues with the parens motion, which is useful in general enough contexts, but a “block” motion only applicable to a subset of contexts is one of those Vim legacy misfeatures that got me interested in alternate implementations in the first place.

martanne:
I slightly tweaked the implementation, let me know if these changes broke anything.

PR #259, opened 2016-04-12
implement the vim-like 'gq' operator

tycho:
This is used to line-wrap text based on the 'textwidth' option value.

Note that normally 'textwidth' would have the alias 'tw', but that's currently taken by 'tabwidth'... Might want to normalize 'tabwidth' to 'tabstop' to match vim's naming.

This hasn't undergone particularly extensive testing, aside from verifying that the line wrapping results matched vim's behavior for the cases I tried. I tested using visual line-selection with multiple paragraphs, single paragraph, etc. There are probably other cases I should try, but this works for the common usecases I have.

### DIFF
--- a/vis-operators.c
+++ b/vis-operators.c
@@ -181,6 +181,36 @@ static size_t op_cursor(Vis *vis, Text *txt, OperatorContext *c) {
 	return EPOS;
 }
 
+static size_t op_wrap_text(Vis *vis, Text *txt, OperatorContext *c) {
+	size_t pos = text_line_prev(txt, c->range.start), prev_pos;
+
+	if (vis->textwidth < 1)
+		return c->range.start;
+
+	do {
+		prev_pos = pos;
+		size_t end = text_line_next(txt, pos);
+		while (pos - prev_pos < vis->textwidth && end > pos)
+			pos = text_longword_end_next(txt, pos);
+		if (pos - prev_pos >= vis->textwidth) {
+			size_t break_pos = text_longword_start_prev(txt, pos);
+			if (break_pos == prev_pos) {
+				/* vis->textwidth is longer than this word, insert line break
+				 * immediately after this word */
+				break_pos = pos + 1;
+			} else {
+				/* found an appropriate line break point just before this word */
+				break_pos = break_pos - 1;
+			}
+			text_delete(txt, break_pos, 1);
+			text_insert(txt, break_pos, "\n", 1);
+			pos = break_pos + 1;
+		}
+	} while (pos != prev_pos && pos < c->range.end);
+
+	return c->range.end;
+}
+
 static size_t op_join(Vis *vis, Text *txt, OperatorContext *c) {
 	size_t pos = text_line_begin(txt, c->range.end), prev_pos;
 
@@ -302,4 +332,5 @@ const Operator vis_operators[] = {
 	[VIS_OP_REPLACE]     = { op_replace     },
 	[VIS_OP_CURSOR_SOL]  = { op_cursor      },
 	[VIS_OP_FILTER]      = { op_filter      },
+	[VIS_OP_WRAP_TEXT]   = { op_wrap_text   },
 };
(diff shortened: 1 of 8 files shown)


vis-operators.c
```diff
@@ -181,6 +181,33 @@ static size_t op_cursor(Vis *vis, Text *txt, OperatorContext *c) {
...
+		prev_pos = pos;
+		size_t end = text_line_next(txt, pos);
+		while (pos - prev_pos < vis->textwidth && end > pos)
+			pos = text_longword_end_next(txt, pos);
+		if (pos - prev_pos >= vis->textwidth) {
+			size_t break_pos = text_longword_start_prev(txt, pos);
+			if (break_pos == prev_pos) {
+				/* vis->textwidth is longer than this word, insert line break
+				 * immediately after this word */
+				break_pos = pos + 1;
```
tycho: Not sure if these direct addition/subtractions to the position are correct. It _looks_ like text_longword_* guarantees that the adjacent character is always whitespace, so this should be OK, but I may have missed something. I'm mostly concerned about jumping into the middle of a unicode character, breaking the encoding.

Might want to use the text iterators here?

tycho:
Bah, one case that doesn't work is if the text is already wrapped at a different column. e.g. if the text is wrapped at column 60, and you try to 'gq' with textwidth=80, nothing changes. Probably need to eat single newlines as they're encountered if the line width isn't met.

tycho:
Alright, I fixed that case. We will now eat newlines **unless** we see that the next line is at the same position as the next *empty* line. So we will preserve existing paragraphs, but reflow them according to the new textwidth.

martanne:
I haven't yet taken a closer look at the code, but what is wrong with making `gq` an alias for `=`? I guess currently you can't set the text width, this could be added though ...

tycho:
Thanks for pointing that out, I actually didn't realize `=` existed in vis. At least this was an interesting coding exercise, even if it's not what ends up getting used. :)

I guess the only downside of `=` is that it depends on coreutils. But is that even a downside? Every *sane* environment I know of has it, and it does seem wrong to duplicate the effort when it's not needed. I'd hazard a guess that fmt is probably more efficient, too.

Looking at vim's documentation, I see that `=` will either invoke the external program specified in option `equalprg` or use the vim-internal formatting (this pull request could be somewhat of an implementation of the latter). I also note that it doesn't seem to suggest that `gq` and `=` are equivalent: I wonder what the distinction is in vim's case?

What would you like to do here?

kopischke:
`={motion}` and `gq{motion}` are conceptually different in Vim: the former nominally is the _filter_ operator and uses `'equalprg'`, the latter nominally is the _format_ operator and uses either `'formatexpr'` or an external '`formatprg'`. Both have an internal fallback, which, in practice, makes the former the `indent` operator and the latter the `reflow` operator (see the pertinent help sections).

tycho:
Interesting. That's a pretty good case against aliasing them. I suppose it depends on how closely we want to follow vim's behavior. I'm sure there are differing opinions, but I personally wouldn't mind having vis be a drop-in replacement for vim, with respect to all basic editing. The README.md seems to agree with me on that.

The current behavior of vis' `=` sounds pretty much like what vim's `gq` is. Perhaps `=` should be renamed `gq` and learn the `textwidth` option.

martanne:
My main point is that these kind of things are best handled by a standalone tool outside the editor.

If we want to more closely follow vim then:
 * `=` should use [clang-format](http://clang.llvm.org/docs/ClangFormat.html) or `indent(1)` or some other language specific tool. However I do not know how well these tools work when given only partial source code? For plain text files this should default to `fmt(1)` 
 * `ga` should always default to `fmt(1)`

Because up until now `ga` isn't supported at all and `=` is easier to type, I chose to use the latter.

martanne:
For now I aliased `=` and `gq` to the existing `fmt(1)` based functionality. I hope this compromise is good enough, it covers my use cases (formatting mails / commit messages etc).

PR #260, opened 2016-04-12
Sregex

jnsquire:
Hi there!

I like what you're doing with this project. The regex issue bothered me, and I did a little digging and found the sregex library (used by nginx) that already has support for streaming regex matching. I've got a working set of changes here to use this library instead of the system regex code, no copying of text blocks required. I hope you like it!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
 -include config.mk
 
+REGEX_SRC ?= text-regex.c
+
 SRC = array.c buffer.c libutf.c main.c map.c register.c ring-buffer.c \
 	sam.c text.c text-motions.c text-objects.c text-util.c \
 	ui-curses.c view.c vis.c vis-lua.c vis-modes.c vis-motions.c \
-	vis-operators.c vis-prompt.c vis-text-objects.c
+	vis-operators.c vis-prompt.c vis-text-objects.c $(REGEX_SRC)
 
 # conditionally initialized, this is needed for standalone build
 # with empty config.mk
@@ -14,15 +16,8 @@ SHAREPREFIX ?= ${PREFIX}/share/vis
 VERSION = $(shell git describe 2>/dev/null || echo "0.2")
 
 CONFIG_LUA ?= 1
-CONFIG_SREGEX ?= 0
 CONFIG_ACL ?= 0
 CONFIG_SELINUX ?= 0
-
-ifeq ($(CONFIG_SREGEX),1)
-SRC += text-sregex.c
-else
-SRC += text-regex.c
-endif
 
 CFLAGS_STD ?= -std=c99 -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -DNDEBUG
 LDFLAGS_STD ?= -lc
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 -include config.mk
 
 SRC = array.c buffer.c libutf.c main.c map.c register.c ring-buffer.c \
-	sam.c text.c text-motions.c text-objects.c text-regex.c text-util.c \
+	sam.c text.c text-motions.c text-objects.c text-util.c \
 	ui-curses.c view.c vis.c vis-lua.c vis-modes.c vis-motions.c \
 	vis-operators.c vis-prompt.c vis-text-objects.c
 
@@ -14,14 +14,21 @@ SHAREPREFIX ?= ${PREFIX}/share/vis
 VERSION = $(shell git describe 2>/dev/null || echo "0.2")
 
 CONFIG_LUA ?= 1
+CONFIG_SREGEX ?= 0
 CONFIG_ACL ?= 0
 CONFIG_SELINUX ?= 0
+
+ifeq ($(CONFIG_SREGEX),1)
+SRC += text-sregex.c
+else
+SRC += text-regex.c
+endif
 
 CFLAGS_STD ?= -std=c99 -D_POSIX_C_SOURCE=200809L -D_XOPEN_SOURCE=700 -DNDEBUG
 LDFLAGS_STD ?= -lc
 
 CFLAGS_VIS = $(CFLAGS_AUTO) $(CFLAGS_TERMKEY) $(CFLAGS_CURSES) $(CFLAGS_ACL) \
-	$(CFLAGS_SELINUX) $(CFLAGS_LUA) $(CFLAGS_STD)
+	$(CFLAGS_SELINUX) $(CFLAGS_SREGEX) $(CFLAGS_LUA) $(CFLAGS_STD)
 
 CFLAGS_VIS += -DVERSION=\"${VERSION}\"
 CFLAGS_VIS += -DCONFIG_LUA=${CONFIG_LUA}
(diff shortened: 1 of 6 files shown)


martanne:
Thanks, this looks interesting! I vaguely remember having looked at sregex before.

I will add some inline comments and hope to try it out later this evening. One thing I noticed is that you used 4 spaces for indentation, whereas the rest of vis uses tabs. Also `make debug` results in some warnings, these should be fixed.

Makefile
```diff
@@ -14,22 +14,29 @@ SHAREPREFIX ?= ${PREFIX}/share/vis
 VERSION = $(shell git describe 2>/dev/null || echo "0.2")
 
 CONFIG_LUA ?= 1
+CONFIG_SREGEX ?= 0
 CONFIG_ACL ?= 0
 CONFIG_SELINUX ?= 0
 
+ifeq ($(CONFIG_SREGEX),1)
```
martanne: The Makefile (as opposed to GNUmakefile) tries to only rely only on POSIX features, if possible this should be done in some other way.

configure
```diff
@@ -304,6 +304,31 @@ else
...
+    return 0;
+}
+EOF
+
+CONFIG_SREGEX=0
+
+printf "checking for sregex... "
+
+CFLAGS_SREGEX="-I${PREFIX}/include/sregex"
+LDFLAGS_SREGEX="-L${PREFIX}/lib -Wl,-rpath=${PREFIX}/lib -lsregex"
```
martanne: So far we avoided specifying `-rpath` and instead advised users to set `$LD_LIBRARY_PATH` as required.

text-sregex.c
```diff
@@ -0,0 +1,146 @@
...
+    if (!r->mpool) goto err1;
+    return 0;
+
+err1:
+    sre_destroy_pool(r->pool);
+    return 1;
+}
+
+void text_regex_free(Regex *r) {
+    if (r) {
```
martanne: The general idiom in vis is to start these `*_free` functions with `if (!r) return;` to avoid one level of indentation. It isn't really that important here because the function is extremely short.

text-sregex.c
```diff
@@ -0,0 +1,146 @@
...
+            break;
+        }
+    }
+
+    destroy_context(re_ctx, r);
+    return ret;
+}
+
+int text_search_range_backward(Text *txt, size_t pos, size_t len, Regex *r, size_t nmatch, RegexMatch pmatch[], int eflags) {
+    sre_int_t *unused;
```
martanne: This variable seems really unused.

text-sregex.c
```diff
@@ -0,0 +1,146 @@
...
+            
+            ret = 0;
+            fill_match(r, nmatch, pmatch, it.pos);
+
+            if (r->captures[0] == 0 && r->captures[1] == 0) {
+                char b = 0;
+
+                do {
+                    text_iterator_byte_next(&it, &b);
+                } while (b != '\n');
```
martanne: This could be costly (although the whole branch should not be that common). This might justify a `text_iterator_line_next` function. I guess this was also the intention behind the `lines_skip_forward` renaming below?

text.c
```diff
@@ -150,7 +150,6 @@ static Action *action_alloc(Text *txt);
 static void action_free(Action *a);
 /* logical line counting cache */
 static void lineno_cache_invalidate(LineCache *cache);
-static size_t lines_skip_forward(Text *txt, size_t pos, size_t lines, size_t *lines_skiped);
```
martanne: `text_skip_forward` seems to be only used within text.c, hence there is no need to export it and the renaming is unnecessary.

jnsquire: Yeah, I forgot to revert that.

text-sregex.c
jnsquire: The regex engine fills in a value indicating how many of the submatches are valid when a match is only partially completed at the end of a text block. We don't have a use for that.

text-sregex.c
jnsquire: Right. In the end, it seemed like too uncommon a corner case, and the corner cases in text_iterator_line_next were tricky and I wasn't sure I'd gotten them all right.

jnsquire:
Thanks for the review, I've made the changes you suggest.

martanne:
There were a couple more issues, namely forward searches were not restricted to the given range and backward searches were completely broken? I fixed those basic issues and pushed the result to the [sregex branch](https://github.com/martanne/vis/tree/sregex).

On a more fundamental level sregex currently lacks two main features:
 * support for the equivalent of `REG_NEWLINE` semantics in libc, that is `.` should not match a new line, `^`/`$` should match the empty string immediately after/before a newline, see regex(3) for more info.
 * there doesn't seem to be a way to run the automaton/vm "backwards". I haven't yet looked at the underlying implementation, but sam used the following technique to support efficient backward searches:
```
The solution is to construct a machine identical to the machine for a forward
search except for a reversal of all the concatenation operators (the other
operators are symmetric under direction reversal), to exchange the meaning
of the operators ^ and $, and then to read the file backwards, looking for the
usual earliest longest match.
```
I'm not yet sure whether this is possible while staying Perl-compatible which seems to be a goal of sregex.

To summarize the current state `^`, `$` and `.` are broken with regard to new line handling, forward searches are fast, backward searches are slow.

jnsquire:
Yeah, I totally missed the effects of len being passed to text_string_alloc0. Thanks for the cleanup.

 I'll take a look at sregex to see how hard adding proper newline handling would be. True backward searching seems like it would be hairier to implement, but worth looking into.

jnsquire:
I've added REGEX_NEWLINE handling to https://github.com/jnsquire/sregex.git

As for backward searching, constructing the regex machine 'backwards' as described also seems doable without a lot of pain, but I think 'reading the file backwards' implies feeding the bytes of each text block into the matching function in reverse order as well, or adding a flag to the matching function to iterate input buffers back to front... doesn't seem pleasant either way.

martanne:
Thanks for working on this (I haven't yet had time to seriously look at the implementation).

There exist some [differences](http://regular-expressions.mobi/refflavors.html) between POSIX Extended Regular Expressions (POSIX-ERE) as currently used by vis and Perl Compatible Regular Expressions (PCRE) as targeted by sregex.

The [major one](http://regular-expressions.mobi/posix.html#alternation 
) being that POSIX specifies that the longest leftmost match shall be returned, whereas PCRE will stop at the first match. For a concrete example: using the pattern `one(self)?(selfsufficient)?` on the string `oneselfsufficient` with PCRE will result in matching `oneself`, but using POSIX the result will be the full string `oneselfsufficient`.

Now the question is which variant should be used in vis? Traditionally the regex variant supported by [Plan 9](https://swtch.com/plan9port/man/man7/regexp.html)/sam are closer to POSIX than PCRE. Having the editor behave differently depending on whether it is compiled against sregex (which should eventually become the default) doesn't seem like a good idea.

As I see it there are 2 phases in making sregex work for vis:

1. Add support for an equivalent of `REG_NEWLINE`, `REG_NOTBOL`, `REG_NOTEOL` (as you seemingly already started). This should not be too complicated and ensure basic functionality. It should already be a huge improvement for searching in large files, because the needless copying of data would be avoided.
2. Implement full [POSIX-ERE compatible](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html) semantics. Implementing the few wrappers required for the [`regex(3)`](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/regex.h.html)  interface would also facilitate testing and benchmarking against other implementations. For the former the [regex testsuite from AT&T Research](http://web.archive.org/web/20130915160837/http://www2.research.att.com/~gsf/testregex/testregex.html) as recently used by [DragonFlyBSD](
http://gitweb.dragonflybsd.org/dragonfly.git/commitdiff/abce74f49c2c19b069958a0b48de0a9987d14e35) might be a good start.

As for matching backwards, the interface would be almost the same, you would still pass individual junks (`sre_char *input`, `size_t len`) to a new execute function, but internally it would iterate backwards i.e. from `input+len-1` to `input`. The `eof` parameter would become a `bof` (begin of file) etc.

Anyway maybe the main developer (@agentzh) of sregex would also like to share his thoughts on the matter.

Unfortunately I personally won't have much time to work on this during the next couple of months.

agentzh:
Well, as the author of sregex, I'm fine with adding POSIX semantics mode to sregex, as well as those modes to change the semantics of dot and `^`/`$`. Patches welcome to sregex :)

PR #265, opened 2016-04-13
Add lexer for taskpaper documents

larryhynes:
This is a 'first cut' of a lexer for taskpaper documents. Local testing looks good, but I'd like to get some feedback on it from use in the wild. Pending that, I'll submit it to upstream.

### DIFF
--- a/lexers/taskpaper.lua
+++ b/lexers/taskpaper.lua
@@ -1,0 +1,64 @@
+-- Taskpaper LPeg lexer
+-- Copyright (c) 2016 Larry Hynes
+-- See LICENSE
+-- Improvements or suggestions to larry-at-larryhynes.com
+-- Important Disclaimer:
+-- I have no idea what I'm doing
+
+local l = require('lexer')
+local token = l.token
+local P, R, S = lpeg.P, lpeg.R, lpeg.S
+
+local M = {_NAME = 'taskpaper'}
+
+local delimiter = P('    ') + P('\t')
+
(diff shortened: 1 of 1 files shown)


martanne:
Some general remarks:

* especially for more obscure lexers, it is generally a good idea to include a comment with an URL to the format specification (if such a thing exists at all). This makes both review and later modifications easier.
* you probably want to associate your lexer to a set of file extensions. Upstream will likely ask for this.
* I'm not sure whether hard coding specific colors is such a good idea

If you plan to submit it upstream, I will merge it when re-syncing.

larryhynes:
I couldn't find a canonical syntax reference. It's straightforward, but more text than code which made it a little slippery to pin down.

I meant to update `visrc.lua` with the `.taskpaper` reference, but somehow managed to misplace that!

I'll send it along to Mitchell at Scintillua and see what he reckons, then hopefully we can merge it in here directly from upstream.

Thanks - and keep up the great work. :smiley_cat:

TieDyedDevil:
On Apr 13, 2016 9:33 AM, "Marc André Tanner" <notifications@github.com>
wrote:

> I'm not sure whether hard coding specific colors is such a good idea

It is generally not a good idea to specify specific colors. Instead, please
adopt existing predefined "styles" (search for STYLE_ in lexers/lexer.lua).

larryhynes:
@TieDyedDevil Yep, that's done now locally, along with a few other little tweaks.

PR #266, opened 2016-04-13
Support zenity alongside slmenu and dmenu

Screwtapello:
Because zenity has a slightly different command-line interface than `slmenu` or `dmenu`, `vis-open` now has a wrapper function that converts `slmenu`-style arguments to what zenity expects.

zenity is rather different in style from `slmenu` and `dmenu`, but it's much more likely to be available. Compare the Debian installation statistics for [zenity](https://qa.debian.org/popcon.php?package=zenity) versus [suckless-tools](https://qa.debian.org/popcon.php?package=suckless-tools) (the Debian package containinng `dmenu`). So far as I can tell, Debian doesn't even *have* a package for `slmenu`.

I considered also trying to add support for `dialog` or `whiptail`, but their interfaces are even more wildly different.

### DIFF
--- a/vis-open
+++ b/vis-open
@@ -4,6 +4,30 @@ PATTERN="."
 [ -z "$VIS_MENU" ] && VIS_MENU="slmenu"
 [ -z "$VIS_MENU_ARGS" ] && VIS_MENU_ARGS="-b"
 
+zenity_wrapper () {
+	local ARGS
+
+	while [ $# -gt 0 ]; do
+		case "$1" in
+		-p)
+			ARGS="$ARGS --text $2"
+			shift
+			shift
+			;;
+		-b)
+			# Not relevant for zenity
+			shift
+			;;
+		*)
+			echo "Unhandled option $1" >&2
+			exit 1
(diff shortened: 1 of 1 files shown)


martanne:
I'm not sure I like this, `zenity` seems to have quite a different philosophy. If you want a more popular tool, maybe take a look at [`fzf`](https://github.com/junegunn/fzf) (I haven't actually tested whether it works with vis). We could also think about importing and distributing `slmenu` as a `vis-menu`.

Screwtapello:
Yeah, zenity is very different and no doubt suboptimal (although if you're working in a terminal and don't already know what to expect, having zenity pop up in the middle of the screen is probably slightly less surprising than having dmenu appear across the edge of your monitor), but I figured it's slightly better than things not working at all.

Bundling `slmenu` as `vis-menu` (or even as `vis-open`) would be a good alternative, if you don't object to adding more code.

martanne:
As discussed in IRC `slmenu` is the currently preferred tool for this.

PR #268, opened 2016-04-13
Add 16 color solarized theme

larryhynes:
This makes solarized play nice with xterm, urxvt and iterm2... 16 color terminals, basically.


larryhynes:
Fixed where I had used color '12' twice, and improved `cursorline` and `colorcolumn` rendering.

martanne:
Could you provide a screenshot with active selection? This does not seem to work at all here, there might be a problem with the color handling code though...

larryhynes:
Here's how it looks locally:

screen shot 2016-04-14 at 21 36 52

Bear in mind that this theme depends on your terminal being set up for Solarized. I had considered adding a commented example `.Xresources` snippet..

larryhynes:
Here's an nvim view of the same file, using the Solarized vim theme:

screen shot 2016-04-14 at 21 41 11

These are both taken in xterm, but look the same in urxvt and iterm2.

larryhynes:
I've brought this down to its bare essentials, in order to try and isolate why `STYLE_CURSOR_LINE` isn't  showing as `back:0`, but I haven't figured it out yet. I'll close this for now and reopen if I get anywhere with it.

PR #277, opened 2016-04-16
Added functionality to complete word

erf:
I added the functionality to complete the current word with `<C-n>` in `insert mode`. This is similar to the autocomplete behaviour in VIM, except i just pick the first match by searching from the start of the file. It also works with multiple cursors.


erf:
A nice addition to this would be to iterate and replace matches by continuously pressing `<C-n>`. This would however require `state`.  One for search position and one for mode - would this be a new `VIS_MODE_COMPLETE` etc.?

PR #288, opened 2016-04-22
Split line with K

erf:
Split line with `K`, as opposite to `J`. 
(as suggested here http://stackoverflow.com/questions/624821/vim-split-line-command)


erf:
Hmm.. I see my other commits snuk in there, but feel free to use this as a suggestion, and then close this PR.

martanne:
This sort of thing is probably best dealt with a custom key binding in your `visrc.lua`?

erf:
Yes maybe, not very biased either way - i thought it might be a nice addition to core if `K` is not used, to have a complementary, `split` to `join`.

Issue #210, opened 2016-03-23
`make standalone` fails for parallel build


eworm-de:
Run `make -j5 standalone` and see the build fail. I got a number of different errors as this is very racy...

martanne:
This was fixed by your latest pull request, or do you still encounter problems?

I noticed that the hardening flags picked up from a previous `./configure` run do not seem to work well with musl (I got a segfault in the relocation code). Maybe we should just build the dependencies and then as a last step run `./confgiure` with `CC=musl-gcc` and `{C,LD}FLAGS` set to point into the appropriate folder in `dependency/install/`?

eworm-de:
> This was fixed by your latest pull request, or do you still encounter problems?

No, everything is just fine for me.

> I noticed that the hardening flags picked up from a previous ./configure run do not seem to work well with musl (I got a segfault in the relocation code). Maybe we should just build the dependencies and then as a last step run ./confgiure with CC=musl-gcc and {C,LD}FLAGS set to point into the appropriate folder in dependency/install/?

Feel free to do so. ;)
Usually I build in a clean chroot, so this is not an issue for me. ;)

martanne:
> Usually I build in a clean chroot, so this is not an issue for me. ;)

Fair enough. In theory I should also be doing other stuff than fiddling with stupid build system issues :/

We should fix `make local` though...

martanne:
If you find the time could you please check that the latest changes did not break parallel builds?

Thanks!

eworm-de:
It breaks for me. When I place tarballs manually build breaks:
```
tar xzf dependency/sources/musl-1.1.14.tar.gz -C dependency/build/
cleaning
tar: dependency/build: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
GNUmakefile:45: recipe for target 'dependency/build/libmusl-extract' failed
```
Should be fixed by #212.

martanne:
I cleaned up the `make standalone` build as described in my [first comment](https://github.com/martanne/vis/issues/210#issuecomment-2003499459) to this issue.

A couple of questions if you don't mind trying it again:

1. does current git master work for you?
2. if yes, does it still work if you revert d6effd021d2aa2b6c452f229046a85c5f1a252a7 i.e. try to build with `-pie`?
3. assuming 2. fails, would you be willing to debug the issue? I suspect it is some bad relocation entries in the libtermkey build due to some libtool stupidity. However I'm not sure why it suddenly broke/why it worked before.

[...]

PR #289, opened 2016-04-22
Pie

eworm-de:
Should fix PIE standalone builds... Please give it a try.
Related to #210, d6effd0.


eworm-de:
Uh, this still does not work. Did my final testing on a system with hardening toolchain enabled - the build always succeeds there. Will give it another try soon.

PR #290, opened 2016-04-22
build: run configure before building dependencies

eworm-de:
We need some basic configuration before building dependencies. For
example useful CFLAGS have to be in place to prevent linker errors.

The configure script is run again later with some extra parameters.


eworm-de:
Ah, this breaks if configure does not find system libraries...
Can we make it stop before checking libraries but after creating config.mk with flags?

eworm-de:
This is still broken for some cases and work in progress. Please do not merge.

martanne:
Thanks for looking into this! I'm not too happy with the implementation, but with the latest commits and binutils >= 2.26 running `make standalone` seems to build a statically linked completely position independent executable. Can you confirm this i.e. does it work for you too?

It works by tweaking musl's gcc wrapper/spec file to use `rcrt1.o` and passing `--no-dynamic-linker` to the linker. Rich Felker has a [pending gcc patch](https://github.com/richfelker/musl-cross-make/blob/master/patches/gcc-5.2.0/0005-staticpie.diff) which should hopefully make this possible in the future with a simple `cc -fPIC -pie --static your-app.c`.

For those wondering what this is all about, these [OpenBSD slides](http://www.openbsd.org/papers/asiabsdcon2015-pie-slides.pdf) give a good overview why (statically linked) PIE is a good idea.

eworm-de:
Yes, looks good! Successfully compiled with `... -fPIE ... -Wl,--no-dynamic-linker -pie ...` and works perfectky fine. Thanks!

Issue #274, opened 2016-04-15
`viB` when bottom of bracket is below screen view, moves view all the way down


erf:
Doing `viB` when bottom of bracket is below screen view, moves the view all the way down so only the bottom of the bracket is showing. It would be better if the bottom of the view is set to the bottom of the bracket, if the bracket is below current screen view.

martanne:
This should be fixed by c8d1e2296688a77d7b0024cc572164dae694cd87.

erf:
Thanks for the fix, but i think it could be improved further by setting the bottom of the view to the bottom of the bracket, so that you can see as much as possible of your selection.

martanne:
Just use `o` to display the other end of the selection.

Issue #164, opened 2016-02-13
View moves down after paste word at end of buffer


erf:
When i paste some text at the last line, the view is moved down, so i only see the last thing i pasted. 
Example:
`ihello<Enter><Esc>kyyppppywjp`

erf:
Thanks for the fix, but this could be improved further by moving the view to the bottom and not stopping at half height ( if you paste the same thing several times, it suddently appear as if its not pasting anymore halfway down ).

martanne:
Won't happen, at least not by myself.

PR #301, opened 2016-04-29
set view to bottom of selection

erf:
This seems to work as desired and disscused in #164 and #274 - could you have a look?

### DIFF
--- a/view.c
+++ b/view.c
@@ -1319,9 +1319,9 @@ void view_cursors_to(Cursor *c, size_t pos) {
 			view_draw(view);
 		}
 
-		if (pos < view->start || pos > view->end) {
+		if (pos > view->end) {
 			view->start = pos;
-			view_viewport_up(view, view->height / 2);
+			view_viewport_up(view, view->height - 1);
 		}
 
 		if (pos <= view->start || pos > view->end) {


martanne:
The idea behind these changes was to show more context *around* the new cursor location. Hence why we attempt to place the cursor in the middle of the screen such that relevant context both before and afterwards is visible.

You will have to make a better case for why you want your proposed behavior. It does not seem to be an improvement to me.

erf:
I believe the behavour i propose is the default in other editors. Also you can see more of the text you are working on (selected).

erf:
Not sure this pull request still work but i found another case which support the proposed behaviour - that is if you have a line that wraps over multiple lines that exceeds the bottom of the screen, if you are at the start of that line and press `j`, it skips the whole line, so you are not able to read it, except if you traverse it by word/char.

apotheon:
What happens in the case that the selected text height is greater than the height of the screen?

erf:
The most important case for me regarding the current view change behaviour, is that when i scroll up or down with `j` and `k`, i loose context of where i am in the document when the view skips up or down half a page - i find it a bit confusing. Maybe the view-change behavour could be configured in the visrc file, with an option to move up/down one line at a time?

erf:
You could also implement vim's `scrolloff` option, which sets "Minimal number of screen lines to keep above and below the cursor.".

http://vimdoc.sourceforge.net/htmldoc/options.html#%27scrolloff%27

PR #306, opened 2016-05-05
Add `vis:map_window()` command

joshaw:
For creating window local mappings to a lua function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -617,6 +617,7 @@ At this time there exists no API stability guarantees.
    - `cursors[1..#cursors]` array giving read access to all cursors
    - `cursor` primary cursor
    - `syntax` lexer name used for syntax highlighting or `nil`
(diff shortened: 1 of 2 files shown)


martanne:
With your proposal `vis:map_window(...)` would always change the mapping of the currently focused window. I'm not sure that this desirable, something like `win:map(...)` might be more appropriate?

Also your patch does not properly [unmap all prefixes](https://github.com/martanne/vis/blob/b04b66e0ab5646342241b2535768337d9d12a93a/vis-lua.c#L414) of the new mapping (see also the `cmd_map` implementation in `vis-cmds.c`, yes there should be an opportunity to share some common code...).

Some slight style issues: the `lua_State` argument is generally passed as the first argument to the functions. Also `map_w` could be renamed to something slightly more descriptive.

PR #307, opened 2016-05-06
some misc changes

eworm-de:


martanne:
Thanks, a couple of comments:

* yes themes in general could need some work from someone with a talent for this kind of stuff ...
* `.cc` also seems more of a C++ than a C file extension
* I left out the patch adding keywords to the C syntax file for now. How did you come up with your set of constants? It seems somewhat arbitrary? In principle I'm not opposed to these kind of changes, but I would like to do them systematically (e.g. referring to the standard etc).

eworm-de:
> * yes themes in general could need some work from someone with a talent for this kind of stuff ...

I am not talented... But I noticed I could hardly read some parts of my files. Thus my changes. ;)
Do we have a sample file that shows all styles?

> * .cc also seems more of a C++ than a C file extension

Never had any files ending with `.cc`... Probably you are right.
Do you want me to make the change?

> * I left out the patch adding keywords to the C syntax file for now. How did you come up with your set of constants? It seems somewhat arbitrary? In principle I'm not opposed to these kind of changes, but I would like to do them systematically (e.g. referring to the standard etc).

The listed constants are what I was missing. For sure the list is far from complete. :-p
A complete list would be awesome... Looks like we have to find good documentation.

Issue #272, opened 2016-04-15
autoindent broken with O


aeosynth:
`O` creates a new line which is not indented; `o` works fine

martanne:
Should be fixed by cd89f1c9898ae02033faab34d72f2e14981088ec.

PR #312, opened 2016-05-08
Fix autoindent when using O

rgburke:
This fixes issue #272.


martanne:
Thanks for looking into this!

PR #313, opened 2016-05-09
style suggestion: black background

erf:


Krak-n:
http://ethanschoonover.com/solarized

erf:
I thought it looked better with black bkg, but i guess we should follow the original. Closing..

PR #317, opened 2016-05-11
build: install support files only when needed

eworm-de:


martanne:
Thanks, I committed it without the need for an additional target and recursive make invocation.

PR #323, opened 2016-05-21
Vis menu fixes

Screwtapello:
A few minor fixes and changes I made while poking around the new `vis-menu` command.

Most of them should be straight-forward, but I'm not 100% sure about the signed/unsigned comparison one; my C is rusty and I basically just beat on things until GCC stopped complaining, so there may well be stupidity in there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ README.html
 *.swp
 *.o
 /vis
+/vis-menu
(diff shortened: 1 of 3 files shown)


Screwtapello:
All the good bits of this PR have been cherry-picked to `master` already, so we don't need this PR.

PR #336, opened 2016-06-14
Update the statusline when starting or ending the recording of a macro

joshaw:


Shugyousha:
Works as advertised.

At the moment vis->event cannot be null so this check can theoretically be removed.

martanne:
Thanks, at some point the drawing related code should be cleaned up further.

I also noticed is that sometimes if you have a modified file opened in more than 1 window and then save the file, only the currently focused window status bar is redrawn. That is the other window still contains the `[+]` marker indicating unsaved changes.

sshbio:
It seems that it only occurs with top-down splits, while saving in the top window.  If the focused window is the bottom, or if this is a left-right split, it does not occur.

And as soon as the focused window changes from top to bottom, the status bar is redrawn correctly.

PR #360, opened 2016-07-05
bash syntax for template files

aeosynth:
`template` files are the Void Linux equivalent of the Arch `PKGBUILD`. example: https://github.com/voidlinux/void-packages/blob/master/srcpkgs/vis/template

possibly a new lexer should be created, since there is a `pkgbuild.lua`. also `template` is a generic name, which could cause unwanted highlighting.


martanne:
I think `template` is a common name which shouldn't be used to indicate a given file type.

And yes, at some point we probably should add some kind of file type detection based on file content.

PR #363, opened 2016-07-14
Allow the :r command to read shell command output

Shugyousha:
This patch allows users to read the output of shell commands into their
files using the :r command. The syntax is the same as the one in Vim
and looks like this.

:r !date

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1117,9 +1117,15 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Curs
 static bool cmd_read(Vis *vis, Win *win, Command *cmd, const char *argv[], Cursor *cur, Filerange *range) {
 	Buffer buf;
 	buffer_init(&buf);
+	char *bang;
 	bool ret = false;
-	if (buffer_put0(&buf, "cat ") && buffer_append0(&buf, argv[1]))
+	if ((bang = strstr(argv[1], "!"))) {
+		while(isspace(*(++bang)))
+			;
+		ret = cmd_pipein(vis, win, cmd, (const char*[]){ argv[0], bang, NULL }, cur, range);
+	} else if (buffer_put0(&buf, "cat ") && buffer_append0(&buf, argv[1]))
 		ret = cmd_pipein(vis, win, cmd, (const char*[]){ argv[0], buf.data, NULL }, cur, range);
+
 	buffer_release(&buf);
 	return ret;
 }


sshbio:
I think @martanne is using sam's syntax for this: `:< command` to read from command.

This works for what I tried, like `:< date`

Shugyousha:
On Thu, Jul 14, 2016 at 01:55:10PM -0700, Josuah Demangeon wrote:
> I think @martanne is using sam's syntax for this: `:< command` to read
> from command.
> 
> This works for what I tried, like `:< date`

I see, I did not know that.

I assume a lot more people are familiar with Vim compared to the sam
language. If @martanne doesn't want the patch (or not in that form), that's
fine too.

sshbio:
I think there were an issue related to '!' being in both vim and sam (hence conflicting).  This [doc from cat-v](http://doc.cat-v.org/bell_labs/sam_lang_tutorial/sam_tut.pdf) (page 10) describes the '!' command in sam.

Thank you for the effort!  This may still be useful as now you got familiar with the source.

It could be great vis to describe on the README (or wiki?) the sam features it uses.  Even if they are nicely presented on the pdf.

Shugyousha:
On Fri, Jul 15, 2016 at 12:17 AM, Josuah Demangeon
> I think there were an issue related to '!' being in both vim and sam (hence
> conflicting). This doc from cat-v (page 10) describes the '!' command in
> sam.

Strictly speaking there is no conflict. :r as a (non-sam) command is
described in the
[README](https://github.com/martanne/vis/blame/master/README.md#L394)
as a command line prompt command.

The question is if you want to add the feature to the (non-sam) :r
command to read external command output using ! or, because the same
can be achieved through the sam command language, you don't want to
have redundant functionality. Having the ":r !"-command would be more
familiar to Vim users but redundant functionality is not desirable.

Maybe it would be best to mention the ":<" sam command in the README
where the :r command is explained so that people coming from Vim could
more easily find the "vis"-way of doing what the :r command does in
Vim?

martanne:
I agree the "documentation" currently sucks. The README is probably not the right place for these things, long term I would like to have a proper manual page describing the structural regexp language and all available commands etc.

For historical reasons the README currently just lists all "vim-like" commands. For the sam related commands we link to the excellent external documentation and only list a few differences in implementation.

As for the problem at hand, `r file` does the same thing in both vim and sam. However as has already been pointed out, sam uses `<`, `>` and `|` to communicate with external commands. In my opinion this is a cleaner approach and the current vis behavior is intentional. For example in vim one might get confused by `w! file` (force write to file) and `w ! command` (write range to stdin of command).

Issue #347, opened 2016-06-29
complete nested filenames


aeosynth:
request: filename completion working in nested directories, not just the current directory

martanne:
Not sure whether this is what you wanted? But now it can at least be tweaked more easily.

The shell script could probably be improved, @Screwtapello, @TieDyedDevil, @rgburke any suggestions?

rgburke:
`vis-complete` looks good. A couple of thoughts come to mind after playing around with it:

- At the moment it's only possible to complete paths below the current working directory. For example, typing `:e /home` and then entering `<C-x><C-f>` is not possible. It would be nice if any path could be completed.
- Performing a full recursive search is expensive when higher up in the directory hierarchy. Opening vis in my home directory I was able to trigger a find command which returned ~ 40,000 results and took several seconds to run. Perhaps one option could be to make `vis-complete` more like a navigtor so that the user can navigate through the filesystem to their file i.e. The user can select a file to use or a directory to traverse into.

PR #385, opened 2016-09-25
Harden vis-complete

rgburke:
Entering a file path into the command prompt containing a forward slash character causes the sed command on line 33 to be invalid. For example, open vis in the root of the repository and on the command prompt enter:
```
:e lexers/
```
Then entering `<C-x><C-f>` causes an error message from sed.

This PR updates the sed command to use the pipe delimiter instead.

### DIFF
--- a/vis-complete
+++ b/vis-complete
@@ -3,6 +3,7 @@ set -e
 
 PATTERN=""
 COMPLETE_WORD=0
+FIND_FILE_LIMIT=1000
 
 while [ $# -gt 0 ]; do
 	case "$1" in
@@ -24,12 +25,16 @@ while [ $# -gt 0 ]; do
 	esac
 done
 
+PATTERN="$(echo "$PATTERN" | sed "s/'/'\\\\''/g")"
+
 if [ $COMPLETE_WORD = 1 ]; then
 	CMD=$(printf "tr -cs '[:alnum:]_' '\n' | grep '^%s.' | sort -u" "$PATTERN")
 else
-	CMD=$(printf "find . ! -path '*/\.*' -a -path './%s*' | cut -b 3- | sort" "$PATTERN")
+	CMD=$(printf "find . ! -path '*/\.*' -a -path './%s*' 2>/dev/null | head -n $FIND_FILE_LIMIT | cut -b 3- | sort" "$PATTERN")
 fi
 
-CMD=$(printf "$CMD | vis-menu -b | sed 's|^%s||' | tr -d '\n'" "$PATTERN")
+PATTERN="$(echo "$PATTERN" | sed 's:/:\\/:g')"
+
+CMD=$(printf "$CMD | vis-menu -b | sed 's/^%s//' | tr -d '\n'" "$PATTERN")
 
 exec /bin/sh -c "$CMD"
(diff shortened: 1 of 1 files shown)


bb010g:
Wouldn't this still break on filenames with pipes in them?

rgburke:
Yes, it will. Ultimately the solution to this is to escape whichever delimiter character is chosen in the sed command. I submitted this as a quick fix that will work in most cases. I don't have time to work on this at the moment and considering there are a couple of other issues with this script:

- If `PATTERN` contains a single quote character it will break.
- `find` gives out "Permission denied" errors for directories it cannot access which are displayed in the command prompt. This could be solved by using the `-perm` option (or simply redirecting stderr to `/dev/null`).
- Running a recursive directory search could be slow especially if there is a deep directory hierarchy or vis is opened under a mounted SD card or perhaps an nfs share for example. Traversing an entire tree when a user is only interested in a single path is quite inefficient.

I thought it would be sufficient for the moment. If anyone else wants to look at this then please do and I will close this PR.

martanne:
I agree with your comments. The naive nested completion doesn't scale. I just added it because it was requested in #347. We already have a rudimentary interactive directory browser in `vis-open`. In fact for your example you can simply type `:e lexers<Enter>` which should list the directory content which you can then further navigate. This should also work with reasonable performance for `:e /home<Enter>` etc.

The `<C-x><C-f>` functionality was intended for use in insert mode e.g. to complete project local C `#include` directives. The assumption here is that the project directory contains only a limited number of files. In this case the recursive directory listing allows to skip intermediate directories.

But as your example shows, it is probably not a good idea to have multiple scripts which do roughly the same. Maybe `<C-x><C-f>` should just use `vis-open`?

rgburke:
It's an interesting question. On the one hand you have the normal usages of vis-complete that you have outlined such as completing header files or files from the project directory. In these circumstances vis-complete is very effective and useful. Even for larger directory hierarchies, those with an SSD may not mind a full recursive search being performed. At the same time there will be users (like me 😄) who will end up using it on large hierarchies causing it to freeze for several seconds.

After reading your response about the fact that `vis-open` is already available as a directory browser that can be used as an alternative if necessary, leads me to thinking that `vis-complete` should continue to exist as a separate tool. Although it somewhat duplicates `vis-open`, it is nonetheless distinct in it's purpose of being a fuzzy finder like tool.

I'm wondering if we could get around the potential cost of the find command by simply limiting the number of files it searches for, to 1000 for example. Piping `find` into `head -n 1000` would cause `find` to receive a `SIGPIPE` and exit when it attempted to write the 1001st result. Hopefully no one would need to scroll through more than 1000 files in normal usage. What do you think?

rgburke:
Made some minor updates to `vis-complete` allowing it to deal with the scenarios mentioned earlier.

josuah:
Why hacking so hard to build a sed command ?  What about using awk's feature to catch variables with the -v flag?

`awk -v awk_variable="$shell_variable" '{ <substitution how you like }' current_file_name.ext`

martanne:
@rgburke thanks, that is certainly an improvement!

Absolute paths should probably be handled better, something like:
```
/home/marc<c-x><c-f>
```
should list the home directory contents instead of the current working directory? Thoughts?

However, at the moment only `marc` is passed as a completion prefix to the script.

PR #388, opened 2016-09-27
Better file type detection: full filename patterns, shebang, custom detector functions

lxyd:
I just thought filetype detection could be better than simply matching filename extension :)

What was done:
• ignoring `~`, `.bak`, `.orig` and other extensions that are not bound with any file format 
• filename patterns (full filename match against lua pattern);
• shebang detection (application name match);
• custom detection functions.

Benefits:
• scripts with no file extension will be correctly syntax-highlighted;
• xml-based files like `.iml` or others (millions of them) will be syntax-highlighted even if particular extension is not listed;
• it is possible to add syntax highlight for mercurial commit message editor in future (its filenames are like `hg-editor-XYZXYZ.txt` and they couldn't be matched without pattern-based detection).

### DIFF
--- a/vis.lua
+++ b/vis.lua
@@ -193,7 +193,7 @@ vis.ftdetect.filetypes = {
 -- array of filetype detecting functions (win, filename, shebang, app) -> string
 vis.ftdetect.customdetectors = {
 	function(win)
-		if win.file.lines[1] ~= nil and win.file.lines[1]:sub(1, 5) == '<?xml' then
+		if win.file:content(0, 5) == '<?xml' then
 			return 'xml'
 		end
 	end,
@@ -221,8 +221,8 @@ vis.filetype_detect = function(win)
 	-- find out via shebang which application would run our file
 	local shebang
 	local app
-	if win.file.lines[1] ~= nil and win.file.lines[1]:sub(1, 2) == '#!' then
-		shebang = win.file.lines[1]:gsub('^#!%s*', '')
+	if win.file:content(0, 2) == '#!' then
+		shebang = win.file:content(0, 256):gsub('^#!%s*', ''):gsub('\n.*$', '')
 		app = shebang:gsub('^/usr/bin/env%s*', ''):gsub('%s.*$', ''):gsub('^.*/', '')
 		if #app == 0 then
 			app = nil
--- a/vis.lua
+++ b/vis.lua
@@ -69,135 +69,212 @@ vis:textobject_new("ii", function(win, pos)
 	return pos, pos
 end)
 
-vis.filetypes =	{
-	[".1|.2|.3|.4|.5|.6|.7|.8|.9|.1x|.2x|.3x|.4x|.5x|.6x|.7x|.8x|.9x"] = "man",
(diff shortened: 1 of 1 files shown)


martanne:
Thanks! I haven't yet had time to properly review or test it. Thus only a few preliminary comments:

1. We want to be fast even when the file only contains 1 long line. Therefore instead of `win.file.lines[1]` you should use `win.file:content(0, 256)` or similar which doesn't require to read the whole line.

2. I would like this to be as "data driven" as possible. Intuitively I would expect each file type to define:
 * a filename pattern
 * a file content pattern for the first few bytes of a file (e.g. for scripting languages like python you would check that it starts with `#!` and ends with `python` i.e. there shouldn't be a need to deal with `/usr/bin/env` in your code)
 * for certain formats a function taking the first few bytes of the file and returning whether it is of the expected type (similar to what you have done with `customdetectors`, but per file type)

Now the question is how to best implement this in Lua. The built in Lua patterns are quite limited, for example they do not support alternation (hence the current hack of splitting the file extension by `|`). We already depend on LPeg for syntax highlighting, but maybe this would be overkill here? Also such patterns to identify file types surely already exist in other editors. However they are likely in regex format which would require another dependency that I would like to avoid if possible.

Suggestions (especially from people with Lua experience and good taste ;) welcome ...

lxyd:
Wow, I thought lines are already split and read in memory by the time detection goes. Ok, will try to handle this somehow (in a few days).

> there shouldn't be a need to deal with /usr/bin/env in your code

And here I need to describe my motivation for some decisions (after that you decide, whether to agree with them or not :) Here they are:

Sometimes (rare, but still…) there are additional arguments for the interpreter mentioned in the shebang, for example `#! /usr/bin/env bash -x` or `#!/usr/bin/perl -T`, and sometimes you can see `python2.7`, `python3.4` and other python shebangs not ending with `python`. So, I tried to find the most simple yet close to universal way to get meaningful part of the shebang and compare against it. Custom detectors might still use whole shebang.

Also, repetitive matching of the `#!` for each scriptable filetype seems to be irritating and error-prone when patterns are written by different people.

Same with filename patterns: `.ext|.anotherext` is much cleaner than `.*%.ext|.*%.anotherext`, so I didn't merge `ext` and `name` patterns which was almost obvious step.

And `|`-separation is looking simpler than separate patterns `'.ext', '.anotherext'` (also, there clearly no file extension or interpreter name exist with `|` in it, so no conflict is possible).

By the way, implicit `^` and `$` are added for the sake of the pattern tidiness too.

martanne:
I'm not sure whether I agree that the separation of `ext` and `name` is simpler. But details like these can be changed later on. Thanks!

martanne:
In 52cff9b8a81f641403df55cf5e081f6f1212d2fd I changed the file type detection to use `file(1)` having these file type detection heuristics in the editor seems wrong. They are more generally useful (for example in plumbing scripts) and should thus be maintained outside the editor.

We will have to add more mime types to our file type mapping table and possibly improve the magic database used by `file(1)` where necessary.

PR #389, opened 2016-09-28
Add `"` as a clipboard register (synonym to `+` and `*`)

lxyd:
Almost no hope this will be approved, but still… `""y`, `""p` would rock, imho.

Adding `"` register as yet another synonym for `clipboard` would save some hand moves when interacting with the outer world (now you have to move your index-finger to press`*y` in a row or move your little finger for `"+`, which is slow). The `"` register will save one slow movement for the user. If user frequently copy/pastes text from/to another application this might be handy.


martanne:
While convenient in some cases this is probably not such a good idea.

PR #394, opened 2016-10-03
Man page from the read me

josuah:
Here we go.  A much more complete `mdoc` formatted man page.

I used the [`mdoc` man page](http://mdocml.bsd.lv/man/mdoc.7.html#Bd) for documentation, and the [tmux man page](https://raw.githubusercontent.com/tmux/tmux/master/tmux.1) as a model.

Does it look ok to you?  I tried to follow the standards and keep semantics, as I never wrote `mdoc` before.

I did not remove the content from the README yet.

### DIFF
--- a/vis.1
+++ b/vis.1
@@ -2,23 +2,31 @@
 .Dt VIS 1
 .Os
 .Sh NAME
+.
 .Nm vis
 .Nd a vim like text editor
+.
 .Sh SYNOPSIS
+.
 .Nm vis
 .Op Ic +command
 .Op Ar files Fl
+.
 .Sh DESCRIPTION
+.
 .Nm vis
 is a highly efficient vim like text editor.
 .Bl -tag
 .It Fl v
 Print version information and exit.
+.
 .It Fl -
 Denotes  the  end  of the options. Arguments after this will be handled as a
 file name.
 .El
+.
 .Ss Modes
+.
 Vis implements more or less functional normal, operator-pending, insert,
 replace and visual (in both line and character wise variants) modes.
 .Pp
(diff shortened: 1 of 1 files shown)


Screwtapello:
It would be nice to fill out the "SEE ALSO" section with cross-references to `vis-open`, `vis-menu`, `vis-complete`, `vis-clipboard` and any others I've forgotten about. For that matter, it'd be worth making any other mentions of those tools in other parts of the document into cross-references, too.

I'd also remove the paragraph beginning "Another idea..."; man pages should describe what the program does, not what it might do in the future.

martanne:
Thanks! I hope to have some time to revisit the man page towards the end of the month.

josuah:
I am getting some trouble to get started with LuaDoc, I may not know the source well enough to spot where should the lua bindings be documented.  I will try again soon.

On the other hand, if you need anything more about any man page, let me know!

martanne:
Did you ever make progress with the Lua documentation?

It seems like [LDoc](https://github.com/stevedonovan/LDoc) is the tool to use. For the Lua source files this should be no problem. The part of the Lua API which is implemented in C will likely be more problematic. Documenting [methods implemented in C](http://stevedonovan.github.io/ldoc/examples/mylib.c.html) should work, but I haven't yet found an example how to document field properties (i.e. the [`index/newindex` implementations](https://github.com/martanne/vis/blob/master/vis-lua.c#L1014-L1017)).

Could you look into it?

martanne:
@josuah Ok I spend some time and [added a few luadoc comments](https://github.com/martanne/vis/commit/960cd1c7c84c4573f99d4e389548183ebbdc4f35) to test things out. It seems to mostly work, but the output is not yet ideal. For example the links (e.g. in parameter lists) to our custom types seem to be broken.

Instead of these inline comments we could add a separate Lua file with the same interface as implemented by the C code and document it there. For now I decided against that, with the hope that the documentation is more likely to be  updated when it is located closer to the source.

josuah:
@martanne Sorry, no I did not spend more time for getting it to work.  You did it, though.  I am having a look at the makefile and ldoc doc to know how.

I do not see which special type is broken: http://martanne.github.io/vis/doc/index.html#Class_Range works for me as an example.

Writing doc where things are implemented looks good to me too.

PR #399, opened 2016-10-08
vis-complete improvements

Screwtapello:
This PR improves vis-complete in the following ways:

- no more repeated shell-evaluation, so shell-sensitive characters in the pattern can't cause problems.
- the pattern is properly quoted before being handed to `grep`, so regex-sensitive characters can't cause problems
- the pattern is properly quoted before being handed to `find`, so glob-sensitive characters can't cause problems
- just as before, giving a relative path will complete relative to the current directory—but now, giving an absolute path works too

This code uses arithmetic expansion (`$(( ))`) but that's [definitely POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_04).

### DIFF
--- a/vis-complete
+++ b/vis-complete
@@ -1,6 +1,9 @@
 #!/bin/sh
 set -e
 
+basic_regex_quote() { printf "%s" "$1" | sed 's|[\\.*^$[]|\\&|g'; }
+glob_quote () { printf "%s" "$1" | sed 's|[\\?*[]]|\\&|g'; }
+
 PATTERN=""
 COMPLETE_WORD=0
 FIND_FILE_LIMIT=1000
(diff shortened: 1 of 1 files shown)


rgburke:
Nice work, it's pleasing to see `vis-complete` refined. After playing around with it for a while I did notice one issue with relative paths. Running `richard@dev:~/code/tig$ vis-complete ../vis/ui-` didn't produce any suggestions, whereas I would have expected `ui-curses.c` and `ui-curses.h` to be listed. This is due to the find command:

```find /home/richard/code/tig/../vis '!' -path '*/\.*' -a -path '/home/richard/code/tig/../vis/ui-*'```

which excludes any path containing a forward slash followed by a dot and in this case matches the parent directory sequence. An initial thought was to use readlink to normalise the paths we provide to the find command, however it seems the behaviour of readlink is different on OSX. Any thoughts?

Screwtapello:
I pushed an new commit that updated the `-path` filter that vis-complete used to avoid hidden directories, but then I realised that approach had too many corner-cases (what about paths like `..veryhidden` or if you were trying to complete a path inside `~/.local/` or some other legitimately hidden directory).

So now I've changed the find invocation so it uses `-name` to filter out hidden files and directories. This means that these filters can never affect the completion prefix. Also, finding should be faster—instead of finding files in hidden directories then filtering them out, we actively avoid traversing hidden directories with `-prune`.

eworm-de:
`vis-complete` does return success (return code `0`) even when user cancels completion. I do not think this behaviour is expected, `vis-open` returns `1` to indicate error.

PR #402, opened 2016-10-11
File completion updates

rgburke:
- Update `vis` to provide full file path to `vis-complete`.
 - Update `vis-complete` to expand tilde to `$HOME` when it's the first character in a file path.
 - Fix issue with `vis-complete` producing incorrect output for relative file paths. The `PATTERN` variable cannot be modified in a pipeline command as each command in a pipeline is run in its own subshell. This caused the value of `${#PATTERN}` to be incorrect for relative paths. This is resolved by modifying `PATTERN` before the pipeline is run.

### DIFF
--- a/main.c
+++ b/main.c
@@ -2103,11 +2103,11 @@ static const char *open_file_under_cursor(Vis *vis, const char *keys, const Arg 
 	return keys;
 }
 
-static char *get_completion_prefix(Vis *vis) {
+static char *get_completion_prefix(Vis *vis, Filerange (*text_object)(Text *, size_t)) {
 	View *view = vis_view(vis);
 	Text *txt = vis_text(vis);
 
-	Filerange r = text_object_word(txt, view_cursor_get(view)-1);
+	Filerange r = text_object(txt, view_cursor_get(view)-1);
 	r = text_range_inner(txt, &r);
 	size_t size = text_range_size(&r);
 	if (size == 0) {
@@ -2140,7 +2140,7 @@ static const char *complete_word(Vis *vis, const char *keys, const Arg *arg) {
 	Text *txt = vis_text(vis);
 	Buffer cmd;
 	buffer_init(&cmd);
-	char *prefix = get_completion_prefix(vis);
+	char *prefix = get_completion_prefix(vis, text_object_word);
 	if (prefix && buffer_printf(&cmd, VIS_COMPLETE " --word '%s'", prefix)) {
 		Filerange all = text_range_new(0, text_size(txt));
 		insert_dialog_selection(vis, &all, (const char*[]){ buffer_content0(&cmd), NULL });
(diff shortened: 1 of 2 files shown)


Screwtapello:
Thanks for picking up that variable-modification-in-a-pipeline bug, this is a good way to fix it.

This PR doesn't support the `~username` syntax for the home directory of another user, but that's probably not very useful in this modern age when everybody's using their own private laptop instead of a multi-user minicomputer, and it would be a pain in the neck to re-implement in shell, so I don't blame you for skipping it.

Looks good to me!

martanne:
Thanks!

As discussed on IRC 55666d0f67bb94d5654a9389031fa9269cf512a5 changed how the prefix is removed at the end of the `vis-complete` script: instead of something along the lines of `sed s/^$PATTERN/ //` we now use `cut -b ...`. This makes it impossible to manually enter a suffix which is not proposed as a completion candidate.

I personally liked the old behavior better. It was also proposed that vis should completely replace the existing prefix with whatever vis-complete outputs. Hence vis-complete would act as a filter and the prefix stripping code could be removed. However I'm skeptical whether this would be an improvement, comments?

@Screwtapello also remarked that the text objects should be right bounded by the current cursor position. That is if we have `foo#ar` (where `#` denotes the cursor) we should complete only for `foo` not `foobar`. The following patch implements that behavior:
```diff
diff --git a/main.c b/main.c
index b2d364a..aecbff9 100644
--- a/main.c
+++ b/main.c
@@ -2106,9 +2106,11 @@ static const char *open_file_under_cursor(Vis *vis, const char *keys, const Arg
 static char *get_completion_prefix(Vis *vis, Filerange (*text_object)(Text *, size_t)) {
        View *view = vis_view(vis);
        Text *txt = vis_text(vis);
-
-       Filerange r = text_object(txt, view_cursor_get(view)-1);
+       size_t pos = view_cursor_get(view);
+       Filerange r = text_object(txt, pos-1);
        r = text_range_inner(txt, &r);
+       if (r.end > pos)
+               r.end = pos;
        size_t size = text_range_size(&r);
        if (size == 0) {
                vis_info_show(vis, "No valid prefix found for completion");
```

Unless someone objects, I plan to apply it.

rgburke:
Thanks for the feedback guys. I agree it's better if text objects are right bounded by the cursor position and I'm happy for the patch to be applied.

My preference would be for `vis-complete` not to act as a filter. It seems safer not to modify user input. For example, in its current form `vis-complete` would convert relative paths to non-canonical absolute paths when acting as a filter. Some users may prefer this behaviour but to others it could be unwanted. It's worth noting that `vim` doesn't appear to modify file paths when they are completed, it leaves them in a non-absolute form.

martanne:
I agree with @rgburke the non-filter behavior seems preferable.

> As discussed on IRC 55666d0 changed how the prefix is removed at the end of the vis-complete script: instead of something along the lines of sed s/^$PATTERN/ // we now use cut -b .... This makes it impossible to manually enter a suffix which is not proposed as a completion candidate.

This is still a problem. We should remove the initial completion prefix iff it occurs at the start of the string. Do you have a better suggestion than reverting to the older mechanism which was:

```diff
diff --git a/vis-complete b/vis-complete
@@ -64,5 +64,5 @@ else
                sort
 fi |
        vis-menu -b |
-       cut -b $(( ${#PATTERN} + 1 ))- |
+       sed "s/^$(printf "%s" "$PATTERN" | sed 's:/:\\/:g' )//" |
        tr -d '\n'
```

PR #414, opened 2016-11-19
Fix date in man pages

michaelforney:
Alternatively, these dates should probably be added the man pages themselves so they reflect the date the document was written, not when it was installed.


martanne:
Thanks! I agree using the install date might be slightly misleading. The "problem" with static dates is of course that they won't get updated when the content changes. Maybe it would be best to remove them completely? This would probably also help with reproducible builds. The version information is already in the manual page which should be more meaningful anyway.

PR #418, opened 2016-11-29
vis-cmds: Check for overflow

Shugyousha:


martanne:
Thanks, however I think your error checking isn't quite right. The [OpenBSD man page](http://man.openbsd.org/OpenBSD-current/man3/strtoul.3) has a nice example.

Also the existing code combines both the signed (`OPTION_TYPE_NUMBER`) and unsigned (`OPTION_TYPE_UNSIGNED`) cases in a single code path which probably is not entirerly correct because of the resulting union based type punning.

Shugyousha:
Thanks for having a look.

On Wed, Nov 30, 2016 at 12:15:39PM -0800, Marc André Tanner wrote:
> Thanks, however I think your error checking
> isn't quite right. The [OpenBSD man
> page](http://man.openbsd.org/OpenBSD-current/man3/strtoul.3) has a
> nice example.

My patch only covers the overflow case (without resetting errno beforehand
though). I will change the patch according to the man page so that it
covers the "trailing characters" case as well.


> Also the existing code combines both the signed (`OPTION_TYPE_NUMBER`)
> and unsigned (`OPTION_TYPE_UNSIGNED`) cases in a single code path
> which probably is not entirerly correct because of the resulting union
> based type punning.

If you prefer I can split these two cases and use strtol for the
OPTION_TYPE_NUMBER case?


Cheers,

Silvan

Shugyousha:
I added the check for unexpected characters. Let me know if you want me to deal with the signed number option case as well.

martanne:
Thanks, in the end I decided to fix it slightly differently. Hope it works as expected.

PR #423, opened 2016-12-05
do not call file but use libmagic binding for lua

eworm-de:
This requires lua-magic [0] to be installed.

[0] https://github.com/mah0x211/lua-magic

### DIFF
--- a/vis.lua
+++ b/vis.lua
@@ -464,19 +464,19 @@ vis.filetype_detect = function(win)
 
 	-- run file(1) to determine mime type
 	if name ~= nil then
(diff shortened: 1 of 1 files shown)


eworm-de:
This has some rough edges, but I think it works. Do we want something like this?

martanne:
> Do we want something like this?

Probably not. What advantages would it have? Run time should not be an issue (yes forking is slower than a library call). Using a sub process is more robust. If the file detection crashes for some reason, we will keep running. Also we can assume `file(1)` is generally available while depending on lua-magic complicates things.

eworm-de:
I feel the same. :-p But it was fun to try. :D
I will keep the branch around in case anybody is interested in this...

PR #426, opened 2016-12-08
link piece table info to external file

erf:


aeosynth:
bikeshedding: instead of its own header, you could place the link inside the 'developer overview' section

erf:
Thanks for you advice @aeosynth - i've now put the link under `developer overview`. 

I was thinking maybe the whole `developer-overview` section could be a separate file..

PR #427, opened 2016-12-13
Allow building lpeg into vis

michaelforney:
If lpeg is built statically, this allows for a completely static vis
binary that still supports syntax highlighting.

Ideally, this would be paired with corresponding configure support. However, even without that I think this change is still useful by itself.


martanne:
Thanks! That is certainly a better approach than what I [hacked together](http://www.brain-dump.org/projects/vis/lua-5.2.4-lpeg.patch1) to achieve the same goal.

> Ideally, this would be paired with corresponding configure support.

I agree, my follow up patches try to do that. Although the whole build system could probably be cleaned up a bit, it seems to work here. Hope it didn't break for you?

PR #444, opened 2016-12-22
fix compiler warning

eworm-de:
Compiler warns:

warning: initialization makes integer from pointer without a cast

Initialize with 0, not NULL.


eworm-de:
Thanks!

PR #449, opened 2016-12-30
Add rc lexer

michaelforney:


martanne:
Do you have a corresponding change for the automatic [file type detection](https://github.com/martanne/vis/blob/master/lua/plugins/filetype.lua)?

Also if you have ideas how to improve/simplify the `file(1)` based mime type detection I would like to hear them.

Thanks

michaelforney:
I do have a patch that I'm applying locally, but it seemed like the `.rc` extension wasn't exclusive to rc shell scripts, and is sometimes used for configuration files, so I wasn't sure if it was suitable for applying upstream.

`file(1)` *does* seem to detect `Plan 9 rc shell script, ASCII text executable`, but unfortunately this additional info is not present in the mime-type, which is just `setup.rc: text/plain; charset=us-ascii`.

I don't really have any suggestions about how to improve the mime type detection, unfortunately. The only thing I could think of is to check for a shebang line interpreter against a regular expression on files that `file(1)` reports are `text/plain` for further classification.

PR #452, opened 2017-01-02
lua: adjust C lexer to recognize labels and POSIX types

s-gilles:


s-gilles:
I find that having `size_t` colored differently from `int` a little jarring, and it's nice to be able to find `err_overflow:` at a glance.

Possible objections to this change might be that this should go to Scintilla instead, or that the POSIX types are stepping a bit beyond ANSI.  That would be fine, I've just been keeping this in /etc/portage/patches/ for a while and thought I might as well try to send it up.

martanne:
Thanks, contributing lexer changes directly here is fine. I will [periodically](https://foicica.com/lists/code/201601/2859.html) send them [upstream](https://foicica.com/lists/code/201610/3373.html) when re-syncing with scintillua.

I agree that the we should be more consistent with regard to the data types which we highlight. Although POSIX reserves the `_t` suffix, a lot of code ignores it and defines other custom types in the same namespace. Therefore it is probably not a good idea to highlight everything ending in `_t`. Instead we should add a fixed set of known types. I suggest to start with `stddef.h`, `stdbool.h` and `stdint.h`. @eworm-de also proposed to add [error codes](https://github.com/martanne/vis/pull/307/commits/32a784880f10ce278b496d97960e38988a62a470). My only objection is to do it somewhat systematically. For controversial stuff we could add it to the lexer file but not actually reference it by default to make customization more convenient.

As for highlighting labels, your current pattern also matches case labels which is probably not a good idea.

Regarding the actual code changes, the rule order as described in the "Rules" section of the [lexer module documentation](https://foicica.com/scintillua/api.html#lexer) is important. Your new rule should come after the existing whitespace rule.

s-gilles:
My apologies for the long delay - my email notifications didn't work as expected and I didn't realize you'd given feedback yet.

Anyway, I've addressed your comments (I hadn't even noticed the matching in switch statements).  If you're really uncomfortable with `uint_24_t` being matched, I can replace the integer-type expression, but the way the standard is worded I think I'd have to leave out some pretty common types then, like `uint32_t`.  The three files you mentioned cover all of my concerns, so I don't have a huge desire to go beyond them.

martanne:
Thanks! I merged the POSIX types part of the patch. Not yet sure whether I like the labels part.

Other things which could probably be improved in the C lexer:
* highlight filenames following an `#include{,_next}` directive
* treat `#if 0` ... `#endif` as comment (might not be worth the trouble though?)

eworm-de:
What I really miss from the C lexer is highlighting `NULL`.

martanne:
I think I will leave out the labels for now. They tend to be aligned on the first column and thus already stand out.

I'm closing this, my "wishlist" improvements are unrelated. The other changes of this pull request have been incorporated and [upstream was notified](https://foicica.com/lists/code/201702/3544.html).

Thanks!

s-gilles:
Sounds good to me.  I've been keeping an eye out for the wishlist items, and if I ever get bit by them I'll open new PRs.

Issue #356, opened 2016-07-03
support 24bit true color


aeosynth:
vis currently only supports 256 colors

pickfire:
Would be nice to bump @XVilka https://gist.github.com/XVilka/8346728

martanne:
As you noted in another comment this is a restriction of curses because the terminfo database doesn't currently handle 24bit color support.

For our use case we don't actually need a lot of different colors. We just want to specify them via RGB values. Interestingly curses provides the [`init_color(3)`](http://pubs.opengroup.org/onlinepubs/7908799/xcurses/init_color.html) function to redefine an existing color to a RGB value. I don't know whether popular terminals support these run-time changes to their color palette.

If you are interested in this feature I suggest to:

1. Implement a minimal curses application which tries to modify one of the standard colors
2. Test it in various terminals to see whether it works
3. If it works, hook it up to the color handling code in [ui-curses.c](https://github.com/martanne/vis/blob/5c2cee9461ef1199f2e80ddcda699595b11fdf08/ui-curses.c#L441)

As for completely ditching curses, a minimal vt100 compatible ui backend might be of interest for some usage scenarios. The re-drawing code of vis is currently intentionally very primitive. The window content is often redrawn from scratch and it is assumed that it can efficiently be "blitted" to the display. For curses this works because it uses double buffering to minimize actual terminal output. I don't know how bad the effects would be for a raw vt100 backend.

Screwtapello:
I believe palette setting is fairly-widely supported, but frowned on because there's no good way to undo the changes after they're made, so you leave a mess for any later tools that run in the same terminal.

There's a lot of useful functionality that is widely implemented but (for whatever reason) is not advertised in the `xterm-256color` terminfo entry, like mouse support or setting the title-bar text. At some point Vis might have to do what everybody else seems to do, which is obey terminfo + use some hard-coded terminal sequences if `$TERM` starts with `xterm`.

mawww:
Stumbled on this issue, I figured I could add some info I learned:

* Most terminal emulators do support changing palette, at least vte, rxvt-unicode, and xterm do. Tmux does not yet.
* Some support OSC 104, which resets a color/the whole palette. notably vte and xterm.

[...]

Issue #250, opened 2016-04-11
Migrate to libtickit (libtermkey deprecated)


DeltaHeavy:
I'm not sure if this will "just work" once libtermkey is cleaned out as a dependency, or if it'll require some deeper updating, but libtermkey is in fact deprecated and so making this project viable into the future will likely require transition.

It'd be best to do this before any extra tech debt is accrued by putting it off.

Thanks! I'm really interested to see how this project grows.

aeosynth:
dupe of #79

DeltaHeavy:
Duplicate

martanne:
Some remarks since this has come up a couple of times:

* libtermkey is only used for terminal input handling
* terminal output / drawing the user interface is currently taken care of by curses
* libtickit merges the two i.e. it would be a replacement for both libtermkey and curses
* libtickit has not yet seen an official initial release? I'm not sure what its development state is? This would make packaging vis even more cumbersome
* I haven't yet looked at what libtickit actually does under the hood. vis currently assumes that the underlying curses library performs double buffering. That is vis always completely redraws its interface and relies on the curses library to figure out the optimal set of escape sequences to update the terminal state.
* I'm in principle not opposed to dropping ncurses. In fact its code is hideous. I would eventually like to experiment building against [netbsd-curses](https://github.com/sabotage-linux/netbsd-curses) (especially for `make standalone`).

erf:
did you look at [termbox](https://github.com/nsf/termbox/) ?

josuah:
I like how termbox strives to be simple, but it requires python to build, as it uses waf, which is a python script.  On the other hand, the library only takes ~1000 lines.

I do not see why it is so hard to draw on a terminal:  We need to know how much width a char takes on the screen that we can do with UTF-8 library, then just have a function that draw a single line, used to redraw the whole buffer, to scroll one line up / down...

This is an [attempt of implementation](https://github.com/josuah/iode/blob/b73825edb93807394cb8b7b56883b6c3f4163029/draw.c), if ever interested.  Still has some issues at this point.

martanne:
[...]

PR #455, opened 2017-01-05
24bit via init color

s-gilles:
This addresses #356.  The first three commits deal with allowing 24bit color on terminals where ncurses allows it, the last two commits implement ideas which were mentioned in #356's discussion.


martanne:
After using this for a day I realized some problems:

* Restoring the color palette for `:!` commands interferes with `vis-menu(1)` which for example is invoked by `:open .` or for word completion.
* Restoring the color palette should probably respect the users default terminal color scheme. Thus I switched it to use the OSC 104 terminal escape sequence.

Let me know whether the latest changes still work for you?

s-gilles:
Everything still works just fine for me.  I'm a little wary of using OSC 104 as the reset (if hard-coding sequences is the way to go, why not just hard-code the 24bit sequence and skip `init_color` altogether?  It would probably be a lot simpler), but since nothing I use relies on the 256color palette being in a particular initial state, I don't think I'll mind if I happen to use a terminal that doesn't understand OSC 104.

I really wish ncurses had an `uninit_color()` though.  Perhaps if it did, we wouldn't have so many conflicting ways of specifying RGB.

martanne:
It is not really "fine" to use hard-coded escape sequences. However, in this case it seems fairly non-invasive and the only solution which at least somewhat works.

Also while I agree that the curses API is suboptimal I would still like to use it for the things it provides (e.g. color setting). In the long term we could consider using better/more lightweight terminal handling libraries. In the past libtickit (#250) was suggested.

s-gilles:
Fair enough.  I'm not about to charge into implementing that however - everything is currently working just fine for me.  Thanks for the merge!

erf:
@martanne i've been playing around with some escape codes at https://github.com/erf/sta, creating a minimal abstraction for VT 100 and implementing Game of Life. Probably not usable for much more right now except some fun. It does seem to performant enough though. Maybe i'll look into supporting more terminals and colors later.

martanne:
@s-gilles it would be nice if you could check whether the recent display code refactoring does anything stupid (e.g. with regard to color management etc.)?

s-gilles:
On 2017-03-14T13:28:05-0700, Marc André Tanner wrote:
> @s-gilles it would be nice if you could check whether the recent display code refactoring does anything stupid (e.g. with regard to color management etc.)?

I don't see anything that jumps out at me as bad.  It seems to work in
all the environments I care about (including 16- and 256-color terminals
and tmux).

The only thing that concerns me is how resuming works. On suspending it
sends OSC 104, but then `fg` brings back the colors just fine without
use of any sort of “I was suspended” flag.  Since none of the
caching machinery (`clobbering_colors`, etc.) gets reset following a
suspend/resume, it's a bit confusing how it comes back.  I remember
wondering about this a while back.  A few brief tests indicate that
ncurses is doing something clever with the palette in `SIGCONT`(?),
but I'm finding it rather difficult to find any kind of documentation
on ncurses' signal handlers to confirm.

That's not really related to the refactor, though.

Issue #430, opened 2016-12-14
<Enter> does not emit input event


aeosynth:
typing `<Enter>` does not emit an input event, in input or replace mode

martanne:
The problem is that `<Enter>` is mapped to a key binding which consumes the event. After:
```
:unmap insert <Enter> 
```
you should get the events (but you will have to handle insertion of newlines etc. yourself).
And yes, the interaction between the editor core and Lua scripts could probably be improved.

Issue #383, opened 2016-09-24
autoindent leaves indent characters on blank lines


Screwtapello:
## Steps to reproduce:
1. Run vis on a new file
2. `:set autoindent on`
3. Type "ia", \<Enter\>, \<space\>, \<space\>, "b", \<Enter\>, "c", \<Enter\>, \<Esc\>
## Expected result:

In these "screenshots", spaces will be replaced by `.` for visibility, and the cursor is represented by `_`

```
a
..b
..c
_
```

After hitting "c", \<Enter\>, there was initially an indent before the cursor, but I expected it to be removed when I hit \<Esc\>.
## Actual result:

```
a
..b
..c
.._
```

After hitting \<Esc\>, the automatically-inserted indent just sticks around.
## Notes:

The Vim docs for `'autoindent'` say:

> If you do not type anything on the new line except \<BS\> or CTRL-D and then type \<Esc\>, CTRL-O or \<CR\>, the indent is deleted again.  Moving the cursor to another line has the same effect, unless the 'I' flag is included in 'cpoptions'.

This is annoying because I'm used to moving between blocks of code with `{` and `}`, and they require completely empty lines, while this autoindent feature leaves non-empty whitespace-only lines all over the place.

PR #472, opened 2017-01-22
[wip] smart auto indent

aeosynth:
i have the feeling that `vis_insert_nl` + `copy_indent_from_previous_line` should be refactored; a newline can be inserted at the start of the line, instead of copying and deleting whitespace. also, dedent on `<Escape>` is  not handled

i'm not experienced with c, looking for feedback

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -1410,12 +1410,17 @@ static void copy_indent_from_previous_line(Win *win, Cursor *cur) {
 		return;
 	size_t begin = text_line_begin(text, prev_line);
 	size_t start = text_line_start(text, begin);
+	size_t end = text_line_end(text, begin);
 	size_t len = start-begin;
 	char *buf = malloc(len);
 	if (!buf)
 		return;
 	len = text_bytes_get(text, begin, len, buf);
 	text_insert(text, pos, buf, len);
+	if (start == end) {
+		text_delete(text, begin, len);
+		pos -= len;
+	}
 	view_cursors_to(cur, pos + len);
 	free(buf);
 }


martanne:
Could you provide an example which your patch tries to improve upon? I think `O` in normal mode should already take care of inserting a newline at the start of the line?

The de-indent issue (see also #383) is "problematic" because it has to remember state of what happened before. For example between inserting the spurious newline and leaving insert mode the cursor(s) could be moved around.

In general I'm not a big fan of adding this kind of "smartness" to the editor core. If you want to enforce consistent formatting of code, you are probably best off with a syntax aware formatter invoked from a pre-save event.

aeosynth:
if #430 is fixed, auto indent could be implemented in lua, keeping the editor core free of smartness

martanne:
This "limited" (non-language aware) smartness is tolerable in the editor core. I agree your proposed behavior seems preferable. However, in its current form it breaks `O` i.e. #272 is once again an issue.

aeosynth:
in which language (besides [whitespace](https://en.wikipedia.org/wiki/Whitespace_(programming_language))) are indented blank lines not frowned on? i don't think this needs language awareness.

this is a 5 line wip, and `O` can be fixed. should i fix it, ie, should i keep working on this pr?

martanne:
I think we are talking past each other a bit. The functionality is fine and I encourage you to fix the remaining issues. Ideally `o` and `O` would both work with (and without) a count and be correctly repeatable using `.`

In the meantime I implemented the deindentation upon leaving insert mode. It should work for the basic cases. You should rebase your changes on top of master.

aeosynth:
sorry, time issues

Issue #460, opened 2017-01-15
Missing digraphs feature


KolesnichenkoDS:
In Vim I can use `<C-k>` shortcut in insert mode to insert special characters ([digraphs](http://vimdoc.sourceforge.net/htmldoc/digraph.html)):

```viml
<C-k>l*    " for λ
<C-k>FA    " for ∀
<C-k>(-    " for ∈
```

joshaw:
I think this would be best implemented in lua as a plugin, it's basically
just a long list of mappings all sharing the `<C-k>` prefix.

On 15 Jan 2017 3:46 pm, "Daniil Kolesnichenko" <notifications@github.com>
wrote:

In Vim I can use <C-k> shortcut in insert mode to insert special characters
(digraphs <http://vimdoc.sourceforge.net/htmldoc/digraph.html>):

<C-k>l*    " for λ<C-k>FA    " for ∀<C-k>(-    " for ∈

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/martanne/vis/issues/460>, or mute the thread

martanne:
With the latest changes something like the following should work:

```Lua
vis:map(vis.modes.INSERT, "<C-k>", function(keys)
        if #keys < 2 then
                return -1 -- need more input
        end
        local digraph = keys:sub(1, 2)
        if digraph == "l*" then
                vis:feedkeys('λ')
                return 2 -- consume 2 bytes of input
        end
end, "Insert digraph")
```
How to (space) efficiently store and lazy load the diagraph table upon first use is left as an exercise for the reader.

However, it might make more sense to implement this as a standalone utility which can be used from multiple applications. It would get the digraph as argument and print the corresponding symbol to stdout. I believe such a tool was once posted to the suckless mailing list, I will have to check my mail archive.

josuah:
This is what I found about digraphs on suckless.org mailing list archives: https://duckduckgo.com/?q=digraph+site%3Ahttp%3A%2F%2Flists.suckless.org%2Fdev%2F&t=fpas&ia=web

> ^Au
> echo "bind u digraph 'U+'" >>~/.screenrc

They talked about a screen command in this case and not a whole program.

I can not promise a lot, but I could try to propose some very basic digraph program with someheader with:

``` C
struct digraph *digraphs[] = {
        { .digraph = "pp", .content = "¶" },
        { .digraph = "*p", .content = "π" },
        ...
}
```

It could be used for dvtm as well.

[...]

PR #475, opened 2017-01-27
Added @martanne's vis-digraphs program from #460

josuah:
I am not sure if this is how you want to integrate it to the tree, so feel free to adjust (or to ask me to do so).

This works for digraphs (not for more than 2 characters):

```
make clean
./configure
make
./vis
i<c-k>L*
```


martanne:
Thanks, some preliminary comments. I will try to take a closer look later on:

* I would prefer if `vis-digraph.h` would be merged directly into `vis-digraph.c`
* The man page has some issues
  - The characters are not necessarily UTF-8 but output according to the user locale
  - The various exit status should be mentioned
  - The `$LC_CTYPE` environment variable should probably be mentioned
  - Just mentioning standard input is enough, no need to talk about interactive keyboard input etc
  -  Some unrelated parts from `vis-menu` seem to have crept in? Run `make man` and proof read & check the formatting in a browser.

You can force push these changes into the corresponding patches.

Also what is the problem with digraphs longer than 2 characters? (This should probably be discussed in the issue?)

josuah:
I am making these changes.

I do not know if I can fill the multiple commits with appropriate changes, but I can merge them all in one commit.

josuah:
My bad, `vis-digraph` works everywhere and for any length of char.  I just had to recompile the with data from the RFC.

martanne:
Thanks! I merged your stuff with a different commit message (github handles aren't really descriptive) and some slight white space tweaks. The author section of the man page was removed and vis-digraph is now also referenced from `vis.1`.

josuah:
You are welcome, and thank you for fixing it all up.

PR #494, opened 2017-02-15
vis: assemble string of compile time features at compile time

eworm-de:


raphael-proust:
Isn't the first form optimised by the compiler and ends up being the same?

Or, more pointedly, what compilers/compiler options prevent the optimisations that make both form equivalent?

eworm-de:
No, this is not optimized - I looked at the binary. I do not know if any compiler can optimize this. Probably the condition is optimized, but separate `fputs()` calls remain.

PR #498, opened 2017-02-23
Rename `caml` to `ocaml`

raphael-proust:
- The lexer supports the OOP parts of the language (it recognises keywords
  such as `class`, `method`, etc.).
- The official name of the language is “OCaml”.


martanne:
What is your motivation? Do you want to fix a specific issue? Like being able to do `:set syntax ocaml`? We could do that by adding another layer of indirection. I used the existing filenames because it was the simplest thing to do. Hopefully most of the time the auto detection should take care of it anyway.

From time to time I try to sync the lexers with the [upstream source](https://foicica.com/hg/scintillua/file/c34bf00e89af/lexers). I usually just copy the source tree over and then do some `git diff` / `git add -i` surgery. For this reason I would like to avoid unnecessary (file name) changes if possible.

raphael-proust:
I'll try to get upstream to change the name.

No motivation other than it's a language I use and I noticed the name was not standard.

Issue #350, opened 2016-06-29
gf `path` fails


aeosynth:
`gf` treats enclosing backticks as part of the path:

    `path/to/foo`

`gf` on this path errors with

    File ``path/to/foo`' not found

Issue #346, opened 2016-06-29
gf doesn't recognize ~


aeosynth:
`:e ~/foo` works, but `gf` with cursor on `~/foo` doesn't.

PR #502, opened 2017-02-26
vis: reimplement `gf` and `<C-w>gf` functionality in lua

Shugyousha:
The functionality is not exactly identical since both bindings open the
file name under the cursor in a new window. The older implementation
opened the file in the same window with `gf` and in a new one with
`<C-w>gf`.

I added this key binding to the standard mapping because I use it a lot. Adding it as a plugin instead may make sense.

### DIFF
--- a/lua/vis-std.lua
+++ b/lua/vis-std.lua
@@ -146,3 +146,15 @@ vis:map(vis.modes.INSERT, "<C-k>", function(keys)
 	end
 	return #keys
 end, "Insert digraph")
+
+vis:map(vis.modes.NORMAL, "gf", function(keys)
+	local r = vis.win.file:text_object_longword(vis.win.cursor.pos)
+	local filename = vis.win.file:content(r)
+	local ok = vis:open(filename)
+	if not ok then
+		vis:info("Could not open file " .. filename .. " for reading.")
+	end
+	return #keys
+end, "Open file under cursor in a new window")
+
+vis:map(vis.modes.NORMAL, "<C-w>gf", "gf")
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -2308,6 +2308,7 @@ void vis_lua_init(Vis *vis) {
 		const char *name;
 	} textobjects[] = {
 		{ VIS_TEXTOBJECT_INNER_WORD, "text_object_word" },
+		{ VIS_TEXTOBJECT_INNER_LONGWORD, "text_object_longword" },
 	};
 
 	for (size_t i = 0; i < LENGTH(textobjects); i++) {


martanne:
The old implementation used the `:open` and `:edit` commands. The question is whether it makes sense to provide a Lua API for this (like the current `vis:open` you use) or rely on the editor commands (as in 
`vis:command(string.format("open '%s'", t:gsub("'", "\\'"))` (untested). The latter variant would also take care of the case when unsaved changes exist.

As for the filename extraction. Using the text object `iW` might not be ideal, for example it doesn't handle quoted filenames (see also #346 and #350). Part of the reason why I prefer a Lua implementation is that it should be more flexible. I would eventually like to support addresses appended to the end of file names (I will have to check the exact syntax acme uses for this). For example it should be possible to do `gf` on a compiler error message and have the cursor moved to the corresponding file/line. This might require some new Lua APIs, because `:e` (reloading the file) would for example lose undo history. It might be best to use an LPeg pattern together with the relatively new `file:match_at()` method.

Shugyousha:
On Sun, Feb 26, 2017 at 01:09:04PM -0800, Marc André Tanner wrote:
> The old implementation used the `:open` and `:edit` commands. The
> question is whether it makes sense to provide a Lua API for this
> (like the current `vis:open` you use) or rely on the editor commands
> (as in `vis:command(string.format("open '%s'", t:gsub("'", "\\'"))`
> (untested). The latter variant would also take care of the case when
> unsaved changes exist.

I wasn't aware of vis:command. Using that should be a more general
approach which avoids the need to implement some more specific Lua API
for this.


> As for the filename extraction. Using the text object `iW` might not
> be ideal, for example it doesn't handle quoted filenames (see also
> #346 and #350). Part of the reason why I prefer a Lua implementation
> is that it should be more flexible. I would eventually like to support
> addresses appended to the end of file names (I will have to check the
> exact syntax acme uses for this). For example it should be possible to
> do `gf` on a compiler error message and have the cursor moved to the
> corresponding file/line. This might require some new Lua APIs, because

There is a vim plugin that does interprets the file.c:123 pattern in
order to jump to the line number mentioned after the file name. I agree
that this would be great to have in vis.


> `:e` (reloading the file) would for example lose undo history. It
> might be best to use an LPeg pattern together with the relatively new
> `file:match_at()` method.

file:match_at does not seem to be mentioned at
http://martanne.github.io/vis/doc/ yet. Using LPeg to get the file
name&address should definitely be possible though.

I will try to find the time to look into implementing a LPeg based
"filename.c:123" format extractor and then rewrite the function to use
vis:command to open the file and jump to the given address.


Cheers,

Silvan

martanne:
> I wasn't aware of vis:command. Using that should be a more general
approach which avoids the need to implement some more specific Lua API
for this.

It might still be needed/useful for other things.

> There is a vim plugin that does interprets the file.c:123 pattern

The line number was just an example. All sam addresses should be usable e.g. `file.c:/main`
(not sure about the exact syntax) should jump to the first occurrence of `main`. I think Acme
supports this, the mouse based "hyperlink" navigation is one of its core strengths.

The goal is obviously not to replicate the C parser, but to reuse it in some way.
Possibly through `vis:command`.

> file:match_at does not seem to be mentioned at http://martanne.github.io/vis/doc/ yet

Yes the online documentation is currently only updated when I explicitly commit+push it.
As described in the [developer section of the wiki](https://github.com/martanne/vis/wiki/Developer-Overview#lua-api-documentation) you can re-generate it locally (assuming you have the necessary tools installed) with `make luadoc`.

kopischke:
>There is a vim plugin that does interprets the file.c:123 pattern in order to jump to the line number mentioned after the file name. I agree that this would be great to have in vis.

Two, actually, the older _[file_line](https://github.com/bogado/file-line)_ and the newer _[vim-fetch](https://github.com/kopischke/vim-fetch)_ (that one is mine). Both are more or less horrible hacks on top of the Vim’s script and event API, so I am very interested to see if _vis_’ much cleaner structure allows for a better solution.

Shugyousha:
On Sun, Feb 26, 2017 at 11:47 PM, Marc André Tanner
> > There is a vim plugin that does interprets the file.c:123 pattern
>
> The line number was just an example. All sam addresses should be usable e.g.
> file.c:/main
> (not sure about the exact syntax) should jump to the first occurrence of
> main. I think Acme
> supports this, the mouse based "hyperlink" navigation is one of its core
> strengths.

Agreed. It's one of the few things I like about Acme to be honest.


> The goal is obviously not to replicate the C parser, but to reuse it in some
> way.
> Possibly through vis:command.

I have seen that text-objects.c still contains the filename detection
function. Is that what you mean by the "C parser"? If so, that
function is not enough if we want to detect the ':([0-9]+|/[:alpha:])'
pattern after the file name of course.


> file:match_at does not seem to be mentioned at
> http://martanne.github.io/vis/doc/ yet
>
> Yes the online documentation is currently only updated when I explicitly
> commit+push it.
> As described in the developer section of the wiki you can re-generate it
> locally (assuming you have the necessary tools installed) with make luadoc.

Writing an LPeg pattern for the file name & addressing detection and
then use match_at and vis:command should be enough to be able to
implement this functionality. Would you accept such a patch?

martanne:
> I have seen that text-objects.c still contains the filename detection
function. Is that what you mean by the "C parser"?

No, it is still there because it is used for the prefix extraction of the file completion functionality `<C-x><C-f>` (which should probably also be moved to Lua?).

I meant the [address parsing logic of sam.c](https://github.com/martanne/vis/blob/da859e4c5d97d3ac38587f5d5ea5ea013b6f3808/sam.c#L747) which is used implicitly when invoking `vis:command`.

I haven't used Acme for a while, checking how it parses the address part (skipping until the next space/newline?) will likely be helpful.

Using an LPeg pattern with `match_at` followed by a `vis:command` sounds like a reasonable first approach.

mcepl:
It works perfectly now. Thank you.

Shugyousha:
Matěj Cepl ***@***.***> wrote:
> This is replaced by #558.

Yes, I will close this one.

Cheers,
Silvan

PR #504, opened 2017-02-26
vis-lua: make vis:win assignable

arames:
I added this to write a plugin that executes commands in all windows. Tested with

```
require('vis')

local module = {}

function module.update()
	local active_window = vis.win
	for win in vis:windows() do
		vis.win = win
		if win == active_window then
			vis:command('set relativenumber on')
		else
			vis:command('set number on')
		end
	end
	vis.win = active_window
end

return module
```

### DIFF
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1169,6 +1169,11 @@ static int vis_newindex(lua_State *L) {
 			vis_count_set(vis, count);
 			return 1;
 		}
+
+		if (strcmp(key, "win") == 0) {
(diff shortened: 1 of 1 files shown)


martanne:
Thanks, you should call `vis_window_focus`.

arames:
Done!

arames:
What does the return value indicate? It doesn't look like it is 0(success) vs non-zero (error).

arames:
Sorry I'm new to lua, I found the answer myself by looking at the lua `lua_CFunction` definition.

PR #506, opened 2017-02-27
Working Lua Code for incremtent/decremt of int,oct,hex

demonking:


martanne:
Thanks again, I did some additional cleanups:
* tabs for indentation, in general more consistent white space usage
* moved to separated file
* removal of unneeded variables
* removal of float support, this doesn't work anyway because the used text object will never match a number with a dot `.` also it is just a bad idea
* `string.format` doesn't handle negative values for hex/oct, I just rounded it to zero for now

I also mentioned some things which could possibly be improved in the commit message of f9cec7b86a6db8ea5c40dd3f930561591bd5362c.

I personally don't really use the feature, therefore I hope users who care about it will eventually fix the issues they encounter.

PR #507, opened 2017-02-27
Honor user's umask when saving files

michaelforney:


martanne:
Thanks, the reason for the unnecessary umask changes was that apparently [ancient glibc](http://man7.org/linux/man-pages/man3/mkstemp.3.html#NOTES) created the temporary file with mode 0666. I agree that relying on the POSIX semantics of `mkstemp(3)` is the right thing to do.

PR #514, opened 2017-03-07
freshen vis.1

josuah:
I propose formatting change for `vis.1`:

- [`Ad`](http://man.openbsd.org/mdoc#Ad) for register and command address
- [`Ic`](http://man.openbsd.org/mdoc#Ic) (Internal or interactive command) for keybindings and all `:commands`
- [`Rs`](http://man.openbsd.org/mdoc#Rs) for articles reference at the end.
- Using more compact lists for keybindings (`.Bl -compact`).
- Fix all `mandoc -Tlint ./man/vis.1` warnings.
- Homogenize the source.

You may ask for any edit you want me do apply before the merge, or simply refuse if you dislike.

### DIFF
--- a/man/vis.1
+++ b/man/vis.1
@@ -3,44 +3,48 @@
 .Os Vis VERSION
 .
 .Sh NAME
+.
 .Nm vis
 .Nd a highly efficient text editor
 .
 .Sh SYNOPSIS
+.
 .Nm
 .Op Fl v
-.Op Ic +command
+.Op Cm + Ns Ar command
 .Op Fl -
 .Op Ar files ...
 .
 .Sh DESCRIPTION
+.
 .Nm
-is a highly efficient screen-oriented text editor combining the strengths of both
+is a highly efficient screen-oriented text editor combining the strengths of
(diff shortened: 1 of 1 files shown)


martanne:
Thanks for working on documentation related things. Unfortunately I'm not that familiar with mdoc and therefore can't really judge whether all of these changes make sense (e.g. the documentation for `Ad` you linked mentions "Memory address", I don't know what that implies and whether it is a good fit for "editor addresses"). Maybe others can give their assessment.

From a user perspective the rendering with `man(1)` and the [html version](http://martanne.github.io/vis/man/vis.1.html) as generated by `make man` seems most important. As an example your reference section now also displays the URLs in plain text not only as a hyperlink.

In general you should separate the formatting changes from the content changes (e.g. mode description etc) into individual commits.

josuah:
The idea with `Ad` was to have greater diversity rather than using `Ic` for everything.  As there is no *editor adress* markup, I brought back `Ic`.

I also split mode description and the formatting in two commits.

PR #522, opened 2017-03-22
vis-lua: add usage documentation to command_register

joshaw:
Taken directly from commit message introducing the feature, useful to have particularly for function arguments.


martanne:
I didn't add it at the time because it is a bad example, most people probably don't even realize that the `print` output goes to standard out etc. Properly documenting the function signature would be more useful, but that probably leads to a fight with the LDoc format...

Also I'm not really happy with API, but I guess it is better than nothing.

PR #523, opened 2017-03-23
added matlab/octave to filetypes

theNerd247:


martanne:
Your Lua pattern is missing the `$` anchor, it will also match stuff like `this.is.my.extension`. The mime type is probably `text/x-matlab`. Furthermore, the `*.m` extension is already used by [Objective-C](https://github.com/martanne/vis/blob/c9a259bbe118d60558a8891e0bd701e8fb0ba91d/lua/plugins/filetype.lua#L253).

You should be able to override the setting from a custom window open event handler in your `visrc.lua`.

PR #553, opened 2017-05-16
Remove leading space in print statement for curses check

george-b:
Not sure if this war erroneous or not? Either way it's totally cosmetic.


erf:
I think it's there to group `libcurses` variants. Same for `liblua`.

george-b:
So it is, closing.

PR #559, opened 2017-05-23
man: Fix a typo

Two-Finger:
(I'm not sure if plural or singular is correct in this case...)


martanne:
This was merged in 48aad577d0525755d2e57c304475ee35f63ca12d.

PR #560, opened 2017-05-26
Try to fix jump match issue

realwhz:
In nested string and char scenarios, the current jump match doesn't work
well.  Examples are
```
{
	"\""
}

{
	'"'
	'\''
}

"hell'o'"
```
where `%` cannot correctly jumps between corresponding '{' and '}'.


text-motions.c
```diff
@@ -506,33 +506,40 @@ size_t text_bracket_match(Text *txt, size_t pos) {
...
 					return it.pos;
 				else if (c == current)
 					count++;
 			}
+			prev_c = c;
 		}
 	} else { /* backwards */
 		while (text_iterator_byte_prev(&it, &c)) {
-			if (c != current && c == '"')
+			if (!inchar && c != current && c == '"' && next_c != '\\')
```
martanne: This looks like the wrong way around?

When you encounter a quote `"` while iterating backwards you would have to check the previous (not next) character for a slash `\`.

More generally, this heuristic based approach will always have some issues ...

realwhz: One of the changes is to ignore quoted, namely `\"`.  When we iterate backward, isn't` \` the next to `"`?  Based on my test with the minimalist examples, this logic works.  A real example is function `text_object_bracket(Text *txt, size_t pos, char type)` in vis's source code file text-objects.c, where jump operator `%` cannot work.

PR #568, opened 2017-05-31
fix format string for 32bit

eworm-de:
The compiler reports:

main.c:2007:32: warning: format '%x' expects argument of type 'unsigned int',
but argument 3 has type 'wchar_t {aka long int}' [-Wformat=]
    buffer_appendf(&info, "U+%04x ", wc);


Shugyousha:
wchar_t size varies (it's a 'short int' on Windows) but since vis does not support non-POSIX systems and there does not seem to be a standard length modifier for wchar_t your commit looks okay to me.

PR #573, opened 2017-06-01
build: produce reproducible self contained executable

eworm-de:
Giving tar the parameter '--sort=name' sorts direcory entries, but keeps
single files as-is. So instead sort the list retrieved by find.

Works for me. [TM] :D


eworm-de:
Building a self-contained `vis` executable (`make docker`) from 6ad1a9739fe685e2ed9ca2c515b15e772f3f12b2 should result in this version output:
```
# ./vis -v
vis v0.3-111-g6ad1a97 +curses +lua +lpeg +acl
```
... and this `sha256` checksum:
```
# sha256sum vis
8ec450e1307e9de17df47db8259060a31d09eabac0ab7a42b3d93d91b72d3e8c  vis
```
Anybody wants to verify?

PR #575, opened 2017-06-04
Add option to hide EOF marker

p-e-w:
Allows the user to type

```
:set show-eof off
```

to hide the tildes used as placeholders for empty lines.

It appears that such an option was already planned because the constant `UI_OPTION_SYMBOL_EOF` was already present in the code, it just wasn't implemented.


p-e-w:
I found a bug: When you

```
:set number
```

with this branch, the EOF markers disappear. Not sure why this is so, I thought I had found every instance where view options are created. Pity there is no idiomatic way to define options defaults in code.

PR #579, opened 2017-06-10
More theme improvements

p-e-w:
Introduces `STYLE_LINENUMBER_CURSOR`, which allows for separate styling of the line number for the line the cursor is on. This makes it easier to see where the cursor is.

Also dims the color used to draw invisible character placeholders to distinguish them better from the text.

Both of these changes are standard features commonly found in code editors.

theme_tweaks


martanne:
I assume you want to use a different style than [`STYLE_CURSOR_LINE`](https://github.com/martanne/vis/blob/081d99dda082a360dad367e8e2cc036a738edca5/lua/themes/solarized.lua#L56)? (Otherwise you could reuse the style used for `:set cul`).

The reason I'm a bit hesitant to merge this and #575 is that in the long term I would like to try to move more of the drawing code into Lua. The basic idea would be to provide a sort of "canvas" where both document content and unrelated text (e.g line numbers, EOF markers, inline annotations, gutter indications, folding, concealed symbols etc.) can be drawn.

However, this will require some more thought and a large refactoring of cursor positioning. For now I should probably just merge it.

p-e-w:
> I assume you want to use a different style than STYLE_CURSOR_LINE? (Otherwise you could reuse the style used for :set cul).

Yes, they should be different IMO, in fact a common theme design (not the one used here though) seems to be to invert the colors of the cursor line for the cursor line number.

I agree that much of the drawing code should move to Lua (which might also be a good opportunity to rethink libtermkey and curses, as with the current architecture adding good quality mouse support appears very difficult). At that point the Lua vs. LuaJIT discussion will probably become relevant again. In general though, the more functionality is in Lua, the better. This includes many of the text operations as well.

PR #584, opened 2017-06-14
Allow opting out from suggested CFLAGS and LDFLAGS

czarkoff:
This PR adds new `configure` flag `--enable-default-flags` (and `--disable-default-flags` counterpart), so that user may opt out from optimizations. This is required for packaging when downstream wants to control the flags that are used.

### DIFF
--- a/configure
+++ b/configure
@@ -218,10 +218,12 @@ fi
 #
 # Figure out options to force errors on unknown flags.
 #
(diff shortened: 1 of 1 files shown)


martanne:
This seems rather ugly/unnecessary. Which flags do you not want? And why can't you override them during make invocation?

```
make CFLAGS=... LDFLAGS=...
```

PR #585, opened 2017-06-17
Gutter API

p-e-w:
Introduces a new Lua API, `gutter_column_add`, that allows plugins to render text in the gutter. This enables new classes of plugins to be written, including linters and VCS helpers.

### Example: (Fake) Git diff plugin

```lua
local GREEN = 100
local RED = 101
local YELLOW = 102

vis.events.subscribe(vis.events.WIN_OPEN, function(win)
  vis:command("set gutter")
  win:style_define(GREEN, vis.lexers.STYLE_LINENUMBER..",fore:#00dd00,bold")
  win:style_define(RED, vis.lexers.STYLE_LINENUMBER..",fore:#ff0000,bold")
  win:style_define(YELLOW, vis.lexers.STYLE_LINENUMBER..",fore:#ffff00,bold")
end)

vis:gutter_column_add("git", 2, function(win, line)
  if line == 6 or (14 <= line and line <= 18) then
    return "+ ", GREEN
  elseif line == 3 then
    return "_ ", RED
  elseif 10 <= line and line <= 11 then
    return "~ ", YELLOW
  else
    return "", win.STYLE_LINENUMBER
  end
end)
```

gutter-1

Arbitrarily many gutter columns can be added, each with independent rendering. This makes it possible to display text with multiple styles:

gutter-2

An attractive extension to this API could be to make the `width` argument a function (receiving the window object) instead of a fixed value. The column width could then be dynamically computed, which among many other uses would let us **render the line numbers completely in Lua(!)** and kick all line number-related code out of the C core (incidentally, there is a subtle bug in the current line number rendering, but that is another matter).

Something I'm not sure about is how this impacts rendering performance. Each gutter column will result in one Lua function invocation per line per render. In my subjective tests, I didn't notice any visible overhead, but I have no idea how to measure this objectively.

I have tried to honor the superb quality of your codebase by documenting everything and maintaining the existing style as much as I could. Still, this is a rather large change so please let me know if you are not happy with any details or with the overall approach. You mentioned that you want to move more drawing code to Lua, and I hope this is in line with what you had in mind. I have thought about this for some time and I feel that APIs like the one in this PR strike a nice balance between static configuration (which is what we currently have) and low-level "draw text `s` at `x` and `y`" calls.

### To do

- [ ] Support text containing Unicode characters. No idea how to do that.


martanne:
Thanks for looking into this, I agree that it should eventually be possible to manage gutter symbols.

However, I feel like we should probably first discuss some use cases.

In your proposed API the symbols would need to be specified on every redraw. I realize this is convenient for the `git diff` scenario. I personally don't really need this functionality, I would just open a new terminal and run git diff there when needed. (Isn't it a bit excessive to run an external command on every redraw!?) Wouldn't it make more sense to use something like marks (or more general "text properties") to set the symbols once and then have them move with the text as it is modified. This would be more tailored to use cases where the symbols are updated more rarely (e.g. after a file is saved).

Ideally the C code shouldn't know about gutter symbols at all. This should all be done on the Lua side. I thought about some `WIN_DRAW` event which would get the starting position from which the file content should be drawn. Then the Lua code would be responsible for the whole appearance (including the sidebar with line numbers and the gutter symbols). There would be APIs to draw text from the buffer content `win:draw_text(content, start, end)` and arbitrary content `win:draw_string(content)`. This would require a big refactoring of the cursor placement code, which currently assumes that the displayed text is always contiguous. However, it seems necessary for things like code folding. So there probably would be a low-level "draw text s at x and y" style API upon which we can layer more abstractions, if needed. These are just my ideas, I'm not sure if/how they will work in practice.

Using fixed style IDs for `win:style_define` seems like a bad idea. It is necessary for the lexer module which uses some hard coded styles. But in general we should provide a function which uses the first currently unused ID to prevent conflicts.

From an API perspective it would be nice if there were no need to specify a column width. It should be adjusted dynamically depending on the content (with a reasonable maximum).

What problems did you encounter with Unicode symbols? Why don't they work as is?

Thanks for keeping the style consistent and following existing conventions. Your contributions are of high quality, but I will need to spend some more time thinking about general architectural design decisions.

p-e-w:
Fair enough. I'm closing this then so it doesn't clutter the PR backlog.

However, I would ask that you reconsider merging #579 (and #575, as soon as I work out the bug). Those are really minor features that are standard in most other editors. No matter how much the architecture changes in the future, it should be straightforward to keep those features around, and until then they are available already.

djanowski:
👋 Has there been any development regarding a gutter API, or how to integrate a code linter?

PR #592, opened 2017-07-04
Remove dependency dir from .gitignore

george-b:
With the removal of the standalone build target I think this can also also.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /config.h
 /config.mk
-/dependency
 /vis
 /vis-menu
 /vis-single


george-b:
I noticed this as I still had this directory in my local repo after pulling the latest changes. I guess this was removed upstream but no action was taken on cloned repos due to it being in `.gitignore`

martanne:
Hi, the directory was never under version control, hence it was also never "removed upstream".

In fact, it is still used by the `make local` stuff which in turn is used by the [AppVeyor](https://ci.appveyor.com/project/martanne/vis/branch/master), [Travis CI](https://travis-ci.org/martanne/vis) and [Coverity Scan](https://scan.coverity.com/projects/3939) builds.

george-b:
> In fact, it is still used by the make local stuff which in turn is used by the AppVeyor, Travis CI and Coverity Scan builds.

Whoops missed that, sorry.

PR #593, opened 2017-07-05
Add s/S normal mode bindings

aeosynth:
vim compat

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -260,6 +260,8 @@ static const KeyBinding bindings_normal[] = {
 	{ "q",                  ACTION(MACRO_RECORD)                        },
 	{ "R",                  ACTION(MODE_REPLACE)                        },
 	{ "r",                  ACTION(REPLACE_CHAR)                        },
+	{ "s",                  ALIAS("cl")                                 },
+	{ "S",                  ALIAS("cc")                                 },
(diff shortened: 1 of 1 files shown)


martanne:
Hi, we are moving to a more selection oriented workflow (yes documentation is currently lacking), being able to quickly save and restore selections to/from registers is an important part of that. I thus prefer to keep the current bindings. Remember the goal is not to be vi(m) compatible just for the sake of it.

Also as is your change breaks `gv` which restores selections from the `^` register. That is my mistake as I used `S` instead of `<vis-selections-restore>` in the alias. (We should probably review all existing aliases).

I'm open for improved key binding suggestions, but `s` and `S` are short mnemonics and the vim aliases seemed rather useless to me (and I hardly every use them).

martanne:
After some discussion on IRC we moved the the mark bindings to:

 * `'{mark}m` to save selections
 * `'{mark}M` to restore selections

This is akin to register usage i.e. `"{reg}y` yank and `"{reg}p` put. Similarly, if `{mark}` is omitted a default mark is used.

This means we can in principle restore the original `s` and `S` mappings for now. Previously they were aliased to `cl` and `^c$` respectively.

Comments?

aeosynth:
Yeah I would like `s`/`S` mappings; should I redo the PR?

PR #605, opened 2017-09-14
vis.1:  Surround with quotes link-name arguments of .Lk

PuercoPop:
As noted on mdoc(7), in groff  "Lk only accepts a single link-name argument; the remainder is misformatted.
The quotes makes groff treat the title as a single argument.

Note that double quotes where originally used, but where remove in
https://github.com/martanne/vis/commit/bee050696651d7740a54644f11adf21825a8c7f2#diff-1b1a076db28be546dfe17f39e05e60e1L1292


josuah:
You are right and I did not know.  Thank you.  :)

PR #647, opened 2017-12-25
Add clojure lexer and filetype support

Gavlooth:
Some basic syntax hightlight for clojure  and clojurescript :)


Gavlooth:
Yes indeed

PR #654, opened 2018-01-01
vis-open: fix for absolute and non-existent paths

kj:
I've had some issues with vis-open and filename completion. As an example, try opening vis in your home directory and typing `:<Tab><Tab><Tab>`. You should end up with something like: `/home/$USER//:***`.

When the shell cannot find any matching files, the glob is not expanded, and vis-open will return the absolute path of the current working directory (because `dirname` prints `.`), followed by the filename, followed by a literal `*`. This commit checks that the final path actually exists, and if not, exits with status 1.

It also uses `text_object_longword` instead of `text_object_word` for the range to match, so that absolute paths are accepted, and replaced properly (else it only matches back to the first '/', and replaces from there. For example, in insert mode, try to  enter `/etc/host<C-x><C-o>`.

Please correct me if I've misunderstood anything, or have made any mistakes!

### DIFF
--- a/lua/plugins/complete-filename.lua
+++ b/lua/plugins/complete-filename.lua
@@ -6,7 +6,7 @@ vis:map(vis.modes.INSERT, "<C-x><C-f>", function()
 	local pos = win.selection.pos
 	if not pos then return end
 	-- TODO do something clever here
-	local range = file:text_object_word(pos > 0 and pos-1 or pos);
+	local range = file:text_object_longword(pos > 0 and pos-1 or pos);
 	if not range then return end
 	if range.finish > pos then range.finish = pos end
 	if range.start == range.finish then return end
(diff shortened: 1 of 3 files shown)


kj:
I had to expose `text_object_longword` to the Lua API though for that to work.

martanne:
Thanks for your contribution!

I agree that this should be improved. However, I'm not sure whether failing in case the path does not exist is the best solution. For example it prevents the use case of creating a new file in a directory selected by `vis-open`:

```
:open /
# navigate to the desired directory
# finally type the new file name and confirm by pressing enter
```

kj:
That's a good point! I made a few changes which allow for that, but it is still a bit weird. I made it trim any trailing glob character from the output. I'm not sure whether that is acceptable. It feels a bit hacky? This means it could never return a file which actually ends in an asterisk (if that matters). It just fixes that case where if you press <kbd>Tab</kbd> and it matches nothing, the `vis-open` call in complete-filename.lua will keep appending `*` to the path. Also, although it is not something you'd actually do in normal usage, if you open the command line with <kbd>:</kbd>, than press <kbd>Tab</kbd> immediately, it now treats the `:` as a 'new file' and prepends the absolute path to the current directory, which is pretty unexpected behaviour if you don't already know how it works (that the command line is just a regular buffer).

It's getting quite convoluted I feel. Maybe it just needs a complete rethink?

martanne:
> It's getting quite convoluted I feel. Maybe it just needs a complete rethink?

I agree completely, the current implementation is a mess. If we had a client/server architecture we could let the user use his regular (shell) environment to launch a new client connecting to an existing session ...

PR #657, opened 2018-01-09
build: use pkg-config to find lpeg

Two-Finger:
On my Debian system "-llpeg" was not the right flag to link to lpeg, so I changed `./configure` to use pkg-config.

Probably more package names should be added to the for loop before considering it for merging.
I added only those that I knew.

### DIFF
--- a/configure
+++ b/configure
@@ -509,11 +509,25 @@ EOF
 		$LDFLAGS $LDFLAGS_LUA $LDFLAGS_LPEG -o "$tmpo" >/dev/null 2>&1 ; then
 		CONFIG_LPEG=1
 		printf "yes\n"
-		break
 	else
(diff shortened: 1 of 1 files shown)


lanodan:
I don’t even know how your system got a pkg-config for lpeg, as there is none in lpeg tarball or https://anonscm.debian.org/git/pkg-lua/lua-lpeg.git (maybe I’m lost in debian packaging).
I think it should try different flags instead.

Two-Finger:
@lanodan The `.pc` files are in a separate package - `lua-lpeg-dev`.

martanne:
This was merged in 5879337bb6905540a8f0797aa1b084b9cdc1d9fa.

PR #660, opened 2018-01-24
lexers: improve scheme lexer

ii8:


martanne:
Thanks, at first glance it looks fine (I'm not that familiar with scheme).

However, could you please split the changes into 3 commits (scheme improvements, additional file extension association, pony changes) and provide a bit more context in the commit messages.

@TieDyedDevil do you have any comments regarding the lexer changes?

TieDyedDevil:
I tried the scheme lexer on several files from the Lilypond distribution. Looks good.

I have no concerns regarding the other two.

PR #661, opened 2018-01-25
vis-complete: Show just basenames in vis-menu

Two-Finger:
The common dirname part makes it harder to spot the right file, and fewer entries can be visible at once.

With only basenames, one can see quite a few more entries, and know in advance how many key presses (roughly) it would take to get to the desired file.

### DIFF
--- a/lua/plugins/complete-filename.lua
+++ b/lua/plugins/complete-filename.lua
@@ -6,12 +6,15 @@ vis:map(vis.modes.INSERT, "<C-x><C-f>", function()
 	local pos = win.selection.pos
 	if not pos then return end
 	-- TODO do something clever here
-	local range = file:text_object_word(pos > 0 and pos-1 or pos);
+	local range = file:text_object_longword(pos > 0 and pos-1 or pos);
 	if not range then return end
 	if range.finish > pos then range.finish = pos end
 	if range.start == range.finish then return end
 	local prefix = file:content(range)
 	if not prefix then return end
+	-- Strip leading delimiters for some languages
+	i, j = string.find(prefix, "[[(<'\"]+")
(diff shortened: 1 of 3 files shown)


martanne:
Thanks, I agree that the current completion implementation sucks.

However, using basename might lead to ambiguous results, consider multiple files with the same name in different sub directories as in:
```
mkdir foo1 foo2
touch foo1/file.c foo2/file.c
```
A completion of the  prefix `foo` will show two `file.c` entries which can't really be distinguished.

Stripping the current working directory might be a better option.

Two-Finger:
> Stripping the current working directory might be a better option.

I fixed it, in a bit more generic way - it strips the common part in all three cases, not just when completing in the current directory.

The regression was mostly because I didn't understand why the recursive behavior is there and chose to ignore it for the time being.

I find it a bit confusing that it descends into directories. If I write `foo` I'd expect to get completions for `foo1` and `foo2`, not their whole sub-trees.

Imagine that you vis-complete `/usr/share/g` (just to be sure that certain subdirectory exists). The current code will descend into many directories, and vis-menu will be full of entries, sorted "depth-first". It will not only be slow but it will be impossible to find the top-level entries, like `gimp`, `gettext`, `gdb`, etc., which are most likely what you want.

If a "flat" completion is done instead, it will be faster and only the most relevant completions will be presented. It can be implemented with a simple `echo $XPATTERN*`, instead of the `find` expression.
If one wants to descend, they can always press `<C-X><C-F>` again.

Two-Finger:
On another note, I just noticed that `lua/plugins/complete-filename.lua` doesn't send the whole preceding path to `vis-complete`, and completing absolute and tilde-expanded paths from within vis doesn't seem to work.

Using `vis-complete` directly works.

Issue #669, opened 2018-02-12
creating a visrc.lua stomps default configuration


xificurC:
After a build from source I have solarized theme and syntax highlighting on by default. After creating `~/.config/vis/visrc.lua` I don't have either. Is this expected behavior? I couldn't find what the defaults are and how to set them again if that's what one should do after creating the local configuration file.

lanodan:
I don’t seen to have this issue here (with vis 0.4). What did you put in ``~/.config/vis/visrc.lua``?
An example/template configuration is present in ``/usr/local/share/vis/visrc.lua``.
What you can do is to (re)set the theme in your config, like that:
```lua
require("vis")

vis.events.subscribe(vis.events.WIN_OPEN, function(win)
        vis:command('set theme solarized-16')
end)
```

xificurC:
Thanks, it wasn't clear to me that I need to copy the structure of `/usr/local/share/vis/visrc.lua`. I just created a new file containing `vis:command(...)`. Putting it into the global or WIN_OPEN section solves it.

PR #671, opened 2018-02-15
Update vis.1

xificurC:
As seen in #669 I didn't *get it* one needs to have a special stub in `visrc.lua` for everything to work properly. Putting at least a few words in the man page might help some others like me in the future :)

### DIFF
--- a/man/vis.1
+++ b/man/vis.1
@@ -1506,7 +1506,9 @@ for a system-wide configuration provided by administrator.
 .Pa /usr/local/share/vis
 or
 .Pa /usr/share/vis
-depending on the build configuration.
+depending on the build configuration. When creating a new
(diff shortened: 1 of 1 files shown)


man/vis.1
```diff
@@ -1506,7 +1506,9 @@ for a system-wide configuration provided by administrator.
 .Pa /usr/local/share/vis
 or
 .Pa /usr/share/vis
-depending on the build configuration.
+depending on the build configuration. When creating a new
+.Pa visrc.lua
+be sure the copy the structure from here.
 .El
```
Shugyousha: should be

"be sure to copy the structure from here."

xificurC:
Thanks, I forgot each day to come back and make that fix.

PR #672, opened 2018-02-18
Support COLOR_PAIRS > SHRT_MAX

michaelforney:
In ncurses 6.1, the TERMINAL structure was updated[0] to store data in `int`
instead of `short`, and terminfo definitions for 256-color terminals were
updated from `pairs#32767` to `pairs#0x10000`.

However, since vis stores the value of COLOR_PAIRS in a short (ncurses
internally stores it as an int), it is now overflowing into negative, breaking
color support completely.

The standard `init_pair` entry points still use `short` for their parameters, so
just restrict the pairs to `SHRT_MAX` during allocation.

[0] http://invisible-island.net/ncurses/announce-6.1.html#h4-new-library

### DIFF
--- a/ui-terminal-curses.c
+++ b/ui-terminal-curses.c
@@ -41,7 +41,7 @@
 #  define NCURSES_EXT_COLORS 0
 # endif
 # if !NCURSES_EXT_COLORS
-#  define MAX_COLOR_PAIRS 256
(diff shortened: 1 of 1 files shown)


Shugyousha:
I can confirm that this patch fixes the color issues for me that I have when upgrading to ncurses 6.1.

PR #673, opened 2018-02-19
Lexer improvements (RFC)

roguh:
I've added Vagrantfile as a Ruby file extension, .yml as a YAML file extension, and a [Reason ML parser](https://reasonml.github.io/) based on the OCaml parser.

### DIFF
--- a/lua/lexers/reason.lua
+++ b/lua/lexers/reason.lua
@@ -1,0 +1,84 @@
+-- Copyright 2006-2017 Mitchell mitchell.att.foicica.com. See LICENSE.
+-- Reason LPeg lexer.
+
+local l = require('lexer')
+local token, word_match = l.token, l.word_match
+local P, R, S = lpeg.P, lpeg.R, lpeg.S
+
+local M = {_NAME = 'reason'}
+
+-- Whitespace.
+local ws = token(l.WHITESPACE, l.space^1)
+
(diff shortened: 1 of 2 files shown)


martanne:
Thanks for your contribution.

It looks like the reason lexer is not just "based on the OCaml" one, but is in fact an identical copy. Not sure whether that is intentional?

In any case we shouldn't duplicate the same source file. Instead we could:

* add a symlink `reason.lua -> caml.lua`
* decouple the lexer name from the file type:
    ```Lua
	reason = {
		lexer = "caml",
		ext = { "%.re$" },
	},
    ```
    and use the `lexer` property if available. This would also cover cases like #498 where the upstream naming could be improved.

Also the indentation in [your changes to filetype.lua](https://github.com/martanne/vis/pull/673/commits/930aa346f1bf00b8e56dd2b725dffd1876fcd487#diff-b01b59570f8c7ee5ff8e8936699af24aR306) seems off. We generally use tabs, not spaces. (This does not apply to the `lua/lexers/` directory.)

roguh:
Thank you for your detailed review.

The Reason lexer was not intentionally duplicated, Reason ML and OCaml are different (for example comments are `/* */` and `(* *)`, respectively). I've added the correct file and fixed the indentation in `filetype.lua`.

PR #675, opened 2018-02-26
[Lua API] Non-blocking subprocess runner

xomachine:
To improve the plugins API it would be nice to have a mechanism to run and communicate with a subprocess without whole editor hang. E. g. such a mechanism needed to make better implementation of my [Nim language plugin](https://github.com/xomachine/nis).
The plugin spawns a process which caries all information about source code and suggests/checks/gives info in interactive mode using stdin/stdout to communicate with caller. At the moment, the communication problem is solved by kicking vis from the shell using window resize signal and some other dirty hacks to make pipes non-blocking.

Attached implementation provides the `vis.communicate` function which starts a process in the shell and returns a file handle to write data to its `stdin`. If Lua code closes the file handle, the process is being killed. When the process writes anything to `stdout` or `stderr`, quits or gets signal - the `PROCESS_RESPONCE` event is being fired with corresponding arguments. The implementation allows multiple subprocesses running in the same time. They can be distinguished by the `name` argument passed to the `vis.communicate` function and to the event handler when the event fires. The responsibility of the `name` uniqueness is laid on the users code.

### DIFF
--- a/lua/vis.lua
+++ b/lua/vis.lua
@@ -152,6 +152,7 @@ local events = {
 	WIN_OPEN = "Event::WIN_OPEN", -- see @{win_open}
 	WIN_STATUS = "Event::WIN_STATUS", -- see @{win_status}
 	TERM_CSI = "Event::TERM_CSI", -- see @{term_csi}
+	PROCESS_RESPONSE = "Event::PROCESS_RESPONSE", -- see @{process_response}
 }
 
 events.file_close = function(...) events.emit(events.FILE_CLOSE, ...) end
@@ -167,6 +168,7 @@ events.win_highlight = function(...) events.emit(events.WIN_HIGHLIGHT, ...) end
 events.win_open = function(...) events.emit(events.WIN_OPEN, ...) end
 events.win_status = function(...) events.emit(events.WIN_STATUS, ...) end
 events.term_csi = function(...) events.emit(events.TERM_CSI, ...) end
+events.process_response = function(...) events.emit(events.PROCESS_RESPONSE, ...) end
 
 local handlers = {}
 
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -53,13 +53,6 @@
 #define debug(...) do { } while (0)
 #endif
 
-typedef struct {
-	/* Lua stream structure for the process input stream */
-	FILE *f;
-	lua_CFunction closef;
-	Process *handler;
-} ProcessStream;
-
 static void window_status_update(Vis *vis, Win *win) {
 	char left_parts[4][255] = { "", "", "", "" };
 	char right_parts[4][32] = { "", "", "", "" };
@@ -1407,6 +1400,14 @@ static int close_subprocess(lua_State *L) {
  * @return the file handle to write data to the process, in case of error the return values are equivalent to @{io.open} error values.
  */
 static int communicate_func(lua_State *L) {
+
+	typedef struct {
+		/* Lua stream structure for the process input stream */
+		FILE *f;
+		lua_CFunction closef;
+		Process *handler;
+	} ProcessStream;
+
 	Vis *vis = obj_ref_check(L, 1, "vis");
 	const char *name = luaL_checkstring(L, 2);
 	const char *cmd = luaL_checkstring(L, 3);
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1405,8 +1405,7 @@ static int communicate_func(lua_State *L) {
 
 	typedef struct {
 		/* Lua stream structure for the process input stream */
-		FILE *f;
-		lua_CFunction closef;
+		luaL_Stream stream;
 		Process *handler;
 	} ProcessStream;
 
@@ -1415,12 +1414,12 @@ static int communicate_func(lua_State *L) {
 	const char *cmd = luaL_checkstring(L, 3);
 	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
 	luaL_setmetatable(L, LUA_FILEHANDLE);
-	inputfd->handler = vis_process_communicate(vis, name, cmd, (void **)(&(inputfd->closef)));
+	inputfd->handler = vis_process_communicate(vis, name, cmd, (void**)(&(inputfd->stream.closef)));
 	if (inputfd->handler) {
-		inputfd->f = fdopen(inputfd->handler->inpfd, "w");
-		inputfd->closef = &close_subprocess;
+		inputfd->stream.f = fdopen(inputfd->handler->inpfd, "w");
+		inputfd->stream.closef = &close_subprocess;
 	}
-	return inputfd->f ? 1 : luaL_fileresult(L, inputfd->f != NULL, name);
+	return inputfd->stream.f ? 1 : luaL_fileresult(L, 0, name);
 }
 /***
  * Currently active window.
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -3229,10 +3229,10 @@ void vis_lua_process_response(Vis *vis, const char *name,
 		case EXIT:
 		case SIGNAL:
 			lua_pushinteger(L, len);
-			lua_pushlstring(L, buffer, 0);
+			lua_pushnil(L);
 			break;
 		default:
-			lua_pushinteger(L, 0);
+			lua_pushnil(L);
 			lua_pushlstring(L, buffer, len);
 		}
 		pcall(vis, L, 4, 0);
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -3207,8 +3207,9 @@ void vis_lua_term_csi(Vis *vis, const long *csi) {
  * The response received from the process started via @{Vis:communicate}.
  * @function process_response
  * @tparam string name the name of process given to @{Vis:communicate}
- * @tparam string|int buffer the available content sent by process; it becomes the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
  * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
+ * @tparam int the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
+ * @tparam string buffer the available content sent by process
  */
 void vis_lua_process_response(Vis *vis, const char *name,
                               char *buffer, size_t len, ResponseType rtype) {
@@ -3219,20 +3220,23 @@ void vis_lua_process_response(Vis *vis, const char *name,
 	vis_lua_event_get(L, "process_response");
 	if (lua_isfunction(L, -1)) {
 		lua_pushstring(L, name);
-		switch (rtype) {
-		case EXIT:
-		case SIGNAL:
-			lua_pushinteger(L, len);
-			break;
-		default: lua_pushlstring(L, buffer, len);
-		}
 		switch (rtype){
 		case STDOUT: lua_pushstring(L, "STDOUT"); break;
 		case STDERR: lua_pushstring(L, "STDERR"); break;
 		case SIGNAL: lua_pushstring(L, "SIGNAL"); break;
 		case EXIT: lua_pushstring(L, "EXIT"); break;
 		}
-		pcall(vis, L, 3, 0);
+		switch (rtype) {
+		case EXIT:
+		case SIGNAL:
+			lua_pushinteger(L, len);
+			lua_pushlstring(L, buffer, 0);
+			break;
+		default:
+			lua_pushinteger(L, 0);
+			lua_pushlstring(L, buffer, len);
+		}
+		pcall(vis, L, 4, 0);
 	}
 	lua_pop(L, 1);
 }
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -23,7 +23,6 @@
 
 #include "vis-lua.h"
 #include "vis-core.h"
-#include "vis-subprocess.h"
 #include "text-motions.h"
 #include "util.h"
 
@@ -1414,7 +1413,7 @@ static int communicate_func(lua_State *L) {
 	const char *cmd = luaL_checkstring(L, 3);
 	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
 	luaL_setmetatable(L, LUA_FILEHANDLE);
-	inputfd->handler = vis_process_communicate(vis, name, cmd, (void**)(&(inputfd->stream.closef)));
+	inputfd->handler = vis_process_communicate(vis, name, cmd, &(inputfd->stream.closef));
 	if (inputfd->handler) {
 		inputfd->stream.f = fdopen(inputfd->handler->inpfd, "w");
 		inputfd->stream.closef = &close_subprocess;
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -23,6 +23,7 @@
 
 #include "vis-lua.h"
 #include "vis-core.h"
+#include "vis-subprocess.h"
 #include "text-motions.h"
 #include "util.h"
 
@@ -51,6 +52,13 @@
 #else
 #define debug(...) do { } while (0)
 #endif
+
+typedef struct {
+	/* Lua stream structure for the process input stream */
+	FILE *f;
+	lua_CFunction closef;
+	Process *handler;
+} ProcessStream;
 
 static void window_status_update(Vis *vis, Win *win) {
 	char left_parts[4][255] = { "", "", "", "" };
@@ -162,6 +170,9 @@ void vis_lua_win_close(Vis *vis, Win *win) { }
 void vis_lua_win_highlight(Vis *vis, Win *win) { }
 void vis_lua_win_status(Vis *vis, Win *win) { window_status_update(vis, win); }
 void vis_lua_term_csi(Vis *vis, const long *csi) { }
+void vis_lua_process_response(Vis *vis, const char *name,
+                              char *buffer, size_t len, ResponseType rtype) { }
+
 
 #else
 
@@ -1368,6 +1379,47 @@ static int redraw(lua_State *L) {
 	return 0;
 }
 /***
+ * Closes a stream returned by @{Vis.communicate}.
+ *
+ * @function close
+ * @tparam io.file inputfd the stream to be closed
+ * @treturn bool the same with @{io.close}
+ */
+static int close_subprocess(lua_State *L) {
+	luaL_Stream *file = luaL_checkudata(L, -1, "FILE*");
+	int result = fclose(file->f);
+	if (result == 0) {
+		file->f = NULL;
+		file->closef = NULL;
+	}
+	return luaL_fileresult(L, result == 0, NULL);
+}
+/***
+ * Open new process and return its input handler.
+ * When the process will quit or will output anything to stdout or stderr,
+ * the @{process_response} event will be fired.
+ *
+ * The editor core won't be blocked while the external process is running.
+ *
+ * @function communicate
+ * @tparam string name the name of subprocess (to distinguish processes in the @{process_response} event)
+ * @tparam string command the command to execute
+ * @return the file handle to write data to the process, in case of error the return values are equivalent to @{io.open} error values.
+ */
+static int communicate_func(lua_State *L) {
+	Vis *vis = obj_ref_check(L, 1, "vis");
+	const char *name = luaL_checkstring(L, 2);
+	const char *cmd = luaL_checkstring(L, 3);
+	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
+	luaL_setmetatable(L, LUA_FILEHANDLE);
+	inputfd->handler = vis_process_communicate(vis, name, cmd, (void **)(&(inputfd->closef)));
+	if (inputfd->handler) {
+		inputfd->f = fdopen(inputfd->handler->inpfd, "w");
+		inputfd->closef = &close_subprocess;
+	}
+	return inputfd->f ? 1 : luaL_fileresult(L, inputfd->f != NULL, name);
+}
+/***
  * Currently active window.
  * @tfield Window win
  * @see windows
@@ -1524,6 +1576,7 @@ static const struct luaL_Reg vis_lua[] = {
 	{ "exit", exit_func },
 	{ "pipe", pipe_func },
 	{ "redraw", redraw },
+	{ "communicate", communicate_func },
 	{ "__index", vis_index },
 	{ "__newindex", vis_newindex },
 	{ NULL, NULL },
@@ -3145,6 +3198,35 @@ void vis_lua_term_csi(Vis *vis, const long *csi) {
 		for (int i = 0; i < nargs; i++)
 			lua_pushinteger(L, csi[2 + i]);
 		pcall(vis, L, 1 + nargs, 0);
+	}
+	lua_pop(L, 1);
+}
+/***
+ * The response received from the process started via @{Vis:communicate}.
+ * @function process_response
+ * @tparam string name the name of process given to @{Vis:communicate}
+ * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
+ * @tparam string|int buffer the available content sent by process; it becomes the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
+ */
+void vis_lua_process_response(Vis *vis, const char *name,
+                              char *buffer, size_t len, ResponseType rtype) {
+	lua_State *L = vis->lua;
+	if (!L)
+		return;
+	vis_lua_event_get(L, "process_response");
+	if (lua_isfunction(L, -1)) {
+		lua_pushstring(L, name);
+		if (rtype == EXIT || rtype == SIGNAL)
+			lua_pushinteger(L, len);
+		else
+			lua_pushlstring(L, buffer, len);
+		switch (rtype){
+		case STDOUT: lua_pushstring(L, "STDOUT"); break;
+		case STDERR: lua_pushstring(L, "STDERR"); break;
+		case SIGNAL: lua_pushstring(L, "SIGNAL"); break;
+		case EXIT: lua_pushstring(L, "EXIT"); break;
+		}
+		pcall(vis, L, 3, 0);
 	}
 	lua_pop(L, 1);
 }
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1372,11 +1372,11 @@ static int redraw(lua_State *L) {
 	return 0;
 }
 /***
- * Closes a stream returned by @{Vis.communicate}.
+ * Closes a stream returned by @{Vis:communicate}.
  *
  * @function close
  * @tparam io.file inputfd the stream to be closed
- * @treturn bool the same with @{io.close}
+ * @treturn bool identical to @{io.close}
  */
 static int close_subprocess(lua_State *L) {
 	luaL_Stream *file = luaL_checkudata(L, -1, "FILE*");
@@ -1388,7 +1388,9 @@ static int close_subprocess(lua_State *L) {
 	return luaL_fileresult(L, result == 0, NULL);
 }
 /***
- * Open new process and return its input handler.
+ * Open new process and return its input stream (stdin).
+ * If the stream is closed (by calling the close method or by being removed by a garbage collector)
+ * the spawned process will be killed by SIGTERM.
  * When the process will quit or will output anything to stdout or stderr,
  * the @{process_response} event will be fired.
  *
@@ -3206,8 +3208,8 @@ void vis_lua_term_csi(Vis *vis, const long *csi) {
  * The response received from the process started via @{Vis:communicate}.
  * @function process_response
  * @tparam string name the name of process given to @{Vis:communicate}
- * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
  * @tparam string|int buffer the available content sent by process; it becomes the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
+ * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
  */
 void vis_lua_process_response(Vis *vis, const char *name,
                               char *buffer, size_t len, ResponseType rtype) {
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -3213,15 +3213,19 @@ void vis_lua_term_csi(Vis *vis, const long *csi) {
 void vis_lua_process_response(Vis *vis, const char *name,
                               char *buffer, size_t len, ResponseType rtype) {
 	lua_State *L = vis->lua;
-	if (!L)
+	if (!L) {
 		return;
+	}
 	vis_lua_event_get(L, "process_response");
 	if (lua_isfunction(L, -1)) {
 		lua_pushstring(L, name);
-		if (rtype == EXIT || rtype == SIGNAL)
+		switch (rtype) {
+		case EXIT:
+		case SIGNAL:
 			lua_pushinteger(L, len);
-		else
-			lua_pushlstring(L, buffer, len);
+			break;
+		default: lua_pushlstring(L, buffer, len);
+		}
 		switch (rtype){
 		case STDOUT: lua_pushstring(L, "STDOUT"); break;
 		case STDERR: lua_pushstring(L, "STDERR"); break;
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ SRC = array.c \
 	vis-prompt.c \
 	vis-registers.c \
 	vis-text-objects.c \
+	vis-subprocess.c \
 	$(REGEX_SRC)
 
 ELF = vis vis-menu vis-digraph
--- a/vis-lua.h
+++ b/vis-lua.h
@@ -5,6 +5,7 @@
 #include <lua.h>
 #include <lualib.h>
 #include <lauxlib.h>
+
 #else
 typedef struct lua_State lua_State;
 typedef void* lua_CFunction;
--- a/vis-lua.h
+++ b/vis-lua.h
@@ -7,10 +7,11 @@
 #include <lauxlib.h>
 #else
 typedef struct lua_State lua_State;
+typedef void* lua_CFunction;
 #endif
 
 #include "vis.h"
-
+#include "vis-subprocess.h"
 /* add a directory to consider when loading lua files */
 bool vis_lua_path_add(Vis*, const char *path);
 /* get semicolon separated list of paths to load lua files
@@ -38,5 +39,6 @@ void vis_lua_win_close(Vis*, Win*);
 void vis_lua_win_highlight(Vis*, Win*);
 void vis_lua_win_status(Vis*, Win*);
 void vis_lua_term_csi(Vis*, const long *);
+void vis_lua_process_response(Vis *, const char *, char *, size_t, ResponseType);
 
 #endif
--- a/vis-subprocess.c
+++ b/vis-subprocess.c
@@ -1,0 +1,176 @@
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/wait.h>
+#include "vis-lua.h"
+#include "vis-subprocess.h"
+
+/* Maximum amount of data what can be read from IPC pipe per event */
+#define MAXBUFFER 1024
+
+/* Pool of information about currently running subprocesses */
+static Process *process_pool;
+
+Process *new_in_pool() {
+	/* Adds new empty process information structure to the process pool and
+	 * returns it */
+	Process *newprocess = (Process *)malloc(sizeof(Process));
+	if (!newprocess) return NULL;
+	newprocess->next = process_pool;
+	process_pool = newprocess;
+	return newprocess;
+}
+
+void destroy(Process **pointer) {
+	/* Removes the subprocess information from the pool, sets invalidator to NULL
+	 * and frees resources. */
+	Process *target = *pointer;
+	if (target->outfd != -1) close(target->outfd);
+	if (target->errfd != -1) close(target->errfd);
+	if (target->inpfd != -1) close(target->inpfd);
+	/* marking stream as closed for lua */
+	if (target->invalidator) *(target->invalidator) = NULL;
+	if (target->name) free(target->name);
+	*pointer = target->next;
+	free(target);
+}
+
+Process *vis_process_communicate(Vis *vis, const char *name,
+                                 const char *command, void **invalidator) {
+	/* Starts new subprocess by passing the `command` to the shell and
+	 * returns the subprocess information structure, containing file descriptors
+	 * of the process.
+	 * Also stores the subprocess information to the internal pool to track
+	 * its status and responses.
+	 * `name` - the string than should contain an unique name of the subprocess.
+	 * This name will be passed to the PROCESS_RESPONSE event handler
+	 * to distinguish running subprocesses.
+	 * `invalidator` - a pointer to the pointer which shows that the subprocess
+	 * is invalid when set to NULL. When subprocess dies, it is being set to NULL.
+	 * If the pointer is set to NULL by an external code, the subprocess will be
+	 * killed on the next main loop iteration. */
+	int pin[2], pout[2], perr[2];
+	pid_t pid = (pid_t)-1;
+	if (pipe(perr) == -1) goto closeerr;
+	if (pipe(pout) == -1) goto closeouterr;
+	if (pipe(pin) == -1) goto closeall;
+	pid = fork();
+	if (pid == -1)
+		vis_info_show(vis, "fork failed: %s", strerror(errno));
+	else if (pid == 0){ /* child process */
+		sigset_t sigterm_mask;
+		sigemptyset(&sigterm_mask);
+		sigaddset(&sigterm_mask, SIGTERM);
+		if (sigprocmask(SIG_UNBLOCK, &sigterm_mask, NULL) == -1) {
+			fprintf(stderr, "failed to reset signal mask");
+			exit(EXIT_FAILURE);
+		}
+		dup2(pin[0], STDIN_FILENO);
+		dup2(pout[1], STDOUT_FILENO);
+		dup2(perr[1], STDERR_FILENO);
+	}
+	else { /* main process */
+		Process *new = new_in_pool();
+		if (!new) {
+			vis_info_show(vis, "Can not create process: %s", strerror(errno));
+			goto closeall;
+		}
+		new->name = strdup(name);
+		if (!new->name) {
+			vis_info_show(vis, "Can not copy process name: %s", strerror(errno));
+			/* pop top element (which is `new`) from the pool */
+			destroy(&process_pool);
+			goto closeall;
+		}
+		new->outfd = pout[0];
+		new->errfd = perr[0];
+		new->inpfd = pin[1];
+		new->pid = pid;
+		new->invalidator = invalidator;
+		close(pin[0]);
+		close(pout[1]);
+		close(perr[1]);
+		return new;
+	}
+closeall:
+	close(pin[0]);
+	close(pin[1]);
+closeouterr:
+	close(pout[0]);
+	close(pout[1]);
+closeerr:
+	close(perr[0]);
+	close(perr[1]);
+	if (pid == 0) { /* start command in child process */
+		execlp(vis->shell, vis->shell, "-c", command, (char*)NULL);
+		fprintf(stderr, "exec failed: %s(%d)\n", strerror(errno), errno);
+		exit(1);
+	}
+	else
+		vis_info_show(vis, "process creation failed: %s", strerror(errno));
+	return NULL;
+}
+
+int vis_process_before_tick(fd_set *readfds) {
+	/* Adds file descriptors of currently running subprocesses to the `readfds`
+	 * to track their readiness and returns maximum file descriptor value
+	 * to pass it to the `pselect` call */
+	Process **pointer = &process_pool;
+	int maxfd = 0;
+	while (*pointer) {
+		Process *current = *pointer;
+		if (current->outfd != -1) {
+			FD_SET(current->outfd, readfds);
+			maxfd = maxfd < current->outfd ? current->outfd : maxfd;
+		}
+		if (current->errfd != -1) {
+			FD_SET(current->errfd, readfds);
+			maxfd = maxfd < current->errfd ? current->errfd : maxfd;
+		}
+		pointer = &current->next;
+	}
+	return maxfd;
+}
+
+void read_and_fire(Vis* vis, int fd, const char *name, ResponseType rtype) {
+	/* Reads data from the given subprocess file descriptor `fd` and fires
+	 * the PROCESS_RESPONSE event in Lua with given subprocess `name`,
+	 * `rtype` and the read data as arguments. */
+	static char buffer[MAXBUFFER];
+	size_t obtained = read(fd, &buffer, MAXBUFFER-1);
+	if (obtained > 0)
+		vis_lua_process_response(vis, name, buffer, obtained, rtype);
+}
+
+void vis_process_tick(Vis *vis, fd_set *readfds) {
+	/* Checks if `readfds` contains file discriptors of subprocesses from
+	 * the pool. If so, reads the data from them and fires corresponding events.
+	 * Also checks if subprocesses from pool is dead or need to be killed then
+	 * raises event or kills it if necessary. */
+	Process **pointer = &process_pool;
+	while (*pointer) {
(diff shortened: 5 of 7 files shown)


xomachine:
I am not quite familiar with vim channels, but at quick glance it looks like there are some similarities indeed.

All data is being sent to a process asyncroniously without waiting for response like it is in vim, although instead of passing a callback for handling the process answers my solution utilizes the event system already implemented in vis.

lua/vis.lua
```diff
@@ -152,6 +152,7 @@ local events = {
 	WIN_OPEN = "Event::WIN_OPEN", -- see @{win_open}
 	WIN_STATUS = "Event::WIN_STATUS", -- see @{win_status}
 	TERM_CSI = "Event::TERM_CSI", -- see @{term_csi}
+  PROCESS_RESPONCE = "Event::PROCESS_RESPONCE", -- see @{process_responce}
```
rokf, line 155: Is responce perhaps a typo of the word response?

xomachine: It is. Thank you for noticing!

mcepl:
Would it be possible to rebase this patch on the top of the current `master`, please? This is making testing rather difficult.

xomachine:
> Would it be possible to rebase this patch on the top of the current `master`, please? This is making testing rather difficult.

Done. Hopefully I did not break anything.

mcepl:
@xomachine see the linked issue report: this patch is suspected to be the cause of the crash.

euclaise:
Any updates on this? My understanding is that this is needed to implement LSP, and losing LSP without manually patching is a big point for many that may prevent switching to vis (e.g. it is the sole reason why I haven't switched from nvim).

mcepl (approved):
Merged into https://git.sr.ht/~mcepl/vis/tree and after more than half of year of testing, it seems to be working.

mcepl:
Hmm, when I pulled in also new tests, I got this one test failing on the aarch64 architecture (not on Intels, there it works just fine):

```
[  133s] make: Leaving directory '/home/abuild/rpmbuild/BUILD/vis/test/core'
[  133s] + /usr/bin/make -O -j4 V=1 VERBOSE=1 -C test/lua
[  134s] make: Entering directory '/home/abuild/rpmbuild/BUILD/vis/test/lua'
[  134s] vis v0.8-git +curses +lua +tre +acl +selinux
[  134s] cursor                        OK
[  134s] file-empty                    OK
[  134s] lines                         OK
[  134s] map-basic                     OK
[  134s] subprocess                    ./test.sh: line 34:  2489 Killed                  $VIS "$t.in" < infifo 2> /dev/null > "$t.busted"
[  134s] FAIL
[  134s] +++
[  134s] 3 successes / 0 failures / 0 errors / 0 pending : 0.021478 seconds
[  134s] The following events did not happened for process	inputtest
[  134s] 1	: 	STDERR	 - 	testdata
[  134s] 
[  134s] 1: 	STDERR	 - (string)	testdata
[  134s] 
[  134s] 2	: 	EXIT	 - 	0
[  134s] 2: 	EXIT	 - (number)	0
[  134s] Tests ok 4/5
```

[Complete build log](https://github.com/martanne/vis/files/11548067/_log.txt)

xomachine:
That's weird. My only guess is different behavior of the `read` command on aarch64. Vis was able to get the EXIT event in previous test, but did not get it in the last one, where the `read` command was involved. Therefore the child process did not finish and still waiting for the input at stdin. So that could be the reason why none of events fired.

mcepl:
I am thinking how to persuade people in charge here (@martanne , @rnpnr , @ninewise ) to review this pull request, and I have to admit that when I tried myself, I don’t find the task easy. Would it be possible, please, to reorganize the patches to be topical ones ([something like this](https://www.kernel.org/doc/html/latest/process/submitting-patches.html#split-changes) or [like this](https://who-t.blogspot.com/2009/12/on-commit-messages.html)), i.e., each change should contain all changes in one area? Commits like “Fixes of …” or “Correct implementation of …” just make reviewer’s life more complicated and less willing to help.

rnpnr:
That would definitely be appreciated. There seems to be a lot of commits that should be squashed together at a minimum.

I do want to get around to reviewing this one since I think it has the potential to be very useful but I've been trying to work through some less complicated things first.

xomachine:
I guess the best thing I can do with this history is squashing everything into one commit and rewriting the commit message. There are no logical changes in this patch set, it is just a trial-and-error story of implementing the feature.

mcepl:
Well, if I do `git diff origin/master 91bb1dd` I get

```diff
diff --git a/Makefile b/Makefile
index c862643..4c473bf 100644
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ SRC = array.c \
 	vis-prompt.c \
 	vis-registers.c \
 	vis-text-objects.c \
+	vis-subprocess.c \
 	$(REGEX_SRC)
 
 ELF = vis vis-menu vis-digraph
diff --git a/lua/vis.lua b/lua/vis.lua
index 39649c1..f2f9421 100644
--- a/lua/vis.lua
+++ b/lua/vis.lua
@@ -152,6 +152,7 @@ local events = {
 	WIN_OPEN = "Event::WIN_OPEN", -- see @{win_open}
 	WIN_STATUS = "Event::WIN_STATUS", -- see @{win_status}
 	TERM_CSI = "Event::TERM_CSI", -- see @{term_csi}
+	PROCESS_RESPONSE = "Event::PROCESS_RESPONSE", -- see @{process_response}
 }
 
 events.file_close = function(...) events.emit(events.FILE_CLOSE, ...) end
@@ -167,6 +168,7 @@ events.win_highlight = function(...) events.emit(events.WIN_HIGHLIGHT, ...) end
 events.win_open = function(...) events.emit(events.WIN_OPEN, ...) end
 events.win_status = function(...) events.emit(events.WIN_STATUS, ...) end
 events.term_csi = function(...) events.emit(events.TERM_CSI, ...) end
+events.process_response = function(...) events.emit(events.PROCESS_RESPONSE, ...) end
 
 local handlers = {}
 
diff --git a/vis-lua.c b/vis-lua.c
index 9bf5629..1bfeabb 100644
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -23,6 +23,7 @@
 
 #include "vis-lua.h"
 #include "vis-core.h"
+#include "vis-subprocess.h"
 #include "text-motions.h"
 #include "util.h"
 
@@ -52,6 +53,13 @@
 #define debug(...) do { } while (0)
 #endif
 
+typedef struct {
+	/* Lua stream structure for the process input stream */
+	FILE *f;
+	lua_CFunction closef;
+	Process *handler;
+} ProcessStream;
+
 static void window_status_update(Vis *vis, Win *win) {
 	char left_parts[4][255] = { "", "", "", "" };
 	char right_parts[4][32] = { "", "", "", "" };
@@ -162,6 +170,9 @@ void vis_lua_win_close(Vis *vis, Win *win) { }
 void vis_lua_win_highlight(Vis *vis, Win *win) { }
 void vis_lua_win_status(Vis *vis, Win *win) { window_status_update(vis, win); }
 void vis_lua_term_csi(Vis *vis, const long *csi) { }
+void vis_lua_process_response(Vis *vis, const char *name,
+                              char *buffer, size_t len, ResponseType rtype) { }
+
 
 #else
 
@@ -1367,6 +1378,47 @@ static int redraw(lua_State *L) {
 	vis_redraw(vis);
 	return 0;
 }
+/***
+ * Closes a stream returned by @{Vis.communicate}.
+ *
+ * @function close
+ * @tparam io.file inputfd the stream to be closed
+ * @treturn bool the same with @{io.close}
+ */
+static int close_subprocess(lua_State *L) {
+	luaL_Stream *file = luaL_checkudata(L, -1, "FILE*");
+	int result = fclose(file->f);
+	if (result == 0) {
+		file->f = NULL;
+		file->closef = NULL;
+	}
+	return luaL_fileresult(L, result == 0, NULL);
+}
+/***
+ * Open new process and return its input handler.
+ * When the process will quit or will output anything to stdout or stderr,
+ * the @{process_response} event will be fired.
+ *
+ * The editor core won't be blocked while the external process is running.
+ *
+ * @function communicate
+ * @tparam string name the name of subprocess (to distinguish processes in the @{process_response} event)
+ * @tparam string command the command to execute
+ * @return the file handle to write data to the process, in case of error the return values are equivalent to @{io.open} error values.
+ */
+static int communicate_func(lua_State *L) {
+	Vis *vis = obj_ref_check(L, 1, "vis");
+	const char *name = luaL_checkstring(L, 2);
+	const char *cmd = luaL_checkstring(L, 3);
+	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
+	luaL_setmetatable(L, LUA_FILEHANDLE);
+	inputfd->handler = vis_process_communicate(vis, name, cmd, (void **)(&(inputfd->closef)));
+	if (inputfd->handler) {
+		inputfd->f = fdopen(inputfd->handler->inpfd, "w");
+		inputfd->closef = &close_subprocess;
+	}
+	return inputfd->f ? 1 : luaL_fileresult(L, inputfd->f != NULL, name);
+}
 /***
  * Currently active window.
  * @tfield Window win
@@ -1524,6 +1576,7 @@ static const struct luaL_Reg vis_lua[] = {
 	{ "exit", exit_func },
 	{ "pipe", pipe_func },
 	{ "redraw", redraw },
+	{ "communicate", communicate_func },
 	{ "__index", vis_index },
 	{ "__newindex", vis_newindex },
 	{ NULL, NULL },
@@ -3148,5 +3201,34 @@ void vis_lua_term_csi(Vis *vis, const long *csi) {
 	}
 	lua_pop(L, 1);
 }
+/***
+ * The response received from the process started via @{Vis:communicate}.
+ * @function process_response
+ * @tparam string name the name of process given to @{Vis:communicate}
+ * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
+ * @tparam string|int buffer the available content sent by process; it becomes the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
+ */
+void vis_lua_process_response(Vis *vis, const char *name,
+                              char *buffer, size_t len, ResponseType rtype) {
+	lua_State *L = vis->lua;
+	if (!L)
+		return;
+	vis_lua_event_get(L, "process_response");
+	if (lua_isfunction(L, -1)) {
+		lua_pushstring(L, name);
+		if (rtype == EXIT || rtype == SIGNAL)
+			lua_pushinteger(L, len);
+		else
+			lua_pushlstring(L, buffer, len);
+		switch (rtype){
+		case STDOUT: lua_pushstring(L, "STDOUT"); break;
+		case STDERR: lua_pushstring(L, "STDERR"); break;
+		case SIGNAL: lua_pushstring(L, "SIGNAL"); break;
+		case EXIT: lua_pushstring(L, "EXIT"); break;
+		}
+		pcall(vis, L, 3, 0);
+	}
+	lua_pop(L, 1);
+}
 
 #endif
diff --git a/vis-lua.h b/vis-lua.h
index da64233..914f590 100644
--- a/vis-lua.h
+++ b/vis-lua.h
@@ -7,10 +7,11 @@
 #include <lauxlib.h>
 #else
 typedef struct lua_State lua_State;
+typedef void* lua_CFunction;
 #endif
 
 #include "vis.h"
-
+#include "vis-subprocess.h"
 /* add a directory to consider when loading lua files */
 bool vis_lua_path_add(Vis*, const char *path);
 /* get semicolon separated list of paths to load lua files
@@ -38,5 +39,6 @@ void vis_lua_win_close(Vis*, Win*);
 void vis_lua_win_highlight(Vis*, Win*);
 void vis_lua_win_status(Vis*, Win*);
 void vis_lua_term_csi(Vis*, const long *);
+void vis_lua_process_response(Vis *, const char *, char *, size_t, ResponseType);
 
 #endif
diff --git a/vis-subprocess.c b/vis-subprocess.c
new file mode 100644
index 0000000..fa8f7cd
--- /dev/null
+++ b/vis-subprocess.c
@@ -0,0 +1,176 @@
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <errno.h>
+#include <string.h>
+#include <sys/wait.h>
+#include "vis-lua.h"
+#include "vis-subprocess.h"
+
+/* Maximum amount of data what can be read from IPC pipe per event */
+#define MAXBUFFER 1024
+
+/* Pool of information about currently running subprocesses */
+static Process *process_pool;
+
+Process *new_in_pool() {
+	/* Adds new empty process information structure to the process pool and
+	 * returns it */
+	Process *newprocess = (Process *)malloc(sizeof(Process));
+	if (!newprocess) return NULL;
+	newprocess->next = process_pool;
+	process_pool = newprocess;
+	return newprocess;
+}
+
+void destroy(Process **pointer) {
+	/* Removes the subprocess information from the pool, sets invalidator to NULL
+	 * and frees resources. */
+	Process *target = *pointer;
+	if (target->outfd != -1) close(target->outfd);
+	if (target->errfd != -1) close(target->errfd);
+	if (target->inpfd != -1) close(target->inpfd);
+	/* marking stream as closed for lua */
+	if (target->invalidator) *(target->invalidator) = NULL;
+	if (target->name) free(target->name);
+	*pointer = target->next;
+	free(target);
+}
+
+Process *vis_process_communicate(Vis *vis, const char *name,
+                                 const char *command, void **invalidator) {
+	/* Starts new subprocess by passing the `command` to the shell and
+	 * returns the subprocess information structure, containing file descriptors
+	 * of the process.
+	 * Also stores the subprocess information to the internal pool to track
+	 * its status and responses.
+	 * `name` - the string than should contain an unique name of the subprocess.
+	 * This name will be passed to the PROCESS_RESPONSE event handler
+	 * to distinguish running subprocesses.
+	 * `invalidator` - a pointer to the pointer which shows that the subprocess
+	 * is invalid when set to NULL. When subprocess dies, it is being set to NULL.
+	 * If the pointer is set to NULL by an external code, the subprocess will be
+	 * killed on the next main loop iteration. */
+	int pin[2], pout[2], perr[2];
+	pid_t pid = (pid_t)-1;
+	if (pipe(perr) == -1) goto closeerr;
+	if (pipe(pout) == -1) goto closeouterr;
+	if (pipe(pin) == -1) goto closeall;
+	pid = fork();
+	if (pid == -1)
+		vis_info_show(vis, "fork failed: %s", strerror(errno));
+	else if (pid == 0){ /* child process */
+		sigset_t sigterm_mask;
+		sigemptyset(&sigterm_mask);
+		sigaddset(&sigterm_mask, SIGTERM);
+		if (sigprocmask(SIG_UNBLOCK, &sigterm_mask, NULL) == -1) {
+			fprintf(stderr, "failed to reset signal mask");
+			exit(EXIT_FAILURE);
+		}
+		dup2(pin[0], STDIN_FILENO);
+		dup2(pout[1], STDOUT_FILENO);
+		dup2(perr[1], STDERR_FILENO);
+	}
+	else { /* main process */
+		Process *new = new_in_pool();
+		if (!new) {
+			vis_info_show(vis, "Can not create process: %s", strerror(errno));
+			goto closeall;
+		}
+		new->name = strdup(name);
+		if (!new->name) {
+			vis_info_show(vis, "Can not copy process name: %s", strerror(errno));
+			/* pop top element (which is `new`) from the pool */
+			destroy(&process_pool);
+			goto closeall;
+		}
+		new->outfd = pout[0];
+		new->errfd = perr[0];
+		new->inpfd = pin[1];
+		new->pid = pid;
+		new->invalidator = invalidator;
+		close(pin[0]);
+		close(pout[1]);
+		close(perr[1]);
+		return new;
+	}
+closeall:
+	close(pin[0]);
+	close(pin[1]);
+closeouterr:
+	close(pout[0]);
+	close(pout[1]);
+closeerr:
+	close(perr[0]);
+	close(perr[1]);
+	if (pid == 0) { /* start command in child process */
+		execlp(vis->shell, vis->shell, "-c", command, (char*)NULL);
+		fprintf(stderr, "exec failed: %s(%d)\n", strerror(errno), errno);
+		exit(1);
+	}
+	else
+		vis_info_show(vis, "process creation failed: %s", strerror(errno));
+	return NULL;
+}
+
+int vis_process_before_tick(fd_set *readfds) {
+	/* Adds file descriptors of currently running subprocesses to the `readfds`
+	 * to track their readiness and returns maximum file descriptor value
+	 * to pass it to the `pselect` call */
+	Process **pointer = &process_pool;
+	int maxfd = 0;
+	while (*pointer) {
+		Process *current = *pointer;
+		if (current->outfd != -1) {
+			FD_SET(current->outfd, readfds);
+			maxfd = maxfd < current->outfd ? current->outfd : maxfd;
+		}
+		if (current->errfd != -1) {
+			FD_SET(current->errfd, readfds);
+			maxfd = maxfd < current->errfd ? current->errfd : maxfd;
+		}
+		pointer = &current->next;
+	}
+	return maxfd;
+}
+
+void read_and_fire(Vis* vis, int fd, const char *name, ResponseType rtype) {
+	/* Reads data from the given subprocess file descriptor `fd` and fires
+	 * the PROCESS_RESPONSE event in Lua with given subprocess `name`,
+	 * `rtype` and the read data as arguments. */
+	static char buffer[MAXBUFFER];
+	size_t obtained = read(fd, &buffer, MAXBUFFER-1);
+	if (obtained > 0)
+		vis_lua_process_response(vis, name, buffer, obtained, rtype);
+}
+
+void vis_process_tick(Vis *vis, fd_set *readfds) {
+	/* Checks if `readfds` contains file discriptors of subprocesses from
+	 * the pool. If so, reads the data from them and fires corresponding events.
+	 * Also checks if subprocesses from pool is dead or need to be killed then
+	 * raises event or kills it if necessary. */
+	Process **pointer = &process_pool;
+	while (*pointer) {
+		Process *current = *pointer;
+		if (current->outfd != -1 && FD_ISSET(current->outfd, readfds))
+			read_and_fire(vis, current->outfd, current->name, STDOUT);
+		if (current->errfd != -1 && FD_ISSET(current->errfd, readfds))
+			read_and_fire(vis, current->errfd, current->name, STDERR);
+		int status;
+		pid_t wpid = waitpid(current->pid, &status, WNOHANG);
+		if (wpid == -1)	vis_message_show(vis, strerror(errno));
+		else if (wpid == current->pid) goto just_destroy;
+		else if(!*(current->invalidator)) goto kill_and_destroy;
+		pointer = &current->next;
+		continue;
+kill_and_destroy:
+		kill(current->pid, SIGTERM);
+		waitpid(current->pid, &status, 0);
+just_destroy:
+		if (WIFSIGNALED(status))
+			vis_lua_process_response(vis, current->name, NULL, WTERMSIG(status), SIGNAL);
+		else
+			vis_lua_process_response(vis, current->name, NULL, WEXITSTATUS(status), EXIT);
+		destroy(pointer);
+	}
+}
diff --git a/vis-subprocess.h b/vis-subprocess.h
new file mode 100644
index 0000000..ae25e21
--- /dev/null
+++ b/vis-subprocess.h
@@ -0,0 +1,23 @@
+#ifndef VIS_SUBPROCESS_H
+#define VIS_SUBPROCESS_H
+#include "vis-core.h"
+#include <sys/select.h>
+
+struct Process {
+	char *name;
+	int outfd;
+	int errfd;
+	int inpfd;
+	pid_t pid;
+	void **invalidator;
+	struct Process *next;
+};
+
+typedef struct Process Process;
+typedef enum { STDOUT, STDERR, SIGNAL, EXIT } ResponseType;
+
+Process *vis_process_communicate(Vis *, const char *command, const char *name,
+                                 void **invalidator);
+int vis_process_before_tick(fd_set *);
+void vis_process_tick(Vis *, fd_set *);
+#endif
diff --git a/vis.c b/vis.c
index f21efa8..24e7f65 100644
--- a/vis.c
+++ b/vis.c
@@ -28,6 +28,7 @@
 #include "vis-core.h"
 #include "sam.h"
 #include "ui.h"
+#include "vis-subprocess.h"
 
 
 static void macro_replay(Vis *vis, const Macro *macro);
@@ -1429,7 +1430,8 @@ int vis_run(Vis *vis) {
 
 		vis_update(vis);
 		idle.tv_sec = vis->mode->idle_timeout;
-		int r = pselect(1, &fds, NULL, NULL, timeout, &emptyset);
+		int r = pselect(vis_process_before_tick(&fds) + 1, &fds, NULL, NULL,
+		                timeout, &emptyset);
 		if (r == -1 && errno == EINTR)
 			continue;
 
@@ -1437,6 +1439,7 @@ int vis_run(Vis *vis) {
 			/* TODO save all pending changes to a ~suffixed file */
 			vis_die(vis, "Error in mainloop: %s\n", strerror(errno));
 		}
+		vis_process_tick(vis, &fds);
 
 		if (!FD_ISSET(STDIN_FILENO, &fds)) {
 			if (vis->mode->idle)
```

That’s 424 line, so I hoped whether you can get some organization to this.

mcepl:
I am still not a real C programmer, so just one (probably nonsensical) comment:

* shouldn't that definition of `ProcessStream` be in the header file?

xomachine:
> That’s 424 line, so I hoped whether you can get some organization to this.

Maybe I missed the point, but according to the links you provided, the patch should implement one feature - no more, no less. So there is no point in splitting the patch to multiple parts that will lead to inability to compile the editor on some intermediate commit or to the dead code.
And I did not rewrite the whole editor. My patch mostly adds new code, so reviewing those changes should not take that much effort as it would take in case of replacing huge amounts of existing code.

On other hand I can try to implement partially working feature. E. g. in the first commit only ability to run a process and send data to its stdin, the second one - event handling, the third - ability to stop a process by closing its handle and so on... But I doubt that it will be more readable and is a good practice overall.


>     * shouldn't that definition of `ProcessStream` be in the header file?

I always try to keep headers as small as possible because their content is included to many compilation units therefore increases build time. So if the definition is not needed outside one compilation unit, I put it into a source file. It is especially important in C++ with its monstrous templates and long compile times, and I thought that it also a good practice in C.
If that is a problem - i will move the `ProcessStream` struct into the header.

rnpnr:
I agree. I don't see anyway of splitting this up into subcommits.

I also don't think `ProcessStream` should be in the header. I would argue that since its only used in a single place it shouldn't even be type defined. I think its better to just leave it as a named struct that is local to `communicate_func()`.

Is there a minimal example of some lua code that makes use the exposed features?

xomachine:
> Is there a minimal example of some lua code that makes use the exposed features?

One example can be found in the test code (https://github.com/martanne/vis-test/pull/12/files#diff-8e2c47bd7e90844a45b8b2e5c9c6995fc3431086494baf17617c924dca8abf8dR67)
Another is in my implementation of the nimsuggest plugin (which actually is not maintained for a long time)
https://github.com/xomachine/nis/blob/async/sessions.lua#L85

tosch42:
There are some smaller things that should be changed too. For example, it's unnecessary to cast the result of a malloc(3) call. Empty parameter lists should be `func(void)` instead of `func()`. All occurrences of `MAXBUFFER` should be replaced by `PIPE_BUF`. `typedef` the `Process` structure *before* its declaration, that way you can just use `Process` inside of the struct. All the local functions should be declared `static`. The naming scheme of those functions is a bit inconsistent to me. I'd rather call `new_in_pool()` `new_process()` and instead of `destroy()` I'd say `destroy_process()`. You get the idea. Then, if you have to write a comment that explains a function, please place it above the function, *not* into it. If it was up to me, I'd toss most of the comments. They don't seem necessary to me. That's just the little things that come to mind, unfortunately I currently don't have the time to review this in any more depth.

mcepl:
> Maybe I missed the point, but according to the links you provided, the patch should implement one feature - no more, no less.

Just for the record that’s not exactly what those links meant (AFAIK): the main point is that instead of the chronological commits (“Let’s try this …”, “Whoops, that didn’t work …”, “It’s Friday, saving working before the weekend” etc.) the branch for review should be organized topically (“necessary changes in C making the ground for everything else”, “implementation of Lua interface”, “documentation”, “tests” or something like it).

And yes, 400+ lines patch is probably somewhere on the edge of being small enough not to split it at all.

xomachine:
> There are some smaller things that should be changed too.

I tried to fix the problems you mentioned. At least some of them.

Each commit addresses one issue found in code. Or should it be one commit per review iteration? Or I should fixup my previous commit to make the history cleaner? Now I think I completely lost the purpose of the commit history. =')

rnpnr (changes requested):
Hi, I have tried my best to conduct a thorough review. First and foremost
I want to make sure that the interface makes sense. Then I checked for
any obvious bugs. If we get the interface right to begin with it will
be much easier to deal with any hidden bugs later. Overall I think
everything looks pretty good. I do have some discussion points though:

First: Is everyone OK with this interface (call `vis:communicate()` with
a **shell command** -> have an event fired to handle the response)? A
potential alternative would be to call `vis:communicate()` with
a lua function as the argument. The lua function could then call
other lua functions such as `vis:pipe()` and an event would still be
fired on completion. This would be more similar to the way [mpv](https://github.com/mpv-player/mpv/) handles
subprocesses. Potentially this alternative could be added as an extension
later though.

If we are OK with the current interface should the `process_response()`
handler have a separate parameter for the exit code number? To me
it seems a little clunky to have a parameter that is either a string
or an integer. With that case though the `buffer` parameter could be
`nil`. Thoughts?

The low level comments on the code are inline below. Some additional
comments:

* places where `while()` is used should be replaced by `for()`. The only
  one where it is maybe OK is in `vis_process_tick()` but even that one
  can be expressed as `for()`.
* `if-else` chains with braces in one case should have them on every
  case and the brace should be on the same line (` } else {`). I believe
  this way will be consistent with the rest of vis.
* In a similar vein please avoid `if` statements with the body on the
  same line.

I will conduct some more tests and see if I can find any issues.

rnpnr:
Maybe I'm doing something wrong but overall it seems this doesn't behave
as intended. My example test:

```lua
vis.events.subscribe(vis.events.INIT, function()
	vis:map(vis.modes.NORMAL, " t", function()
		vis:communicate("timer", "sleep 10; echo timer done")
	end, "start a 10s timer")
end)
vis.events.subscribe(vis.events.PROCESS_RESPONSE, function(name, resp, buf)
	vis:message("PROCESS_RESPONSE\n\tname = " .. name)
	vis:message("\tresp_type = " .. resp)
	vis:message("\tbuffer = " .. buf)
end)
```
After I press `␣t` I can move the cursor for a little bit, then I can't
move the cursor for some seconds, and finally I get this printed to the
message window:
```
PROCESS_RESPONSE
	name = timer
	resp_type = 143
	buffer = EXIT
```

vis-lua.c
```diff
@@ -1367,6 +1371,55 @@ static int redraw(lua_State *L) {
 	vis_redraw(vis);
 	return 0;
 }
+/***
+ * Closes a stream returned by @{Vis.communicate}.
+ *
+ * @function close
```
rnpnr, line 1377: This function needs to be added to `vis_lua[]`.

fischerling:
> Maybe I'm doing something wrong but overall it seems this doesn't behave as intended. My example test:
> 
> ```lua
> vis.events.subscribe(vis.events.INIT, function()
> 	vis:map(vis.modes.NORMAL, " t", function()
> 		vis:communicate("timer", "sleep 10; echo timer done")
> 	end, "start a 10s timer")
> end)
> vis.events.subscribe(vis.events.PROCESS_RESPONSE, function(name, resp, buf)
> 	vis:message("PROCESS_RESPONSE\n\tname = " .. name)
> 	vis:message("\tresp_type = " .. resp)
> 	vis:message("\tbuffer = " .. buf)
> end)
> ```
> 
> After I press `␣t` I can move the cursor for a little bit, then I can't move the cursor for some seconds, and finally I get this printed to the message window:
> 
> ```
> PROCESS_RESPONSE
> 	name = timer
> 	resp_type = 143
> 	buffer = EXIT
> ```

For me your test case works with my own `visrc.lua` file and bunch of plugins and configured events without problems. I can continue using vis normally
during the ten seconds and afterwards I get the expected output:

	PROCESS_RESPONSE
	    name = timer
	    resp_type = timer done

	    buffer = STDOUT
	PROCESS_RESPONSE
	    name = timer
	    resp_type = 0
	    buffer = EXIT

My login shell is fish.

Sometimes when trying this simple test case with a minimal vis config the command was terminated by SIGTERM, but I can not reliably reproduce it.

xomachine:
Thank you for review! Though I can not guarantee that I will answer or fix everything soon, I will definitely look into it.

> Sometimes when trying this simple test case with a minimal vis config the command was terminated by SIGTERM, but I can not reliably reproduce it.

This happens because the lua garbage collector collects the process handle and therefore vis kills spawned process. To prevent it you should put the result of the `vis:communicate()` to a variable and keep it in scope until the process is finished.

mcepl:
> Maybe I'm doing something wrong but overall it seems this doesn't behave as intended. My example test:
> 
> ```lua
> vis.events.subscribe(vis.events.INIT, function()
> 	vis:map(vis.modes.NORMAL, " t", function()
> 		vis:communicate("timer", "sleep 10; echo timer done")
> 	end, "start a 10s timer")
> end)
> vis.events.subscribe(vis.events.PROCESS_RESPONSE, function(name, resp, buf)
> 	vis:message("PROCESS_RESPONSE\n\tname = " .. name)
> 	vis:message("\tresp_type = " .. resp)
> 	vis:message("\tbuffer = " .. buf)
> end)
> ```
> 
> After I press `␣t` I can move the cursor for a little bit, then I can't move the cursor for some seconds, and finally I get this printed to the message window:
> 
> ```
> PROCESS_RESPONSE
> 	name = timer
> 	resp_type = 143
> 	buffer = EXIT
> ```

Works for me just fine (as applied in https://git.sr.ht/~mcepl/vis), I had to just switch off [vis-lspc](https://gitlab.com/muhq/vis-lspc), because it got confused.

rnpnr:
> ```
>     resp_type = timer done
> 
>     buffer = STDOUT
> ```

But even this isn't correct. `resp_type` should be `STDOUT` and `buffer`
should be `timer done\n`.

> This happens because the lua garbage collector collects the process handle and therefore vis kills spawned process

This seems to be the issue with my example. It works fine if I have a
global variable holding the handle. A note about this should definitely
be added to the documentation. Interestingly in only happens when I
actually try to do something. If I just start the timer and do nothing
the process finishes successfully. Maybe that is expected though.

Is there a way we can avoid this altogether? It seems like an annoying
limitation.

xomachine:
> But even this isn't correct. resp_type should be STDOUT and buffer
> should be timer done\n.

The order of input parameters is wrong in the example. According to the documentation the second parameters is the answer itself, and the third one is the type. Maybe it is counterintuitive and should be changed.
 ```
 * @function process_response
 * @tparam string name the name of process given to @{Vis:communicate}
 * @tparam string response_type can be "STDOUT" or "STDERR" if new output was received in corresponding channel, "SIGNAL" if the process was terminated by a signal or "EXIT" when the process terminated normally
 * @tparam string|int buffer the available content sent by process; it becomes the exit code number if response\_type is "EXIT", or the signal number if response\_type is "SIGNAL"
 ```

> Is there a way we can avoid this altogether? It seems like an annoying
> limitation.

It is done this way intentionally. If we don't keep track of processes spawned by vis it will lead to possible process leak. E. g. some process spawned by vis hangs for some reason and does not terminate in a proper way - in that case lua code is unable to kill it and the process will be in background until the editor is terminated.

rnpnr:
> I got caught by my own trap

No worries, I probably should have called the variable `rtype` not `resp`.

> If we don't keep track of processes spawned by vis it will lead to possible process leak

That makes sense, I wasn't really thinking about all potential uses. In
any case we need to make sure that it is documented otherwise we will
get lots of bug reports from doing exactly what I did.

rnpnr:
> running `hunspell` on the background

But in that case you would want to keep the file handle around to
keep writing to it as new text is input. @pippi1otta is talking about
examples like my test where you have something that needs to be run
in the background but doesn't need data written to it and can finish
whenever it finishes. I won't block merging this over such an addition
but I think it could be an acceptable extension later on.

mcepl:
> But in that case you would want to keep the file handle around to keep writing to it as new text is input. @pippi1otta is talking about examples like my test where you have something that needs to be run in the background but doesn't need data written to it and can finish whenever it finishes. I won't block merging this over such an addition but I think it could be an acceptable extension later on.

That’s correct.

mcepl:
And we shouldn’t forget https://github.com/martanne/vis-test/pull/12.

xomachine:
For relatively short-living processes that don't take any input there could be a timer that will kill the process when the time is out. Honestly I can not imagine the long living process that we don't want to track in the plugin.
It might be possible to detach the process and prevent it from being killed by vis by using the functionality of the shell. ( `&` + `disown`)

mcepl:
@xomachine Could we expect your reaction to @rnpnr ’s review any time soon? This apparently is the most important PR blocking #1115 .

vis-lua.c
xomachine: Actually, no. It is not a function, it is a method. So it is already exposed by setting it to a `closef` member of the ProcessStream structure.

vis-lua.c
```diff
@@ -1367,6 +1371,55 @@ static int redraw(lua_State *L) {
...
+		lua_CFunction closef;
+		Process *handler;
+	} ProcessStream;
+
+	Vis *vis = obj_ref_check(L, 1, "vis");
+	const char *name = luaL_checkstring(L, 2);
+	const char *cmd = luaL_checkstring(L, 3);
+	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
+	luaL_setmetatable(L, LUA_FILEHANDLE);
+	inputfd->handler = vis_process_communicate(vis, name, cmd, (void **)(&(inputfd->closef)));
```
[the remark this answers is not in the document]
xomachine:
```
vis-lua.c: In function 'communicate_func':
vis-lua.c:1417:68: warning: passing argument 4 of 'vis_process_communicate' from incompatible pointer type [-Wincompatible-pointer-types]
 1417 |         inputfd->handler = vis_process_communicate(vis, name, cmd, &(inputfd->stream.closef));
      |                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                    |
      |                                                                    int (**)(lua_State *)
In file included from vis-lua.h:14,
                 from vis-lua.c:24:
vis-subprocess.h:21:41: note: expected 'void **' but argument is of type 'int (**)(lua_State *)'
   21 |                                  void **invalidator);
      |                                  ~~~~~~~^~~~~~~~~~~
```
The compiler sees it differently.

vis-lua.c
rnpnr: Sorry that was my bad. So should it be used like:
```
s = vis:communicate(...)
s:close()
```

vis-lua.c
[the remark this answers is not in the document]
rnpnr: Is `invalidator` supposed to be a function pointer? If it is then the
struct member and function parameter should both be function pointers
(eg `int **(*invalidator)(lua_State *)`).

vis-lua.c
```diff
@@ -1367,6 +1371,55 @@ static int redraw(lua_State *L) {
...
 }
+/***
+ * Closes a stream returned by @{Vis.communicate}.
+ *
+ * @function close
+ * @tparam io.file inputfd the stream to be closed
+ * @treturn bool the same with @{io.close}
+ */
+static int close_subprocess(lua_State *L) {
+	luaL_Stream *file = luaL_checkudata(L, -1, "FILE*");
```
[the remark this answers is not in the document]
xomachine: I did not get what are we trying to achieve here. The return value of the `Vis:communicate()`  mimics an object of the native for lua `FILE*` type. It is not a type exclusively implemented in vis, so it is not registred by `obj_new_type()` and does not have any functions set by `luaL_setfuncs()`. As far as I could understand the wrapper is needed for custom lua types implemented in the vis code. But it is not the case with the `FILE*` type.

Or maybe I get it all wrong.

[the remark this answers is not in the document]
rnpnr: `obj_ref_check()` has error checking and leaves the lua stack in a proper
state. It is currently only being used with the custom types but that
doesn't mean in can't  be used with  types like `FILE *`.

That was the only reason I suggested using it.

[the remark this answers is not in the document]
xomachine: I tried and I failed. Moreover, I could not find an example of object created without `obj_ref_new` being checked by `obj_ref_check`. So I guess that the problem here is that the `FILE*` is not actually a reference type. It is created by `lua_newuserdata` with the size of the structure (like in `obj_new`) right in the lua stack and the `FILE*` metatable is assigned to it to achieve compatibility with the native `FILE*` type. On the other hand `obj_ref_check` checks objects that are already in C heap and have a reference in Lua created with `obj_ref_new`.

So I am not sure what should I do in this case. Should I create a reference to a `FILE*` with `obj_ref_new` which probably will lead to inability to use native lua methods like `write` and `close` (they are provided by metatable)? Or there is another way to handle this kind of objects I did not found?

vis-lua.c
xomachine: That is right. The return value for `vis:communicate()` behaves exactly like any other `FILE*` object (e.g. returned by `io.open`)

vis-lua.c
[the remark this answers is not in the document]
xomachine: For `vis_process_communicate` it does not matter what kind of pointer it is. It just a canary pointer which is being set to `NULL` if the stream is closed. According to the documentation in the Lua source code the `closef` member of the [luaL_Stream](https://www.lua.org/source/5.4/lauxlib.h.html#luaL_Stream) structure becomes `NULL` when the stream is closed. So I used it to detect stream status. 

```c
typedef struct luaL_Stream {
  FILE *f;  /* stream (NULL for incompletely created streams) */
  [lua_CFunction](https://www.lua.org/source/5.4/lua.h.html#lua_CFunction) closef;  /* to close stream (NULL for closed streams) */
}
```

In C++ I'd use a closure here (capture the `fclose` member address, check it in the closure and return `true` or `false` depending on the stream status)

vis-lua.c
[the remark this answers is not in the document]
rnpnr: I missed what the `obj_ref_get()` function was doing. In that case can
you just add an error check like:

```
if (file)
	lua_pop(L, 1);
else
	luaL_argerror(L, -1, "invalid object reference");
```

xomachine:
Hopefully I answered on all review comments (by commenting or committing). @rnpnr Is there anything I missed?

rnpnr:
> Is there anything I missed?

I think you have corrected most of the issues but there are still a
couple points missing:

* I still don't like that a function pointer is being cast to a void
  pointer. I know you will need to include an extra header but I would
  still prefer that the struct member and function parameter be the
  correct type. I'm just worried that something hacky like that is going
  to lead to a difficult to find bug in the future.

* For the `PROCESS_RESPONSE` event, I think that the order of returned
  variables as documented was logical before. I know this will cause
  churn in the existing plugins but I think it is worth it to get the
  interface right. (Only the order of the variables in the documentation
  was changed)

* Additionally I think the exit code should always be returned as a
  separate parameter. Since you will already have to create churn I
  think the following makes the most sense:
  ```
  vis.events.subscribe(vis.events.PROCESS_RESPONSE, function(name, resp_type, exit_code, data)
  ```
  Note that doing this means you can't repurpose the `len` variable in
  `vis_lua_process_response()` but I think that should have been fixed
  anyways.

If anyone disagrees with me feel free to speak up. This is just what
seems best to me but I could be wrong.

rnpnr:
Also let me just reiterate: 

The most important thing about this patch for me is the interface. As long
as we get the interface correct we can fix any underlying issues later.

vis-lua.c
```diff
@@ -1367,6 +1371,61 @@ static int redraw(lua_State *L) {
...
+ * Closes a stream returned by @{Vis:communicate}.
+ *
+ * @function close
+ * @tparam io.file inputfd the stream to be closed
+ * @treturn bool identical to @{io.close}
+ */
+static int close_subprocess(lua_State *L) {
+	luaL_Stream *file = luaL_checkudata(L, -1, "FILE*");
+	if (file) {
+		lua_pop(L, 1);
```
rnpnr, line 1384: I'm sorry for  making you do this again but the `lua_pop()` is actually
incorrect. You just need:

```
if (!file)
	return luaL_argerror(...)
```

After reading `obj_ref_get()` in more detail I noticed it was modifying
the lua stack. This wasn't documented where it should be. It won't
matter to you but I am going to either add a comment or change it to
not do that so that I don't fall into that trap again.

xomachine: I looked to the Lua source code and now I do not even sure that remaining condition is necessary. The `luaL_checkudata` function never returns `NULL` according to [its source code](https://github.com/lua/lua/blob/v5.4.0/lauxlib.c#L345). 

It calls `luaL_testudata` which does return `NULL`, but after that it calls [a macro](https://github.com/lua/lua/blob/v5.4.0/lauxlib.h#L135) `luaL_argexpected` which will call the `luaL_typeerror` [function](https://github.com/lua/lua/blob/v5.4.0/lauxlib.c#L193) and eventually `luaL_argerror` with proper error message when the return value is `NULL`.

xomachine:
If we are going to break the interface anyway, let's at least make it more convenient. So I would like to get another proposals for better interface. One of them was about performing reading the spawned process notifications via callback functions. Would it be better to rewrite my patch to implement this interface? Or may be there are another ideas?

vis-lua.c
rnpnr: Yep you are right. Sorry for the wild goose chase. I'm also pretty new
to the Lua-C API.

I do wish the documentation was a little more specific. The only indication
of this is [here][lua]:

> ... luaL_check* or luaL_opt*. All of these functions throw an error if the check is not satisfied.

Which doesn't really tell you that the function doesn't return in that case.

[lua]: https://www.lua.org/manual/5.1/manual.html#4

rnpnr:
I guess it depends on what you mean by callback functions. I'm all
for removing the need for the `name` parameter and replacing it with a
`function to call on data/finish` parameter but if it's a significant
effort to implement I wouldn't worry about it.

mcepl:
> First: Is everyone OK with this interface (call `vis:communicate()` with
> a **shell command** -> have an event fired to handle the response)? A
> potential alternative would be to call `vis:communicate()` with
> a lua function as the argument. The lua function could then call
> other lua functions such as `vis:pipe()` and an event would still be
> fired on completion. This would be more similar to the way [mpv](https://github.com/mpv-player/mpv/) handles
> subprocesses. Potentially this alternative could be added as an extension
> later though.

Actually, if we are changing API of `vis:communicate()` could we return to this question again? That was the one thing I hated about Neovim’s `jobsend()` (now `chansend()`), that you had to run a shell command in order to have a true subprocess. Could I have a subprocess out of pure Lua function?

rnpnr:
> Could I have a subprocess out of pure Lua function?

I think this would be much more powerful but it is also a little more
complicated to implement than a simple `fork()` and keeping track of
some input and output pipes.

My original idea wasn't really fleshed out but I think one possibility
could be:

```
vis:communicate(background_fn, handler_fn)
```

`background_fn` would run in the background until it returns some
data. `handler_fn` would run in the main process and would take as its
arguments the returned data from `background_fn`. I'm just not sure
how possible this is from the Lua-C interface since `background_fn`
can't share the same Lua stack as the main process.

@xomachine what are your thoughts? I think what you have already wrote
is pretty good and will already allow for some more advanced plugins.

xomachine:
> Could I have a subprocess out of pure Lua function?

It could be achieved in multiple ways but all of them have disadvantages.

1. By forking. In this case it is actually equal to calling Lua interpreter because there will be no easy way to use any vis specific functions.
2. By creating a thread. The vis-specific API can be available, but it will require a lot of hassle with synchronization between main thread and subprocesses. Every call should be wrapped with the critical section.
3. By using Lua coroutines. Does not require synchronization, but requires wrapping every blocking IO call with code that registers file descriptors in the `pselect` call and suspends the coroutine.

In my opinion the first way does not make any significant improvement in interface due to inability to call vis API. The second one is quite dangerous and error prone as any multithreading. So the third way looks reliable, but requires full rewrite of the patch (basically writing new patch) and I can not tell when it will be ready.

Or we can concentrate on improvement of the process spawning and add callbacks instead of events to get rid of the process naming.

mcepl:
> Or we can concentrate on improvement of the process spawning and add callbacks instead of events to get rid of the process naming.

I agree, let’s shelve Lua subprocess for now, and return to the shell-only stuff.

xomachine:
After pondering about the way to use callback functions instead of events I came to conclusion that it will be a bit tricky. As far as I understand, there is no way to save a Lua function in the C code and use it later. All operations with Lua functions should be performed without leaving the Lua stack. So to achieve the interface with callbacks I need to use some kind of registry in a global variable in the Lua stack and match every element against the C structure representing the process.

Are we ok with this approach?

I am starting to think that implementing async interface based on coroutines might be an easier task. At least it is possible to save coroutine in the C code.

rnpnr:
> So to achieve the interface with callbacks I need to use some kind of registry in a global variable in the Lua stack and match every element against the C structure representing the process.

If thats the case it kinda sounds like we are reimplementing something
that can be easily done in lua. I wouldn't worry about it. I think this
interface is still quite reasonable:

```lua
local process_handlers = { ["some_name"] = { ["STDOUT"] = stdout_fn, ["STDERR"] = stderr_fn } }
vis.events.subscribe(vis.events.PROCESS_RESPONSE, function(name, resp_type, exit_code, data)
	local fn = process_handlers[name][resp_type]
	if fn ~= nil then
		fn(exit_code, data)
	end
end)
```

I would still like the points in this [comment][] covered before merging.

[comment]: https://github.com/martanne/vis/pull/675#issuecomment-1668861300

xomachine:
I've changed the interface of the `PROCESS_RESPONSE` event to `(process_name, event_type, code, buffer)`.
And yet I can not force myself to include Lua header to `vis-subprocess.h` nor find another solution to avoid casting to the raw pointer `void**`.

rnpnr:
Thanks for your hard work!

> And yet I can not force myself to include Lua header to `vis-subprocess.h`

As it turns out the header is already included via `vis-core.h`. I'm
not sure I see any problem with this:

```diff
diff --git a/vis-lua.c b/vis-lua.c
index 2349061..f6388c2 100644
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1420,9 +1420,9 @@ static int communicate_func(lua_State *L) {
 	Vis *vis = obj_ref_check(L, 1, "vis");
 	const char *name = luaL_checkstring(L, 2);
 	const char *cmd = luaL_checkstring(L, 3);
-	ProcessStream *inputfd = (ProcessStream *)lua_newuserdata(L, sizeof(ProcessStream));
+	ProcessStream *inputfd = lua_newuserdata(L, sizeof(ProcessStream));
 	luaL_setmetatable(L, LUA_FILEHANDLE);
-	inputfd->handler = vis_process_communicate(vis, name, cmd, (void**)(&(inputfd->stream.closef)));
+	inputfd->handler = vis_process_communicate(vis, name, cmd, &(inputfd->stream.closef));
 	if (inputfd->handler) {
 		inputfd->stream.f = fdopen(inputfd->handler->inpfd, "w");
 		inputfd->stream.closef = &close_subprocess;
diff --git a/vis-subprocess.c b/vis-subprocess.c
index 83b36cc..1aff421 100644
--- a/vis-subprocess.c
+++ b/vis-subprocess.c
@@ -70,7 +70,7 @@ static void destroy_process(Process **pointer) {
  * killed on the next main loop iteration.
  */
 Process *vis_process_communicate(Vis *vis, const char *name,
-                                 const char *command, void **invalidator) {
+                                 const char *command, int (**invalidator)(lua_State *)) {
 	int pin[2], pout[2], perr[2];
 	pid_t pid = (pid_t)-1;
 	if (pipe(perr) == -1) {
diff --git a/vis-subprocess.h b/vis-subprocess.h
index 569b753..0fd4944 100644
--- a/vis-subprocess.h
+++ b/vis-subprocess.h
@@ -11,14 +11,14 @@ struct Process {
 	int errfd;
 	int inpfd;
 	pid_t pid;
-	void **invalidator;
+	int (**invalidator)(lua_State *);
 	Process *next;
 };

 typedef enum { STDOUT, STDERR, SIGNAL, EXIT } ResponseType;

 Process *vis_process_communicate(Vis *, const char *command, const char *name,
-                                 void **invalidator);
+                                 int (**invalidator)(lua_State *));
 int vis_process_before_tick(fd_set *);
 void vis_process_tick(Vis *, fd_set *);
 #endif
```

vis-lua.c
```diff
@@ -3219,20 +3220,23 @@ void vis_lua_process_response(Vis *vis, const char *name,
...
 		}
-		pcall(vis, L, 3, 0);
+		switch (rtype) {
+		case EXIT:
+		case SIGNAL:
+			lua_pushinteger(L, len);
+			lua_pushlstring(L, buffer, 0);
+			break;
+		default:
+			lua_pushinteger(L, 0);
```
rnpnr, line 3236: In this case `lua_pushnil(L)` is more appropriate since the program
hasn't exited yet and `0` represents a normal exit in general.

rnpnr:
These changes are looking good! I will give it another overall look over
and run through soon. I think its likely good to go.

rnpnr:
Applied in 0b01532! Thanks again for all the hard work!

There is one thing that might come up later which is that the `close()`
method doesn't actually kill the underlying process. So in my sleep
example calling `close()` will just cause vis to hang until the `sleep`
program exits. But its up for debate whether or not that is incorrect
behaviour so I didn't think it was worth blocking this over.

mcepl:
> Applied in [0b01532](https://github.com/martanne/vis/commit/0b015320382e74fcb385a46a81304f588ed27f77)! Thanks again for all the hard work!

\o/
 |
/ \

So, what remains from 0.9 (and hopefully subsequent merge of tests, see https://lists.sr.ht/~martanne/devel/%3C04ffb382-b86a-b8be-c812-8a001ed21973%40cepl.eu%3E)?

PR #676, opened 2018-03-05
Fix a typo in identifiers

Two-Finger:


Two-Finger:
I have no idea how to fix the coverage issue.

PR #678, opened 2018-03-09
Fix 'vis-single' target on FreeBSD

SDoradus:
The 'vis-single' target fails to build on FreeBSD due to differences in core utilities and
missing definitions found in 'signal.h'

This might allow this to work on MacOS and other operating systems if they have similar
issues although it would require adjusting the uname check.

The if uname check in 'Makefile' might not be appropriate if worked before on systems
other than Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,27 @@ vis-single-payload.inc: $(EXECUTABLES) lua/*
 	echo '#ifndef VIS_SINGLE_PAYLOAD_H' > $@
 	echo '#define VIS_SINGLE_PAYLOAD_H' >> $@
 	echo 'static unsigned char vis_single_payload[] = {' >> $@
-	tar --mtime='2014-07-15 01:23Z' --owner=0 --group=0 --numeric-owner --mode='a+rX-w' -c \
-		$(EXECUTABLES) $$(find lua -name '*.lua' | LC_ALL=C sort) | xz -T 1 | \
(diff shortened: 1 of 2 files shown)


martanne:
Thanks, rather than adding various system specific checks we should instead only use standard utilities whenever possible.

* for `tar` we could add a make variable, then you could override the default using something like `make vis-single TAR=gtar`.
* the generation of the C array should surely be possible using only POSIX tools/options? Does the following work?
```
	echo '#ifndef VIS_SINGLE_PAYLOAD_H' > $@
	echo '#define VIS_SINGLE_PAYLOAD_H' >> $@
	echo 'static unsigned char vis_single_payload[] = {' >> $@
	$(TAR) --mtime='2014-07-15 01:23Z' --owner=0 --group=0 --numeric-owner --mode='a+rX-w' -c \
		$(EXECUTABLES) $$(find lua -name '*.lua' | LC_ALL=C sort) | xz -T 1 | \
		od -t x1 -v | sed -e 's/^[0-9a-fA-F]\{1,\}//g' -e 's/\([0-9a-f]\{2\}\)/0x\1,/g' >> $@
	echo '};' >> $@
	echo '#endif' >> $@
```
* the `signal.h` header should be included unconditionally

martanne:
Do the latest commits work as expected on BSD?

@eworm-de can you confirm that it still works, produces reproducible builds etc. for you too?

Thanks

eworm-de:
Looks good to me. `sha256` checksum for `v0.4-83-g85919f6` is `a6075a57c0e41ab5aa3016ddb83086a9cfab13f54b6e8f9fe4325c8ea83e335b`.

Issue #342, opened 2016-06-27
folding support?


lammermann:
Could you implement folding in vis? In textadept editor this is done by the language lexers and I think you use the same lexers for vis syntax highlighting. So maybe you could use that to implement something similar in vis.

sshbio:
Here is a message where @martanne talked about folds: https://github.com/martanne/vis/issues/262#issuecomment-209937871

v217:
Is it possible to implement folding support that behaves identical to the vim zf command in vis only using the lua api?

martanne:
@v217 in its current form not, no. The current cursor placement code assumes a contiguous text range. This restriction should eventually be lifted. Proper folding support will require quite a few changes to core editor primitives.

PR #680, opened 2018-03-11
[security hotfix] buffer overflow vulnerability fix #679

### LINKED ISSUE #679: [security] Crash when parsing special files (e.g. file ACCD)
**Pull Request**: #680 

I found a crash for this program.
The issue arises due to incorrect parsing of unicode invalid and incomplete bytes in function `view_draw()` (file `view.c`) and leads to a **buffer overflow**.
Luckily, this program is compiled with hardened instructions by default. Which means it is **not exploitable** because `strcat()` is guarded by `__strcat_chk()`.

Here is the code where the overflow occurs:
``` {.C .numberLines startFrom=381}
if (cell.width == 0 && prev_cell.len + cell.len < sizeof(cell.data)) {
    ...
    strcat(prev_cell.data, cell.data);
```

Here is why the overflow occurs:
1. at some point vis reads an invalid unicode sequence and handles it semi-correctly (wrong length, see below).
2. immediatly afterwards vis reads an incomplete  unicode sequence and handles it incorrectly (buffer cursor is reset to position of invalid sequence)
3. the invalid sequence is read again but this time incorrectly accepted and its width is set to zero
4. the `strcat()` code at line 381 is triggered and `prev_cell` contains its previous value plus the invalid byte sequence.
5. go to step 1 -> infinite loop -> overflow at some point with `prev_cell.data` equal to `efbfbdXX...XX` (where `XX` is the invalid sequence and `...` represents its repetition each loop).

and the overflow is due to 3 programming mistakes:
1. data and control information are mixed together when calculating the `.len` value for an invalid unicode sequence.
   This is because invalid sequences are replaced with 3 bytes and yet the `len` is set to 1. 
   This is wrong because `len` is then used for meaningful program execution (such as overflow checks mentioned previously) and declaring a lower number of bytes than is the case may lead to an overflow.
   
   The code where the error occurs:
``` {.C .numberLines startFrom=351}
if (len == (size_t)-1 && errno == EILSEQ) {
			/* ok, we encountered an invalid multibyte sequence,
			 * replace it with the Unicode Replacement Character
			 * (FFFD) and skip until the start of the next utf8 char */
			for (len = 1; rem > len && !ISUTF8(cur[len]); len++);
cell = (Cell){ .data = "\xEF\xBF\xBD", .len = len, .width = 1 };
```
2. data and control information are mixed together when advancing buffer and file position variables.
   In short, the position pointer `pos` for the file is always one character behind the buffer pointer `cur`.

   This happens because the very first `prev_cell` is assigned `.len = 0` and after the first character is read the position for the file is advanced by `prev_cell.len`.
   The code where the error occurs:
``` {.C .numberLines startFrom=387}
pos += prev_cell.len;
prev_cell = cell;
```
3. `mbrtowc()` is not pure and no error checks are made after `wcwidth()` is called.
   After personal analysis, it appears that `mbrtowc()` does not return output solely based on its inputs.
   Here is the issue: if vis encounters an INVALID sequence and then an INCOMPLETE sequence and then the first INVALID sequence again, this leads to the INVALID sequence being accepted in vis. On my machine, I was not able to reproduce this error in a sandbox. I assume it is because `mbrtowc()` is not pure.

   Anyway, even if `mbrtowc()` detects this sequence as valid, the error can still be avoided by checking for `wcwidth()`'s output.
   Since the output is not checked, not only is an incorrect sequence accepted by the program, but it ends up leading to the `strcat()` code mentioned above.

   Here is the code where the error occurs:
``` {.C .numberLines startFrom=369}
		} else {
			if (len >= sizeof(cell.data))
				len = sizeof(cell.data)-1;
			for (size_t i = 0; i < len; i++)
				cell.data[i] = cur[i];
			cell.data[len] = '\0';
			cell.len = len;
			cell.width = wcwidth(wchar);
			if (cell.width == -1)
				cell.width = 1;
}
```

I will soon follow with a proposed "zero-day" patch.
   

## Further Details
I fuzzed this program using vis for a school assignment.
[Attached](https://github.com/martanne/vis/files/1800805/crashes.zip) you will find all files that made vis crash. You can analyze them as follows:
```bash
$ valgrind vis $CRASHFILE 2>&1 | sed 's/\x1b//g'
$ # sed removed escape codes for the terminal
```

Perhaps the best example is the following file (2 bytes): `accd`.

mabbamOG:
This pull request refers to issue #679

The solution is just a hotfix that is intended to leave the codebase as is, apply minimal semantic changes and prevent the buffer overflow identified in the issue linked above. The solution also takes care of error 2 highlighted in the issue linked above.

Much better solutions would be to completely redefine the way the function works, it is quite unclean from a semantic point of view.
I would separate all data and control information.
Avoid using too many unnecessary variables and try to handle `mbrtowc()` in a cleaner fashion.
Furthermore, it is unclear why many lines of code even exist to begin with (such as the one containing `strcat()`).
`view_draw` is a core part of vis and it deserves more attention.

If one does not wish to refactor the function, I suggest to at least take care of the 3 bugs highlighted by the issue linked above. It is still possible to trigger the overflow if incredibly long and particular unicode byte sequences are picked. While it is a difficult task, it does not hurt to completely prevent it from happening.

Further minor issues:
- some variables need to be initialized (e.g. `pwc`). This is good practice in C.
- `mbstate` needs to be handled accordingly. Documentation states that after finding an invalid multibyte sequence it should be reset (since effects to the state are undefined).

As a suggestion to solve all issues in one fell swoop, it might be a good idea to find a more resilient and complete all-in-one alternative to `mbrtowc()`.

### DIFF
--- a/view.c
+++ b/view.c
@@ -384,7 +384,7 @@ void view_draw(View *view) {
 		} else {
 			if (prev_cell.len && !view_addch(view, &prev_cell))
 				break;
-			pos += prev_cell.len;
+			pos += (prev_cell.len == 0? 1: prev_cell.len);
 			prev_cell = cell;
 		}
 


martanne:
I agree that the display related code is probably the ugliest part of the whole project. I'm open to suggestion on how to improve it. When redesigning it, we should also take issues like #342 and #585 into account.

The `mbrtowc` related issue should be fixed by d0efde07da335310b2e9da3bb9a690c2123fbc5e.

Your proposed patch isn't the right approach to fix the underlying problem. Please let me know whether current master works as expected? I'm closing this for now.

mabbamOG:
I have replied to everything in the previosly stated issue :+1: .
Regarding the approach for how to handle `pos`, it may be fixed both within the incomplete sequence branch or just be kept in sync. I opted for the second one for future resilience but [ae6c380](https://github.com/martanne/vis/commit/ae6c3801292c0279a6263ce2b363d0719a0737da) also resolves the infinite loop (and thus the vulnerability, at least partly).

I do not have time to further analyze that code, but take extra care when using `len`. Originally my plan was to use `pos + (cur - text)` but if `len` is used correctly that should not be a problem.

As for the remaining part of the vulnerability: an overflow may still occur at the `strcat` (now replaced with an inline equivalent) block. The two cells are appended. But what happens if both cells contain at least 8 bytes (afaik `sizeof(cell.data)` was 16)? I'm not sure if there are any unicode sequences that allow for this, or if other encodings allow for it, but in general I would not concatenate two buffers of the same size into each other.

PR #681, opened 2018-03-13
vis-complete: Non-recursive filename completion

Two-Finger:
This is the "flat" completion I [mentioned](/martanne/vis/pull/661#issuecomment-360976708) in my previous pull request.


Two-Finger:
This ended up a bit too complicated.
If you would consider merging a PR for flat completion in general, let me know. I think I can simplify it by using `realpath` to normalize the prefix instead of handling all the corner cases manually.

PR #691, opened 2018-04-03
wip: Fix: /^ doesn't respect anchor.

### LINKED ISSUE #685: /^ doesn't respect anchor
Steps
-------

- `iffffoo<esc>`
- `/^f<ret>`
- `nnnn`

Observed
-----------

- Cursor moves through each `f`.

Expected
-----------

- Cursor should remain anchored to beginning of line, on the first `f`.

nochiel:
E.g. if a line is `fffff`, search forward will no longer return all the `f`s.
Fixes #685

### DIFF
--- a/text-motions.c
+++ b/text-motions.c
@@ -618,12 +618,12 @@ size_t text_search_forward(Text *txt, size_t pos, Regex *regex) {
 	size_t start = pos + 1;
 	size_t end = text_size(txt);
 	RegexMatch match[1];
-	bool found = start < end && !text_search_range_forward(txt, start, end - start, regex, 1, match, 0);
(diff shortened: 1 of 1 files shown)


martanne:
You are on the right track, but you will need to consider line boundaries. The following file, starting with an empty line, illustrates a bug in your patch:

```

ffff
```
Also your logic for backward searches seems wrong? You will want to handle `$` in that case.

We probably also want to fix it for `:/` and `:?`

(Long term we might want to completely remove `/` and `?` in favour of the a aforementioned sam constructs, but that is an unrelated design decision.)

nochiel:
Thank you for the response.  I'll work on it some more.

martanne:
This was superseded by 886a5ed603df862a26f0d6d78d6c8dd161623738.

PR #699, opened 2018-05-14
vis: Fix a crash when re-entering prompt window

### LINKED ISSUE #628: Crashes because vis->mode is NULL in vis_run()
Hello. I found vis crashes when I do the following action:
1. Open vis
2. Enter ":vsp" and then Tab; the Ex line expands to some file path
3. Press Esc and then enter ":vsplit", then press Enter; the Ex line still expands to the same file path
4. Press Enter again, it will crash.
Based on the core dump file, it crashes at line 1394 of file vis.c.  The reason is that vis->mode points to NULL.

Two-Finger:
Fixes #628.
(A NULL check for `win->parent_mode` in `prompt_restore()` is probably a good idea, although not necessary after this change.)

While in the command window, going to Normal mode and pressing `:` (or `/`, or `?`) again would create "nested" command windows. This change makes pressing any of these keys just act like an `i`.
Another possible behavior is to go to the last line and clear it.


TmpFinger1:
Just want to mention that this patch, while solving the crash from #628 indeed, also makes it impossible to search _within_ the command window.
Back when I wrote this I didn't realize it was even an option.
Unfortunately I can't think of a less crippling fix.
@martanne, feel free to close this.

mcepl:
Hmm, I wonder what to do with this patch. Even its author doesn’t like it, it seems.

mcepl:
This should be probably closed, shouldn’t it?

rnpnr:
Since the author has long since disappeared I will close this. The related issue will remain open but any proposed fixes should keep the search functionality intact.

PR #701, opened 2018-05-16
remove indent error

thinlines:
According to the [TOML github page][1], indentation can be tabs *and/or* spaces, so `indent_error` is removed. Don't know if the way I did it is the best way.

[1]: https://github.com/toml-lang/toml#example

### DIFF
--- a/lua/lexers/toml.lua
+++ b/lua/lexers/toml.lua
@@ -44,7 +44,7 @@ local keyword = token(l.KEYWORD, word_match{
 local identifier = token(l.IDENTIFIER, l.word)
 
 -- Operators.
-local operator = token(l.OPERATOR, S('#=+-,.{}[]()'))
+local operator = token(l.OPERATOR, S('=+-,.{}[]()'))
 
 M._rules = {
   {'indent', indent},
--- a/lua/lexers/toml.lua
+++ b/lua/lexers/toml.lua
@@ -9,7 +9,7 @@ local M = {_NAME = 'toml'}
 
 -- Whitespace
 local indent = #l.starts_line(S(' \t')) *
-               (token(l.WHITESPACE, ' ') + token('indent_error', '\t'))^1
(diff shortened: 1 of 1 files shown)


martanne:
Thanks for your contribution.

I'm not really familiar with toml and I'm not completely sure what the original intend of the indent clause was, maybe @abaez can comment? I don't see right away why it can't be removed completely?

After quickly looking through the toml example within vis, the lexer seems in a bad shape:
* comments aren't recognized? PEG use ordered choice, hence the order in which the individual clauses are listed is important. Because the operator rule is listed before the comment one, the `#` is matched as an operator instead of a comment.
* the highlighting of dates seems overly specific/incomplete, `1979-05-27T07:32:00-08:00` is not highlighted correctly
* the `[table]` headings are not highlighted?

Would be nice to fix some of these issue at the same time.

olBlueEyes:
Yes, I fixed the comments issue, as well. I was going to create a separate pull request, but maybe I should have just done it as one pull request. (I don't really know how Github works.)

The issue with the comments is that the operators token is incorrectly catching the "#" character, so the comments token never sees them. I'll go ahead and add that change, too, right now.

I didn't look too much at the dates. And as for the `[table]`, I don't think there's even a rule for it.

abaez:
@olBlueEyes,

Thanks for the quick review in removing (#) out of operators rule. 

I remember when I was writing the file I didn't think much of defining comment as an operator. Back when I wrote this, I wrote for Textadept or scintilla lua lexer, which happened at the time to have a sequential read to how the lexer was parsed. To resolve issues of whether a comment was highlighted, simply switching the order of rules execution for lexers would achieve stated goal. So operator rules were always in low priority, second only to identifiers. 

Anyhow, I terribly do need to give all these lexers an overhaul. They may have not aged well... Thanks again for the help!

martanne:
Thanks, the indent error has been removed in 4af29a07f02fbc92a320206d51554f210786c73b and bb6de703f7ef309ce0ca4de829021293fb63d219 fixed the comment issue.

Issue #359, opened 2016-07-04
search broken in binary file


aeosynth:
i opened a `.png` and `.jpg` file in vis, search behaved strangely / failed. opened same files in nvim, search worked fine.

i can look for a simple test case, but i assume this will behave similarly in any binary file.

martanne:
This is currently a known limitation due to the fact that we use the `regex(3)` interface of libc which can't deal with NUL bytes. To properly fix that we need an internal regex engine based on some kind of iterator API. The [sregex](https://github.com/martanne/vis/tree/sregex) branch might work (I haven't actually tested it in a while).

In general the basic functionality for binary file editing is there, but it will require some polish to make it really usable in practice. In particular most basic cursor motions expect that they are dealing with UTF-8, there should be some kind of binary mode which switches to byte based access.

Once I have more time, I will come back to these issues.

PR #713, opened 2018-06-19
WIP: Allow regexes to contain ␀ byte

lluchs:
As previously noted in #359, vis has some trouble with searching in binary files. I noticed that the optional TRE library does actually support NUL bytes in the regex, and started modifying vis to make use of that.

With this patch, it's possible to search for NUL bytes by yanking such a pattern into the "/ register and then using n/N. It's not possible yet to enter a pattern containing NUL bytes directly, as the command line works with zero-terminated strings. Actually entering the pattern works fine, but it discards everything after the NUL byte. I haven't figured out how to change that yet.

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -1646,16 +1646,18 @@ void vis_insert_nl(Vis *vis) {
 }
 
 Regex *vis_regex(Vis *vis, const char *pattern) {
-	if (!pattern && !(pattern = register_get(vis, &vis->registers[VIS_REG_SEARCH], NULL)))
+	size_t len = 0;
+	if (!pattern && !(pattern = register_get(vis, &vis->registers[VIS_REG_SEARCH], &len)))
 		return NULL;
 	Regex *regex = text_regex_new();
 	if (!regex)
 		return NULL;
-	if (text_regex_compile(regex, pattern, REG_EXTENDED|REG_NEWLINE) != 0) {
+	if (text_regex_compile(regex, pattern, len, REG_EXTENDED|REG_NEWLINE) != 0) {
 		text_regex_free(regex);
 		return NULL;
 	}
-	register_put0(vis, &vis->registers[VIS_REG_SEARCH], pattern);
+	if (len == 0)
+		register_put0(vis, &vis->registers[VIS_REG_SEARCH], pattern);
 	return regex;
 }
 
--- a/text-regex-tre.c
+++ b/text-regex-tre.c
@@ -120,8 +120,11 @@ void text_regex_free(Regex *r) {
 	free(r);
 }
 
-int text_regex_compile(Regex *regex, const char *string, int cflags) {
-	int r = tre_regcomp(&regex->regex, string, cflags);
+int text_regex_compile(Regex *regex, const char *string, size_t len, int cflags) {
+	int r = len > 0
+	       ? tre_regncomp(&regex->regex, string, len, cflags)
+	       : tre_regcomp(&regex->regex, string, cflags);
+
 	if (r)
 		tre_regcomp(&regex->regex, "\0\0", 0);
 	return r;
--- a/text-regex.c
+++ b/text-regex.c
@@ -15,7 +15,7 @@ Regex *text_regex_new(void) {
 	return r;
 }
 
-int text_regex_compile(Regex *regex, const char *string, int cflags) {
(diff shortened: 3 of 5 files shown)


mcepl:
@rnpnr I really don’t know what to do with this one. Do we really care about binary files in `vis`?

lluchs:
Thanks for having a look at this PR! I rebased it to current master.

>I think you probably need to use tre_regnexec() instead of tre_regexec() when you actually go to perform the match.

`text_regex_match` is actually only used exactly once for a null-terminated string here: https://github.com/martanne/vis/blob/a7aac1044856abc4d1f133c6563fc604d7fe6295/sam.c#L1590

The actual matching in the text is already null-byte-safe (see `str_next_char`).

>I would also prefer that this patch was modified so that tre_regncomp() is always used in text_regex_compile() and therefore the length is always required.

I did that now and managed to make `/` commands with null bytes work. With that, I think everything works correctly now.

vis.c
```diff
@@ -1590,8 +1589,7 @@ Regex *vis_regex(Vis *vis, const char *pattern) {
 		text_regex_free(regex);
 		return NULL;
 	}
-	if (len == 0)
-		register_put0(vis, &vis->registers[VIS_REG_SEARCH], pattern);
+	register_put(vis, &vis->registers[VIS_REG_SEARCH], pattern, len+1);
```
lluchs, line 1592: The `+1` here is not very intuitive. For some reason, `register_get` gives us a length without the terminating null byte (which is still part of the buffer), but `register_put` expects a length including the null.

vis.c
```diff
@@ -1578,8 +1578,7 @@ void vis_insert_nl(Vis *vis) {
 	vis_window_invalidate(win);
 }
 
-Regex *vis_regex(Vis *vis, const char *pattern) {
-	size_t len = 0;
+Regex *vis_regex(Vis *vis, const char *pattern, size_t len) {
 	if (!pattern && !(pattern = register_get(vis, &vis->registers[VIS_REG_SEARCH], &len)))
```
rnpnr, line 1582: I'm a little confused  by this. Why does `vis_regex` need to take a length? It just gets overwritten by `register_get` right here. I think most of the changes in the second commit are cascading from this one even though it doesn't seem to do anything.

lluchs: Sorry, I somehow missed that you commented here. Thanks for having a look!

`vis_regex` can either be called with a new regex (`vis_regex(vis, "foo.*", sizeof("foo.*"))`) or without one (`vis_regex(vis, NULL, 0)`). In the latter case, it gets the pattern from the search register / (which is what happens when pressing `n`). So it does use the pattern that is passed in.

vis-motions.c
```diff
@@ -16,10 +16,10 @@ static Regex *search_word(Vis *vis, Text *txt, size_t pos) {
 	if (!buf)
 		return NULL;
 	snprintf(expr, sizeof(expr), "[[:<:]]%s[[:>:]]", buf);
-	Regex *regex = vis_regex(vis, expr);
+	Regex *regex = vis_regex(vis, expr, strlen(expr));
 	if (!regex) {
 		snprintf(expr, sizeof(expr), "\\<%s\\>", buf);
-		regex = vis_regex(vis, expr);
+		regex = vis_regex(vis, expr, strlen(expr));
```
rnpnr, line 22: These `strlen`s can be avoided by saving the return value from `snprintf`.

rnpnr (commented):
Hi Lukas,
After reviewing again I think this looks good. I just left one other comment and then this can be merged!

vis-motions.c
lluchs: I did that now by removing the ~500 character limit on the word search. That seemed to be better than handling the case where `snprintf` truncates its output.

mcepl:
@rnpnr Weren’t you saying that with that one change @lluchs made in https://github.com/martanne/vis/pull/713/commits/68e5a313b5cb124f7012cab97227af1150beb88b this PR would be ready to merge?

rnpnr:
> @rnpnr Weren’t you saying that with that one change @lluchs made in [68e5a31](https://github.com/martanne/vis/commit/68e5a313b5cb124f7012cab97227af1150beb88b) this PR would be ready to merge?

I think what happened was that I wanted to write a test for this but was having problems. I will look into that soon and merge this.

mcepl:
@rnpnr soon?

PR #718, opened 2018-07-16
build: update alpine in docker build to version 3.8

eworm-de:


eworm-de:
Closed in favor of #749

PR #741, opened 2018-10-22
lexers: ruby squiggly heredocs and code following delimiter

kj:
This adds support to the Ruby lexer for squiggly heredocs:

```
class Foo
  def call
    <<~FOO
      ...
    FOO
  end
end
```

-and also allows code between the heredoc delimiter and the end of the line, for cases like this:

```
foo(<<FOO, bar, baz)
...
FOO
```

### DIFF
--- a/lua/lexers/ruby.lua
+++ b/lua/lexers/ruby.lua
@@ -42,7 +42,7 @@ local dq_str = l.delimited_range('"')
 local lit_str = '%' * S('qQ')^-1 * literal_delimitted
 local heredoc = '<<' * P(function(input, index)
   local s, e, indented, _, delimiter =
-    input:find('(%-?)(["`]?)([%a_][%w_]*)%2[\n\r\f;]+', index)
+    input:find('([%-~]?)(["`]?)([%a_][%w_]*)%2.-[\n\r\f;]+', index)
   if s == index and delimiter then
     local end_heredoc = (#indented > 0 and '[\n\r\f]+ *' or '[\n\r\f]+')
     local _, e = input:find(end_heredoc..delimiter, e)


kj:
Also, is there still an upstream for lexers? It sounds (if I'm reading it right) from the [Scintillua page](https://foicica.com/scintillua/), like the Lua lexers have been abandoned since Scintilla 4.x. If that's the case, maybe the line:

> Changes to existing lexers should also be sent upstream for consideration.

-should be removed from the [lexers README](https://github.com/martanne/vis/blob/master/lua/lexers/README.md).

martanne:
What is the status of this?

I updated the upstream reference in ce341f24dc58f34c438f0c9bbfb3b495d032bbdf.

kj:
It looks like HEAD still doesn't support squiggly heredocs `<<~`. That's just a matter of adding a character to the lexer. The other issue (which I probably shouldn't have bundled in with this) with code following the opening delimiter on the same line, isn't really fixed ideally by this pull request. It just treats the remaining code on the line as part of the heredoc string (but at least highlights the rest of the lines in the heredoc).

In the bash lexer of the current vis release, heredocs are treated as they are in this pull request, where the remaining text on the same line is highlighted as a string. For example:

```
command <<EOF -x >/dev/null
```

The `-x >/dev/null` would all be treated as part of the heredoc, but at least the following lines would be correctly highlighted until the heredoc end delimiter.

However in vis HEAD, the bash lexer now works as it does in Ruby, where if any code follows the opening delimiter, then the entire heredoc highlighting is cancelled.

Both solutions are flawed. I'm just not sure if there's any way to stop highlighting trailing code on the first line while still highlighting the heredoc content below?

martanne:
I agree that these are two separate issues. Support for squiggly here docs was committed in 8ecc5915690112a2ea02ee3a9d4c30c017f6bb98. I think this is "good enough" for now ...

PR #764, opened 2019-06-26
filetype: Match known filenames exactly

michaelforney:
Otherwise, a file like `passwd.c` will match both ansi_c and dsv.
The one that gets chosen depends on the iteration order of table,
which is non-deterministic.

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -102,13 +102,13 @@ vis.ftdetect.filetypes = {
 		ext = { "%.d$", "%.di$" },
 	},
 	dockerfile = {
-		ext = { "Dockerfile" },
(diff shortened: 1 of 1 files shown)


lua/plugins/filetype.lua
```diff
@@ -102,13 +102,13 @@ vis.ftdetect.filetypes = {
 		ext = { "%.d$", "%.di$" },
 	},
 	dockerfile = {
-		ext = { "Dockerfile" },
+		ext = { "^Dockerfile$" },
```
rokf: Dockerfiles are allowed to have a prefix, see https://stackoverflow.com/questions/26077543/how-to-name-dockerfiles for examples.

michaelforney: Ah, I didn't know that. Updated to handle both cases.

Issue #767, opened 2019-08-08
How do I conditionally execute vis:command 's?


ThomasMeulendijks:
Hi I am trying to change my cursorline based on the mode I am in.

Since there are no mode change events in lua yet I tryed this.

```
vis.events.subscribe(vis.events.WIN_STATUS, function(win)
    if vis.mode == vis.modes.INSERT then
	vis:command('set cursorline on')
    end
    if vis.mode == vis.modes.NORMAL then
	vis:command('set cursorline off')
    end
end)
```

Unfortunately this throws two stack overflow exceptions many times upon user input.

```
vis-std.lua:83: C stack overflow
vis-std.lua:41: C stack overflow
```

Can anyone help me out to get the desired out behavior and understand the project a bit more?

ThomasMeulendijks:
Oke I got a bit further, I tried making a C patch

```
diff --git a/vis.c b/vis.c
index 03b01ab..71454c6 100644
--- a/vis.c
+++ b/vis.c
@@ -311,6 +311,8 @@ static void window_draw_cursorline(Win *win) {
 		return;
 	if (view_selections_count(view) > 1)
 		return;
+	if (!vis->mode->input)
+		return;
 
 	int width = view_width_get(view);
 	CellStyle style = win->ui->style_get(win->ui, UI_STYLE_CURSOR_LINE);

```

This works kind of, it draws the cursorline when entering insertmode but it keeps the line for one extra frame after exiting insertmode for example:

+ open new vis window | no line
+ enter insert mode | line
+ exit insert mode | line
+ pres `j` (or any other key that forces an new frame) | no line

Hope it is a bit clear :)

Any help is welcome!

ThomasMeulendijks:
This delay problem also seems to occur while using visual mode. (tag as a bug?)

PR #769, opened 2019-08-16
redraw after mode change to ensure visual effects are changed

ThomasMeulendijks:
My (succesful?) attempt at fixing my own issue #767 

In this issue the cursorline dissapears to late when mode switching.

Even when the vis.c draw_cursorline() (or something like that) is called
and propperly returns without drawing.

so the issue seems to be that the screen (or at least the portion what contains 
the cursor since the status bar [insert] does change to [normal] )

The PR is open for suggestions, this is my first contribution to the project and hope it is of any use, greetings o/

### DIFF
--- a/main.c
+++ b/main.c
@@ -2192,6 +2192,7 @@ static const char *join(Vis *vis, const char *keys, const Arg *arg) {
 
 static const char *switchmode(Vis *vis, const char *keys, const Arg *arg) {
 	vis_mode_switch(vis, arg->i);
+	vis_draw(vis);
 	return keys;
 }
 


Shugyousha (approved):
Not too familiar with this code but this LGTM!

rnpnr (commented):
Hi,

I am able to recreate the cursorline not being removed issue by using a keybinding from lua:
```lua
vis:map(m.NORMAL, " t", function()
	vis.win.options.cursorline = not vis.win.options.cursorline
end, "toggle cursorline")
```

However this commit does not fix that. Instead, with or without this commit, the following will work:

```lua
vis:map(m.NORMAL, " t", function()
	vis.win.options.cursorline = not vis.win.options.cursorline
	vis.win:draw()
end, "toggle cursorline")
```

I'm afraid I don't really see the point of this commit.

rnpnr:
I think I should close this for now. Feel free to leave a comment or open again if there is something I am missing.

PR #786, opened 2020-01-12
vis: don't draw colorcolumn after the end

zsugabubus:
Check for end-of-text.

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -286,17 +286,18 @@ static void window_draw_colorcolumn(Win *win) {
 		if (l->lineno != lineno) {
 			line_cols = 0;
 			line_cc_set = false;
-			lineno = l->lineno;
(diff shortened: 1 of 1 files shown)


vis.c
```diff
@@ -286,17 +286,18 @@ static void window_draw_colorcolumn(Win *win) {
 		if (l->lineno != lineno) {
 			line_cols = 0;
 			line_cc_set = false;
-			lineno = l->lineno;
+			if (!(lineno = l->lineno))
+				break;
```
3dc1d3: why not the much simpler and clearer:
```
lineno = l->lineno;
if (! lineno)
```

zsugabubus: I searched through the codebase and I saw similar checks, so that's why I chose this style.

PR #789, opened 2020-01-14
vis: don't search off screen when highlighting matches

TmpFinger1:
When the cursor is on a delimiter whose match should be highlighted, this patch limits the search to the viewport. If the matching delimiter is out of sight there's no point in highlighting it.

I found this when I had a very large binary file opened, and the cursor got stuck for a long time when i moved over a parenthesis.


martanne:
Thanks, merged in 3a1023ee5022071b46f659a82290bb0b3da42137 with `Filerange const *limits` changed to `const Filerange *limits` to match the surrounding style.

PR #791, opened 2020-01-22
Add layout option

kdsch:
This introduces a new `set` option for setting the layout to vertical or horizontal, which previously could not be done without creating a new window (via `split` or `vsplit`). Now, `set layout|lay h|v` will control this without creating a new window.


martanne:
Thanks, I'm going to apply this. Do you feel strongly about the `lay` abbreviation? We are using a unique prefix search for option names. That means it is currently redundant, unless we add other layout related options ...

kdsch:
@martanne I don't feel strongly about it. Looks like "la" would be the next shortest (to differentiate from "lo" for "loadmethod"). I'll put that task in my queue.

I also thought it would be nice to expose this option in the Lua API, but haven't had time to try it.

martanne:
I didn't want to suggest a different abbreviation, but to simply drop it. Even when removing the explicit `lay` entry, it would currently still be accepted as a unique prefix of `layout`.

In fact, removing it would also make `la` valid, whereas currently it doesn't work because it is a common prefix of both `lay` and `layout` and therefore not unique. This is also true for `nu`, but in that case it was probably introduced for explicit vim compatibility.

Relying on those implicit short hand versions only becomes a problem when a conflict with a new command is introduced either by modifying the C core or through the Lua API.

Ideally the Lua API should be generic and independent of specific options. I vaguely remember that there was already an existing ticket, also talking about introducing different scopes e.g. window, file/buffer, global etc.

kdsch:
@martanne Thanks, I understand better now.

PR #795, opened 2020-01-28
clean up docker build

eworm-de:


Dockerfile
```diff
@@ -1,25 +1,19 @@
...
-# make
-# docker cp vis:/tmp/vis/vis .
-# make vis-single
-# docker cp vis:/tmp/vis/vis-single .
+# Run 'make docker' to build a statically linked vis executable!
 FROM i386/alpine:3.11
-ENV DIR /tmp/vis
-WORKDIR $DIR
-RUN apk update && apk add musl-dev fortify-headers gcc make libtermkey-dev \
+RUN apk update
```
martanne: Is there a specific reason you want to split this into two commands? The [Docker documentation](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run) recommends a single `RUN` directive to avoid caching issues.

Dockerfile
```diff
@@ -1,25 +1,19 @@
...
+WORKDIR /build
 RUN wget https://github.com/martanne/libuntar/tarball/7c7247b442b021588f6deba78b60ef3b05ab1e0c -O libuntar.tar.gz && \
 	tar xf libuntar.tar.gz && cd *-libuntar-* && \
 	make && \
-	mkdir -p /usr/local/include && \
-	cp lib/libuntar.h /usr/local/include && \
-	cp lib/libuntar.a /usr/local/lib
+	mkdir -p /usr/include && \
+	cp lib/libuntar.h /usr/include && \
+	cp lib/libuntar.a /usr/lib
```
martanne: Any particular reason you prefer that?

Makefile
```diff
@@ -84,16 +85,12 @@ vis-single: vis-single.c vis-single-payload.inc
...
-	docker exec vis apk upgrade
-	docker cp . vis:/tmp/vis
-	docker exec vis sed -i '/^VERSION/c VERSION = $(VERSION)' Makefile
-	docker exec vis ./configure CC='cc --static' --enable-acl
-	docker exec vis make clean vis-single
-	docker cp vis:/tmp/vis/vis-single vis
-	docker kill vis
+	$(DOCKER) build -t vis .
+	$(DOCKER) run --rm --name vis -v $(PWD):/build vis ./configure CC='cc --static' --enable-acl
+	$(DOCKER) run --rm --name vis -v $(PWD):/build vis make VERSION="$(VERSION)" clean vis-single
```
martanne: This sets up a bind mount which makes the current working directory of the host available to the guest. I would prefer if we kept the more rigorous isolation and explicitly copy what is needed. To copy the source tree we could probably add an appropriate `COPY` instruction to  the Dockerfile. For the reverse direction `docker cp` is probably needed.

Dockerfile
eworm-de: Docker creates read-only layers for every step. Removing caches in a later step does not save the disk space as files are still available in earlier layer. Actually we do not remove any cache. This makes sense only for something like:
```
RUN apk update && apk add musl-dev ... && find /var/cache/apk/ -name 'APKINDEX.*' -delete
```
Looks like `apk` does not cache any package files, so nothing to clean there.

Dockerfile
eworm-de: No... Not sure why I changed this. I am fine with dropping this change.

Dockerfile
martanne: The point they are making is: if you add a new package to install, docker will not run the update command, but instead use the cached layer. As a result you might end up with an outdated package.

Makefile
eworm-de: The idea of docker is to build an image that has everything you need from a `Dockerfile`. The work is done running short-living and volatile containers, persistent data is stored in volumes.

Our current code starts a background container and executes commands inside this running container. If the build process breaks for whatever reason you have left over containers - running or stopped. IMHO it is really annoying to clean up a stopped or broken docker build process.

martanne: Containers can more easily be cleaned up than my working directory. I don't want to expose it to the container and let it leave files (with wrong UIDs etc) behind.

Dockerfile
eworm-de: Ah, got it...
Probably less likely to hit as we run with a defined version, but still valid.

eworm-de:
Pushed an updated version... The main change is that it uses a volume now.

martanne:
I merged the majority of your cleanups, but ultimately kept the current mechanism. Being able to use `docker exec` is convenient to make sure we have the latest packages.

If you have strong use cases which currently do not work, we can discuss further changes.

Issue #603, opened 2017-09-12
xml syntax highlighting extremely slow


peskal:
Editing large xml files with syntax highlighting on is very slow. There is a delay of almost a second each time the cursor is moved. Turning syntax highlighting off makes the cursor usable again. Here is a large xml file to test on:
https://github.com/android/platform_manifest/blob/master/default.xml

Interestingly, I have not seen this issue with other languages.

My ```:help```
```
vis v0.4

 Lua paths used to load C libraries (? will be replaced by filename):

  /usr/lib/lua/5.3/?.so
  /usr/lib/lua/5.3/loadall.so

 Compile time configuration

  Curses support:                 	yes
  Lua support:                    	yes
  Lua LPeg statically built-in:   	no
  TRE based regex support:        	no
  POSIX ACL support:              	no
  SELinux support:                	no
```
I also have ```/usr/lib/lua/5.3/lpeg.so``` on my system.

michaelforney:
I've noticed this as well. Editing XML with syntax highlighting is unusably slow, so I disable it in my `visrc.lua`:

```lua
if win.file.path and win.file.path:find('%.xml$') then
  vis:command('set syntax off')
end
```

I did a bit of investigation, and found that if I comment out `{'equals', equals}` in `lexers/xml.lua`, it performs much better. I suspect https://github.com/martanne/vis/blob/d7e0cbd7795bec43e9e9f11a42a7019f18bf9e18/lua/lexers/xml.lua#L24 might be the culprit.

Shugyousha:
On Tue, Sep 12, 2017 at 10:45 PM, Michael Forney
<notifications@github.com> wrote:
> I've noticed this as well. Editing XML with syntax highlighting is unusably
> slow, so I disable it in my visrc.lua:
>
> if win.file.path and win.file.path:find('%.xml$') then
>   vis:command('set syntax off')
> end

Yes, that is one of the viable workarounds. The other one is to limit
the syntax highlighting search space ("horizon") to a managable value
using https://github.com/martanne/vis/blob/d7e0cbd7795bec43e9e9f11a42a7019f18bf9e18/lua/vis-std.lua#L34
.


Cheers,

Silvan

martanne:
If not fixed, this should at least be improved by 7e9e0a2ca868aaa214fb38a79fe71da34d6e00da. I'm closing this issue for now, feel free to re-open/comment if you still encounter performance problems.

PR #797, opened 2020-01-31
lexers: XML performance improvement

TmpFinger1:
The new pattern isn't perfect, but the old one wasn't either.

One of its purposes was to highlight 123, but not 456 below:

```xml
<tag attr=123>text=456</tag>
```

Both the new and the old pattern can be tricked.
Here, 123 is not recognized as a token:
```xml
<tag glitch=">" attr=123 glitch="<">text=456</tag>
```

You can test it by positioning the cursor over one of the numbers and hitting `dii`.
If the _whole_ number is deleted, then it's recognized as a token.
(At least in the themes I use, numbers, even when recognized as such, are not highlighted any differently than text...)

Edit: forgot to mention, this is about issue #603.

### DIFF
--- a/lua/lexers/xml.lua
+++ b/lua/lexers/xml.lua
@@ -20,13 +20,7 @@ local dq_str = l.delimited_range('"', false, true)
 local string = #S('\'"') * l.last_char_includes('=') *
                token(l.STRING, sq_str + dq_str)
 
-local in_tag = P(function(input, index)
-  local before = input:sub(1, index - 1)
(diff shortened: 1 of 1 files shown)


TmpFinger1:
Just noticed that the exact same pattern is used in the `html` and `wsf` lexers.
In `html` it's commented out, that's why it's fast.

martanne:
Thanks, I applied the same change to the other lexers as well.

PR #798, opened 2020-02-01
build: properly kill docker container

eworm-de:
We keep a running docker container in the background. If the build
process fails this becomes dangling. Make sure the container is properly
killed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,10 @@ vis-single: vis-single.c vis-single-payload.inc
 	${CC} ${CFLAGS} ${CFLAGS_AUTO} ${CFLAGS_STD} ${CFLAGS_EXTRA} $< ${LDFLAGS} ${LDFLAGS_STD} ${LDFLAGS_AUTO} -luntar -llzma -o $@
 	${STRIP} $@
 
-docker: clean
+docker-kill:
+	$(DOCKER) container ls --filter=name=vis -q | grep -q . && $(DOCKER) kill vis && sleep 1 || true
+
+docker: docker-kill clean
 	$(DOCKER) build -t vis .
 	$(DOCKER) run --rm -d --name vis vis tail -f /dev/null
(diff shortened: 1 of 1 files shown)


eworm-de:
Sorry, still not happy. :-p

Makefile
```diff
@@ -84,7 +84,10 @@ vis-single: vis-single.c vis-single-payload.inc
 	${CC} ${CFLAGS} ${CFLAGS_AUTO} ${CFLAGS_STD} ${CFLAGS_EXTRA} $< ${LDFLAGS} ${LDFLAGS_STD} ${LDFLAGS_AUTO} -luntar -llzma -o $@
 	${STRIP} $@
 
-docker: clean
+docker-kill:
+	$(DOCKER) container ls --filter=name=vis -q | grep -q . && $(DOCKER) kill vis && sleep 1 || true
```
martanne: Doesn't the following, simplified version also work?

```
$(DOCKER) kill vis && $(DOCKER) wait vis || true
```

eworm-de: I think you did two changes here:

- removed the check for existence of container

Fine with me, I added this to not add error message to a successful build:

```
Error response from daemon: Cannot kill container: vis: No such container: vis
```

- changed `sleep` to `docker wait`

`docker kill` just sends a kill signal, but does not wait for the container to exit and clean up. Running the container before it is gone makes it fail, thus added the `sleep`. I was unsure about `docker wait`, help says it blocks _until one or more containers stop_. So does it wait for the container to be _removed_ (we have `docker run --rm`?
Did some tests and it looks ok. So fine with me.

eworm-de: We could also change this to:
```
-$(DOCKER) kill vis && $(DOCKER) wait vis
```
... which would give just another error message from make:
```
make: [Makefile:88: docker-kill] Error 1 (ignored)
```

martanne:
Thanks, applied.

PR #800, opened 2020-02-06
lua: add <plugin>/init.lua to package.path

TmpFinger1:
This patch makes vis follow a Lua convention about modules, and also makes it easier to keep plugins up to date.
```lua
require'plugin'
```
will match both `plugin.lua` and `plugin/init.lua`.

Currently, most of the existing single-file plugins have the following structure - `vis-plugin/plugin.lua`, which is not a Lua convention or standard. One either has to

- download the individual file, losing the possibility to easily update it via git-pull,
- clone the repo, and use `require'vis-plugin.plugin'`
- clone the repo, add `~/.config/vis/vis-?/?.lua` to `package.path`, and use `require'plugin'`

With this change, just cloning a repo and `require`-ing its name would be enough,
and you get to use git-pull to check for any updates.

It doesn't harm the existing plugins in any way, AFAICT.

### DIFF
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -2567,8 +2567,10 @@ bool vis_lua_path_add(Vis *vis, const char *path) {
 	lua_getglobal(L, "package");
 	lua_pushstring(L, path);
 	lua_pushstring(L, "/?.lua;");
-	lua_getfield(L, -3, "path");
-	lua_concat(L, 3);
+	lua_pushstring(L, path);
+	lua_pushstring(L, "/?/init.lua;");
+	lua_getfield(L, -5, "path");
+	lua_concat(L, 5);
 	lua_setfield(L, -2, "path");
 	lua_pop(L, 1); /* package */
 	return true;


erf:
@TmpFinger1 what do you think of a plugin manager like [vim-plug](https://github.com/junegunn/vim-plug) for `vis` once we have some conventions? It could be a plugin in itself e.g. `vis-plug` and you'd describe your plugins in the visrc.lua file ( or elsewhere? ) and you'd be able to install and update plugins into a default folder e.g. `$HOME/.config/vis/plugins`, on startup or using a `vis-install` or `vis-update` command, which would use `wget`, `curl` or `git` on the list of plugins. We could default to github urls or use full urls for source. This might make it easier to setup an editor with third party plugins. Overkill? Doable?

erf:
Regarding my previous comment, i made a prototype for a vis plugin manager here:
https://github.com/erf/vis-plug

It is very simple per now.. You configure a list of plugin paths before requiring the plug plugin, it will then automatically download and require the plugins upon start.

TmpFinger1:
@erf I already have a "plugin manager" - https://repo.or.cz/vis-bundle.git
I don't even use it.

martanne:
Applied as b87a64ec90ced056b96b1833477bc3253f393531.

Issue #652, opened 2017-12-30
segfault/memory corruption when piping buffer contents through a command in FILE_SAVE_PRE hook


zenhack:
Given this visrc.lua:

```lua
require("vis")

vis.events.subscribe(vis.events.FILE_SAVE_PRE, function(file, path)
   vis:command(",|cat")
end)
```

I get frequent memory-corruption related errors after saving a file; sometimes double-free assertions, sometimes segfaults. I originally hit this trying to use `goimports`, but per the above it triggers even when used with a command that doesn't actually change the text. I can reproduce this on both v0.4 and master.

martanne:
Thanks for the report, the issue is that the [transcript structure](https://github.com/martanne/vis/blob/6557d0ee823418d3e78a4133fd5d5af9b5319bca/vis-core.h#L152) only exists once per file. As a result it gets corrupted in these nested commands. The `w`rite command will initialize (and free it) while in the meantime the `cat` command does so too.

This will require some code restructuring to fix (i.e. there should be one transcript per file and command).

Furthermore, there are some semantic issue to deal with. Currently the address of the write command (by default the whole file) is calculated *before* the hook is executed. It will thus only write the range corresponding to the initial file size. In case the hook adds additional text it is silently ignored. If the hook removes text, the write fails because not enough data was available.

We could try to delay the address calculation until *after* the hook is completed. In any case this needs a bit more thoughts.

bugabinga:
This can be reliably reproduced on OpenBSD 6.3+.
Attaching core dump:
[core.zip](https://github.com/martanne/vis/files/2173817/core.zip)

hyphenrf:
don't know if related but this also causes a similar segfault:
```lua
vis:command_register('paste',
    function() vis:command('<cat /dev/tty') end,
    'paste mode, end with C-d'
)
```

lepotic:
@hyphenrf I believe it deserves a separate issue.
Also related - #357

PR #809, opened 2020-02-21
sam: re-take range after pre-save hooks are run

### LINKED ISSUE #715: Undefined error 0 in FILES_SAVE_PRE handler when mutating buffer
Given following `visrc.lua`:

```lua
 -- trim trailing whitespace before saving                                       
 vis.events.subscribe(vis.events.FILE_SAVE_PRE, function(file, _)                
   local lines = file.lines                                                      
   for i=1, #lines do
     local trimmed = lines[i]:match('^(.-)%s+$')                                 
     if trimmed then lines[i] = trimmed end                                      
   end                                                                           
   -- proceed with saving the file...                                            
   return true                                                                   
 end) 
```
I see the info message:

```sh
Can't write `/home/oliver/.config/vis/visrc.lua': Undefined error: 0
```

To reproduce:

1. Put trailing whitespace somewhere in a file.
2. Save with `:w`.
3. Error appears as info. File is trimmed, but not saved.
4. Save again. No error, file is saved.

Might be related to https://github.com/martanne/vis/issues/652.

```sh
oliver:114$ vis -v
vis v0.5 +curses +lua +tre
```

### LINKED ISSUE #806: FILE_SAVE_PRE hook: file text is cut on write
Consider this:

```lua
-- somewhere in $HOME/.config/vis/format.lua
local function dofmt(file, path)
  local win = vis.win
  if win.syntax ~= "cpp" and win.syntax ~= "ansi_c" then
    return true
  end

  local fmt = "clang-format"
  local status, out, err = vis:pipe(file, {start = 0, finish = file.size}, fmt)
  if status ~= 0 then
    if err then
      vis:message(err)
    end
    return false
  end
  local pos = win.selection.pos
  file:delete(0, file.size)
  file:insert(0, out)
  win.selection.pos = pos
  return true
end

vis.events.subscribe(vis.events.FILE_SAVE_PRE, dofmt)
```

In visrc.lua: `require('format')`.

```sh
$ cat /tmp/hello.c
#include <stdio.h>
int main() { printf("hello\n"); printf("hello\n"); printf("hello\n"); }

$ vis +wq /tmp/hello.c
$ cat /tmp/hello.c
#include <stdio.h>
int main() {
  printf("hello\n");
  printf("hello\n");
  printf("hello\n

$ tail -n1 /tmp/hello.c  | hexdump -C
00000000  20 20 70 72 69 6e 74 66  28 22 68 65 6c 6c 6f 5c  |  printf("hello\|
00000010  6e                                                |n|
00000011
```

Am I doing something wrong or is it a bug?

x1ddos:
A vis.events.FILE_SAVE_PRE callback may mutate the file text,
making the original range passed to cmd_write incorrect.

It is unclear how to realign the range after the callback
is done for the cases where an active selection in visual mode
is present or an explicit range like :1,2w! was specified.

However, this commit resolves the issue for the case where
the whole file is expected to be written.

Tests are in https://github.com/martanne/vis-test/pull/19.

Closes #806
Closes #715
Updates #652

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1573,6 +1573,10 @@ static bool cmd_substitute(Vis *vis, Win *win, Command *cmd, const char *argv[],
 	return false;
 }
 
+/* cmd_write stores win->file's contents end emits pre/post events.
+ * If the range r covers the whole file, it is updated to account for
+ * potential file's text mutation by a FILE_SAVE_PRE callback.
+ */
 static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Selection *sel, Filerange *r) {
 	if (!win)
 		return false;
@@ -1582,6 +1586,9 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 		return false;
 
 	Text *text = file->text;
+	Filerange range_all = text_range_new(0, text_size(text));
+	bool write_entire_file = text_range_equal(r, &range_all);
+
 	const char *filename = argv[1];
 	if (!filename)
 		filename = file->name;
@@ -1599,6 +1606,9 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			vis_info_show(vis, "Rejected write to stdout by pre-save hook");
 			return false;
 		}
+		/* a pre-save hook may have changed the text; need to re-take the range */
+		if (write_entire_file)
+			*r = text_range_new(0, text_size(text));
 
 		bool visual = vis->mode->visual;
 
@@ -1624,8 +1634,7 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			vis_info_show(vis, "WARNING: file will be reduced to active selection");
 			return false;
 		}
-		Filerange all = text_range_new(0, text_size(text));
-		if (!text_range_equal(r, &all)) {
+		if (!write_entire_file) {
 			vis_info_show(vis, "WARNING: file will be reduced to provided range");
 			return false;
 		}
@@ -1648,6 +1657,9 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			vis_info_show(vis, "Rejected write to `%s' by pre-save hook", path);
 			goto err;
 		}
+		/* a pre-save hook may have changed the text; need to re-take the range */
+		if (write_entire_file)
+			*r = text_range_new(0, text_size(text));
 
 		TextSave *ctx = text_save_begin(text, path, file->save_method);
 		if (!ctx) {


x1ddos:
ubuntu-16.04 seems to have an unrelated failure:

```
Compiling keys utility
cc -pipe -Os -ffunction-sections -fdata-sections -fPIE  keys.c -Wl,-z,now -Wl,-z,relro -ltermkey -lncursesw -ltinfo -o keys
keys.c: In function ‘printkey’:
keys.c:138:7: error: ‘TERMKEY_TYPE_DCS’ undeclared (first use in this function)
  case TERMKEY_TYPE_DCS:
       ^
keys.c:138:7: note: each undeclared identifier is reported only once for each function it appears in
keys.c:139:7: error: ‘TERMKEY_TYPE_OSC’ undeclared (first use in this function)
  case TERMKEY_TYPE_OSC:
       ^
Makefile:4: recipe for target 'keys' failed
```

martanne:
Thanks and sorry about the unrelated CI failures (I already fixed them on another machine). I will try to review your changes over the weekend.

sam.c
```diff
@@ -1582,6 +1582,9 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 		return false;
 
 	Text *text = file->text;
+	Filerange range_all = text_range_new(0, text_size(text));
```
martanne: Bellow you use `text_object_entire`, to be more consistent I would both times use the same construct.

Which one doesn't really matter, I would probably chose the explicit `text_range_new`.

sam.c
```diff
@@ -1599,11 +1602,19 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			vis_info_show(vis, "Rejected write to stdout by pre-save hook");
 			return false;
 		}
+		/* a pre-save hook may have changed the text; need to re-take the range */
+		range_all = text_object_entire(text, 0);
```
martanne: how about:
```
if (write_entire_file)
    *r = text_range_new(0, text_size(text));
```
and leaving the loop body as is?

The same comment also applies to the other hunk.

sam.c
x1ddos: Done.

sam.c
x1ddos: The thing is, it'll of course modify the caller's original range and I'm not sure if this will cause any side effects.

x1ddos: Ah wait, did you mean to do it on purpose, as if to somehow signal the caller that the range has changed?

martanne: Yeah I would do it on purpose. Currently it is probably only used by the grouping logic (which has other issues). The looping constructs are other candidates. But as mentioned earlier, changing the content while operating on it doesn't really fit sam's model.

x1ddos: Ok. Done. Also, added a doc comment for `cmd_write`.
I wish `cmd_write` took `Filerange r` instead of `Filerange *r` - not a big fan of mutating args like that.

Anyway, I briefly checked and it seems the range passed to `cmd_write` isn't used anywhere afterwards. Also, a similar thing already happens in `sam_execute` at https://github.com/martanne/vis/blob/d2407f9ebc089e11eb12054f614a37ba6dbcb1ef/sam.c#L1120 so I guess it shouldn't introduce new unwanted side effects. I was just worried about potential regressions in the future.

PR #814, opened 2020-02-28
sam: initialize struct

eworm-de:
We do not want random data in case stat() fails.

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1702,7 +1702,7 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			file_name_set(file, path);
 			same_file = true;
 		}
-		if (same_file)
+		if (same_file || !existing_file)
 			file->stat = text_stat(text);
 		vis_event_emit(vis, VIS_EVENT_FILE_SAVE_POST, file, path);
 		free(path);
--- a/sam.c
+++ b/sam.c
@@ -1646,7 +1646,7 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 		if (!path)
 			return false;
 
-		struct stat meta;
+		struct stat meta = { 0 };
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -1702,7 +1702,7 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 			file_name_set(file, path);
 			same_file = true;
 		}
-		if (same_file)
+		if (same_file || !existing_file)
```
martanne: While this fixes the case of opening an non-existing file, there is another issue illustrated by:

    $ vis exisiting-file
    :w non-existing-file
    :w

In this case we update `file->stat` with the data of the wrong file, resulting in a warning when attempting to write the one initially opened.

I will shortly push a change with:

    if (same_file || (!existing_file && strcmp(file->name, path) == 0))

which should hopefully fix this.

sam.c
```diff
@@ -1646,7 +1646,7 @@ static bool cmd_write(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 		if (!path)
 			return false;
 
-		struct stat meta;
+		struct stat meta = { 0 };
 		bool existing_file = !stat(path, &meta);
 		bool same_file = existing_file && file->name &&
 		                 file->stat.st_dev == meta.st_dev && file->stat.st_ino == meta.st_ino;
```
martanne: I don't think this is necessary because:

- if stat(2) fails, it returns -1
- existing_file is !-1 which is 0
- same_file is 0 && [...] which is 0

but meta is only accessed if these variables are true.

eworm-de: It is not necessary, but I was confused by strange numbers while debugging. Feel free to ignore.

sam.c
eworm-de: Thanks!

eworm-de:
Well... Thanks again. :P

PR #818, opened 2020-03-08
Set single cursor style as primary, not secondary

3dc1d3:
Currently vis defaults to rendering a single cursor using the secondary cursor color theme style - not sure how others feel about it but to me it makes more sense that it should be the primary, since there are no secondary cursors visible.

So that's what this patch does - your main cursor is always the same color, and any additional cursors are colored different.

If people disagree then perhaps we can add an option to toggle this behavior?


3dc1d3:
Good to know, will do.  Cheers.

PR #819, opened 2020-03-08
color-column: Don't change fg/bg if not set explicitly

3dc1d3:
eg. if your long line is a comment with green fg, and you set your
column color bg red while not specifying the fg, then the result is
green fg on red bg.

Prior to this change the result would be default fg on red bg, thus
one char in the long line of green text would look odd/wrong.

Of course if you do explicitly set the column color fg to default in your
theme then the result will not be what you expect - ideally we need
an UNSPECIFIED color type instead of relying on DEFAULT.

The `#if CONFIG_CURSORS` block is not ideal but I'm not sure what would be the right place to put it as currently all those defines are in other source files - is there a reason they're not in a header file?

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -27,6 +27,8 @@
 #include "util.h"
 #include "vis-core.h"
 #include "sam.h"
+#include "ui.h"
+
 
 static void macro_replay(Vis *vis, const Macro *macro);
 static void macro_replay_internal(Vis *vis, const Macro *macro);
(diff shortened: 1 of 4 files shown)


3dc1d3:
Found a neater solution.  Tested it with and without curses.

vis.c
```diff
@@ -294,7 +296,10 @@ static void window_draw_colorcolumn(Win *win) {
 
 		/* This screen line contains the cell we want to highlight */
 		if (cc <= line_cols + width) {
-			l->cells[(cc - 1) - line_cols].style = style;
+			CellStyle* orig = & l->cells[(cc - 1) - line_cols].style;
```
martanne: The placement of `*` and `&` isn't really inline with the rest of the code base. Also while you are changing that code section, the parentheses are unnecessary:

```
CellStyle *orig = &l->cells[cc - 1 - line_cols].style;
```

Furthermore, there is no need for two separate commits (also applies to #818). If you squash and rebase, I will merge it.

Thanks

PR #845, opened 2020-06-30
vis-open: add "/" to each dir entry

vkochan:
It allows to easy identify if the entry is file or directory.

### DIFF
--- a/vis-open
+++ b/vis-open
@@ -8,6 +8,15 @@ NL='
 VIS_MENU_PROMPT=""
 ALLOW_AUTO_SELECT=1
 
+wrap_dir() {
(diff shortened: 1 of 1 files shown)


vkochan:
Looks like 'ls -F' should work better here, will re-submit it.

martanne:
Hi, I don't think the `ls` based approach is preferable because it also prints other indicators e.g. for symlinks.

Also POSIX does not specify any options for `echo`.

vkochan:
So, what about to handle output from "ls" in separate func and append "/" if [ -d ] (actually it was my
initial version but -F looked more simpler) ?

PR #847, opened 2020-07-03
vis-open: add trailing "/" for the folders

vkochan:
Add trailing "/" for the folder entries which allows to easy
differentiate folders and the regular files. Additionally it allows
easy filter only folders by simply enter "/" in the vis-open prompt.


martanne:
Thanks, applied as f34f6d968e711f3cab019a12bb60da1bbf52161f with a slight tweak to the printf invocation.

PR #848, opened 2020-07-06
Makefile: add distclean target

paride:


paride:
I know `config.h` and `config.mk` do not get included in the `make dist` generated tarball. The change is mainly to slightly ease the downstream packaging.

PR #849, opened 2020-07-08
Support for Primary Clipboard

jeremybobbin:


vis-registers.c
```diff
@@ -262,23 +274,24 @@ Array vis_register_get(Vis *vis, enum VisRegister id) {
...
-	[VIS_REG_8]          = { '8', VIS_HELP("8th sub-expression match")                         },
-	[VIS_REG_9]          = { '9', VIS_HELP("9th sub-expression match")                         },
-	[VIS_REG_AMPERSAND]  = { '&', VIS_HELP("Last regex match")                                 },
-	[VIS_REG_BLACKHOLE]  = { '_', VIS_HELP("/dev/null register")                               },
-	[VIS_REG_CLIPBOARD]  = { '*', VIS_HELP("System clipboard register, see vis-clipboard(1)")  },
-	[VIS_REG_DOT]        = { '.', VIS_HELP("Last inserted text")                               },
-	[VIS_REG_SEARCH]     = { '/', VIS_HELP("Last search pattern")                              },
-	[VIS_REG_COMMAND]    = { ':', VIS_HELP("Last :-command")                                   },
-	[VIS_REG_SHELL]      = { '!', VIS_HELP("Last shell command given to either <, >, |, or !") },
-	[VIS_REG_NUMBER]     = { '#', VIS_HELP("Register number")                                  },
```
martanne, line 283: Please avoid changing the code formatting while also introducing logical changes. Either do it in a separate commit, or preferably use a shorter help text.

vis-registers.c
```diff
@@ -188,6 +199,7 @@ bool register_resize(Register *reg, size_t count) {
 enum VisRegister vis_register_from(Vis *vis, char reg) {
 	switch (reg) {
 	case '+': return VIS_REG_CLIPBOARD;
+	case '*': return VIS_REG_PRIMARY;
```
martanne, line 202: Instead of adding this you can remove the entry above for `+`. Both cases should be covered by the for loop further down.

vis-clipboard
```diff
@@ -76,27 +80,27 @@ vc_paste() {
 }
 
 vc_wlclipboard_copy() {
-	wl-copy -t TEXT
+	wl-copy --"$sel" -t TEXT
 }
 
 vc_wlclipboard_paste() {
-	wl-paste -t text
+	wl-paste --"$sel" -t text
```
martanne, line 87: Does your `wl-{copy,paste}` actually support a `--clipboard` option? It seems to be the default and only `--primary` is recognized.

jeremybobbin: Thanks! That does not appear to be the case.
I appreciate your feedback.

martanne:
Thanks, applied with a slight tweak to fix compiler warnings in `make debug` mode.

PR #852, opened 2020-07-26
Improve performance of view_selections_dispose_all

mokulus:


martanne (changes requested):
In general you are right that due to the renumbering of the selections this is currently an _O(n^2)_ algorithm.

However, you are introducing a use-after-free bug, that is you are accessing `s` after it has been freed. That is the reason why various tests fail. Try running `make test` to see whether your changes work as expected.

One possible approach would be to walk the selection list backwards. That would first require to find the end, but still have linear time complexity.

Another option is to free all but the primary selection directly and then manually set the selection related fields such as `selections`, `selection_dead`, `selection_latest`, `selection_count` etc. That is a bit ugly though.

PR #876, opened 2020-09-18
Add ignorecase option

deepcube:
Add a global ignorecase boolean option.  When set add REG_ICASE to cflags
when calling text_regex_compile().


deepcube:
I couldn't make up my mind as to where in each of the lists (sam.c, vis-cmds.c, vis-core.h, vis.1) to put ignorecase so I just added it to the bottom of each. I was tempted to put it with the other boolean settings but they seemed to all be about display settings so I wasn't sure if it belonged with them. I'm happy to move things around or have you move them around to keep a more consistent style/order.

deepcube:
Updated to reduce line length as per IRC discussion

PR #881, opened 2020-10-20
Improve behavior of inner text objects when working linewise

autumn-canfield:
Currently inner and outer text objects both behave exactly the same for linewise operations. I think it would make a lot more sense to limit inner text objects to only the lines contained within the range. So for example right now `>i{` and `>a{` from with in this if
```c
if (linewise && (a->textobj->type & TEXTOBJECT_DELIMITED_INNER)) {
r.start = text_line_next(txt, r.start);
r.end = text_line_prev(txt, r.end);
}
```
both result in
```c
	if (linewise && (a->textobj->type & TEXTOBJECT_DELIMITED_INNER)) {
	r.start = text_line_next(txt, r.start);
	r.end = text_line_prev(txt, r.end);
	}
```

With these changes `>i{` would only affect the lines contained inside like so:

```c
if (linewise && (a->textobj->type & TEXTOBJECT_DELIMITED_INNER)) {
	r.start = text_line_next(txt, r.start);
	r.end = text_line_prev(txt, r.end);
}
```

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -976,6 +976,9 @@ void vis_do(Vis *vis) {
 				if (a->textobj->type & TEXTOBJECT_DELIMITED_OUTER) {
 					r.start--;
 					r.end++;
+				} else if (linewise && (a->textobj->type & TEXTOBJECT_DELIMITED_INNER)) {
+					r.start = text_line_next(txt, r.start);
+					r.end = text_line_prev(txt, r.end);
 				}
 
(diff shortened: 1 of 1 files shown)


moesasji:
This patch works for me and makes vis behave the same as vim on this example for these two motions.

rnpnr (approved):
This seems correct to me. It seems like something that was missed since without his patch the behaviour of visual mode commands such as `vi{` and `va{` are not consistent with this indentation example.

rnpnr:

PR #892, opened 2020-11-14
vis: add a ScrollType for window scrolling and remove default case

erf:
This makes the window scrolling a bit more explicit and clearer to read.

The default case is not used (as far as i can see) and can therefor be removed.
`view_scroll_up` and `view_scroll_down` can still be used for arbitrary scrolling.

### DIFF
--- a/main.c
+++ b/main.c
@@ -21,8 +21,12 @@
 #include "array.h"
 #include "buffer.h"
 
-#define PAGE      INT_MAX
-#define PAGE_HALF (INT_MAX-1)
+enum ScrollType {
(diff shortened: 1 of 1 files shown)


martanne:
I don't think it is more readable and as the faulting tests indicate, it changes the behavior/introduces bugs. (The default case is used through `selections_navigate`).

erf:
I can only see `selections_navigate` passing the `PAGE_HALF` argument which passed that on directly to `wscroll`, so that should not trigger the default case, but I'm probably missing something here..

PR #893, opened 2020-11-16
text: ignore unused-result in error path

eworm-de:
We are in error path and there's nothing we can do. Just ignore.

As clang tries to be compatible with gcc this should be portable.

### DIFF
--- a/text-io.c
+++ b/text-io.c
@@ -242,7 +242,10 @@ static int mkstempat(int dirfd, char *template) {
 	fd = mkstemp(template);
 err:
 	if (cwd != -1) {
+#pragma GCC diagnostic ignored "-Wunused-result"
+		/* we are in error path and there's nothing we can do */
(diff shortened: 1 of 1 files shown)


martanne:
I don't like to clutter the source with ugly `#pragma` stuff just to shut up compiler warnings. It might also not be supported by more niche compilers like `cproc` which are known to compile vis without problems. Casting the return value to `void` would be acceptable, but it seems like gcc decided to ignore that long standing idiom and warn anyway.

Also the the code path is currently never taken because the function is always called with `AT_FDCWD`. It was done in preparation for some experimentation with a client/server architecture where the the file descriptor could be passed from the client to the server.

eworm-de:
Yes, all current solutions are more or less ugly... No idea why casting to `void` does no longer work.
How about something like this?
```
if (fchdir(cwd)) /* ignore return */ ;
```

PR #896, opened 2020-11-25
lua: get lua to recognize bashrc file

ccao001:


ninewise:
Wouldn't you want to leave off the `%`? IIRC that's for the rest of the filename, but there should not be anything there in this case.

ccao001:
@ninewise 

Thank you for catching that! I edited the last commit and pushed to this branch in order to update. It should reflect what you suggested.

Let me know if there are any other issues with this PR.

all the best.

lepotic:
@ccao001 The correct Lua pattern for that is `"^%.bashrc$"`. The percent sign is an escape character in Lua patterns - https://www.lua.org/manual/5.3/manual.html#6.4.1

erf:
Need the escape char "%", but not the "start-of-line" char "^" i think, as you'd only want the end of the line.

ninewise:
Oh, excuse me, seems I mixed up all those grammars again.

lepotic:
The `^` is needed, as `.bashrc` is the whole name, hence the pattern needs to be anchored on both ends.

The code is a bit misleading, using `ext` as key. It is up to each pattern whether to match the extension or the whole name.
There are a few other filetypes that match the whole name as well - Dockerfile, crontab, PKGBUILD, etc.

martanne:
@lepotic is right about the correct pattern to use.

But more fundamentally, I don't really want to add a listing of all possible shell related files. Why don't we add entries for `.bash_profile`, `.bash_logout`, ... ? And this is only for bash, what about all other shells (currently we use the same lexer for all)?

If you really want this, then patch it locally or tweak your `file(1)` configuration to recognize it properly.

PR #902, opened 2020-12-06
Adding Zig lexer.

KaneRoot:


rokf:
You're missing the  `lua/plugins/filetype.lua` entry.

KaneRoot:
Done. Should be okay now.

PR #904, opened 2020-12-08
lexers: add mail lexer

silasdb:


martanne (changes requested):
Could you please enforce that header field names need to start with a capital letter? Otherwise URLs at the start of lines are also highlighted.

Would it make sense to recognize inline PGP blocks? MIME sections? Mail signatures?

In general I wouldn't spend much time on it, but a more strict header field name rule seems useful.

silasdb:
On Tue, Dec 08, 2020 at 09:29:16AM -0800, Marc André Tanner wrote:
>Could you please enforce that header field names need to start with a capital letter? Otherwise URLs at the start of lines are also highlighted.

Sure, I will do that.

>Would it make sense to recognize inline PGP blocks? MIME sections? Mail signatures?
>
>In general I wouldn't spend much time on it, but a more strict header field name rule seems useful.

Perfect.  I'll work on that.  Besides improving header field parsing, and adding PGP blocks, MIME sections and mail signatures, what else am I missing?

lua/plugins/filetype.lua
```diff
@@ -247,6 +247,13 @@ vis.ftdetect.filetypes = {
 		ext = { "%.lua$" },
 		mime = { "text/x-lua" },
 	},
+	mail = {
+		detect = function(file, data)
```
martanne, line 251: Please use `detect = function(_, data)` here, otherwise luacheck reports a warning because of the unused function argument, see #918.

martanne (changes requested):
Please consider migrating the lexer to the new format, see #908 and #916.

Thanks

ninewise:
Could you send the changes to the lexers upstream instead, to https://github.com/orbitalquark/scintillua? We're going to avoid diverging. The filetype changes are welcome, though, when you've send the actual lexer upstream.

PR #907, opened 2020-12-09
fix comment typos in vis.h

moesasji:
This pull request fixes typos in comments that I spotted reading through the code.

### DIFF
--- a/array.h
+++ b/array.h
@@ -100,7 +100,7 @@ bool array_truncate(Array*, size_t length);
  * Change length.
  * @rst
  * .. note:: Has to be less or equal than the capacity.
- *           Newly accesible elements preserve their previous values.
+ *           Newly accessible elements preserve their previous values.
  * @endrst
  */
 bool array_resize(Array*, size_t length);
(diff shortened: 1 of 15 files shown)


martanne (changes requested):
Hi, I am willing to merge this, if you drop the NUL -> NULL change. The former is generally used to refer to the [NUL character](https://en.wikipedia.org/wiki/Null_character) wheres the later denotes a NULL pointer.

The usual objections to such changes (producing unnecessary merge conflicts, cluttering commit history etc.) are less of an issue due to our small development community.

moesasji:
> Hi, I am willing to merge this, if you drop the NUL -> NULL change. The former is generally used to refer to the [NUL character]

Learned something as I wasn't aware of the NUL usage and I can definitely fix that. Do you prefer to have all this edits as a single commit to have less pollution?

btw) more used to the openbsd approach where typos in comments do get fixed....

martanne:
One commit is fine, but I can also merge a patch series.

I agree that fixing typos makes sense. In retrospect my remark is more directed at white space only changes.

Thanks for looking through the code/comments!

PR #914, opened 2020-12-15
vis-open: allow to show files vertically

vkochan:
It might be more comfortable to navigate through the files
arranged vertically.

Add ability to specify VIS_OPEN_LINES environment variable
which is passed to vis-menu as -l option.

It would be better to add vis option for this to set it
via lua config, but it might be added later.

### DIFF
--- a/vis-open
+++ b/vis-open
@@ -5,6 +5,10 @@ set -e
 NL='
 '
 
+if [ -z "$VIS_OPEN_LINES" ]; then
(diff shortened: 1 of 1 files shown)


mcepl:
Merged into https://git.sr.ht/~mcepl/vis/ (`master` branch).

rnpnr:
Applied! Thanks for the patch!

PR #915, opened 2020-12-19
Remove restriction that f,t character motion are line-based 

### LINKED ISSUE #735: multiline f/t
https://github.com/martanne/vis/wiki/FAQ#what-are-structural-regular-expressions

> The main insight [of sam] is that text editing should not be artificially limited by the concept of a line.

the `f` and `t` motions only search for characters on the current line; instead, it should work across lines. Multiple vim plugins exist to add this behavior, and kakoune supports it by default.

moesasji:
This pull request makes the char motions f,F and t,T work more powerful by removing the restriction that they work on lines.

This behavior is implemented by many plugins for vim and is the default behavior for Kakoune. In my opinion removing this restriction fits better with the sam concept that operations should not be line-based without affecting the vim grammar.

This pull-request would close #735 

btw) this change breaks some of the motion-based tests. I can look at fixing those if this would be acceptable.

### DIFF
--- a/vis-motions.c
+++ b/vis-motions.c
@@ -114,9 +114,7 @@ static size_t longword_next(Vis *vis, Text *txt, size_t pos) {
 
 static size_t to(Vis *vis, Text *txt, size_t pos) {
 	char c;
-	if (pos == text_line_end(txt, pos))
-		return pos;
-	size_t hit = text_line_find_next(txt, pos+1, vis->search_char);
+	size_t hit = text_find_next(txt, pos+1, vis->search_char);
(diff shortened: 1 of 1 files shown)


martanne:
The reason for the current behavior is that with multiple selections it is often convenient to only adjust selections which actually have a match on the current line. The multiline behavior seems somewhat redundant with `?` and `/`, albeit without the need to quote special characters.

Anyway, adf65e1dca8d3c0381133be8446eebddf412d772 introduces the multiline functionality which can be enabled by overriding the various mappings.

Issue #823, opened 2020-04-02
Broken syntax highlighting for blocks in Markdown lexer (and fix idea)


xoich:
The syntax highlighting of indented blocks, like code blocks, works only on the first line of the file! The reason is that the blocks pattern is checked for being at the beginning of a line, but this never happens, as the whitespace token will absorb all the indentation after a newline.

My fix idea is to split the whitespace token into whitespace proper and newlines:

patch to `lexers/markdown.lua`
``` {.diff}
11c11,12
< local ws = token(l.WHITESPACE, l.space^1)
---
> local ws = token(l.WHITESPACE, S("\t\v ")^1)
> local nl = token(l.WHITESPACE, l.newline^1)
75a77
>   {'newline', nl},
```

Issue #726, opened 2018-08-08
Performance issues in YAML highlighting


peaceant:
Hello.

I'm editing .yml file (link below) in vis editor and the performance is really terrible. Scrolling and editing text is really laggy. Why is that? It's only 286 lines of yaml code. I'm editing in vis editor 500000 lines of xml code with better peformance.

[custom-theme.yml](https://github.com/peaceant/perf-vis/blob/master/custom-theme.yml)

Shugyousha:
Hi

On Wed, Aug 08, 2018 at 12:26:26AM -0700, Martin Radimec wrote:
> I'm editing .yml file (link below) in vis editor and the performance
> is really terrible. Scrolling and editing text is really laggy. Why
> is that? It's only 286 lines of yaml code. I'm editing in vis editor
> 500000 lines of xml code with better peformance.

Yes, there seem to be some issues with the syntax highlighting in some
cases. Big XML files show similar symptoms.

> [custom-theme.yml](https://github.com/peaceant/perf-vis/blob/master/custom-theme.yml)

I have recorded some perf sample data showing where the time is spent
when scrolling around in the file in vis. Interestingly enough, the
farther down you scroll, the slower everything gets.

The perf data indicates that most of the time is spent in the lua string
matching functions. The recorded perf data can be downloaded from
[here](https://sillymon.ch/data/perf-data-for-custom-theme-yml.data.gz).

I wanted to create a flame graph from the perf data but the flamegraph
scripts didn't work. I wonder if there is another way to get an easier
interpretable view of the data.

lanodan:
Maybe one way would be to have a configurable limit to syntax highlighting depending on something like the file size.

erf:
I also noticed it now when scrolling down to the end of [this file](https://github.com/tmcdonell/config-alacritty/blob/master/alacritty.yml)

TmpFinger1:
For YAML, you can try [this patch](https://repo.or.cz/vis/gkirilov.git/blobdiff/9eff1bc595a465572f79bfcb5c9e1ccf1f4a65b7..0cc2a5be96ebffe262842ca5751df8ee816fb430:/lua/lexers/yaml.lua).

It is only some of the lexers that have terrible performance. It's the way they are written - they ignore the "horizon" and re-read the whole file from the very beginning, multiple times.

erf:
@peaceant i believe 79859b57fc3556c2ed74950a5a5668c64face4e3 solved this issue

rokf:
[...]

PR #916, opened 2020-12-20
Scintillua

moesasji:
This is a first stab at bringing the lexers in sync with upstream. What has been done:

1) all lexers unchanged in vis since last sync updated.
2) added upstream lexers not present in vis.
3) where patches were obvious the lexer was updated with the changes pushed upstream as pull-requests.
4) opened a pull-request for all lexers in vis not in upstream, but left them in the legacy format for now.

There are still a number of lexers to do, but those are not that obvious or the commits I see in vis seems to have removed functionality. I'll add a list of those to the bug-report on this.

### DIFF
--- a/lua/lexers/README.md
+++ b/lua/lexers/README.md
@@ -2,8 +2,7 @@ Lua LPeg lexers for vis
 =======================
 
 Vis reuses the [Lua](http://www.lua.org/) [LPeg](http://www.inf.puc-rio.br/~roberto/lpeg/)
-based lexers from the [Scintillua](http://foicica.com/scintillua/) project
-which is now part of the [Scintilla 3.x branch](https://foicica.com/hg/scintilla/file/tip/lexlua).
+based lexers from the [Scintillua](https://orbitalquark.github.io/scintillua/index.html) project.
 
 # Vis integration
 
@@ -26,13 +25,13 @@ where `<name>` corresponds to the filename without the `.lua` extension.
 
 To add a new lexer, start with the template quoted below or a lexer of a
 similiar language. Read the
-[lexer module documentation](http://foicica.com/scintillua/api.html#lexer).
+[lexer module documentation](https://orbitalquark.github.io/scintillua/api.html#lexer).
 The [LPeg](http://www.inf.puc-rio.br/~roberto/lpeg/) introduction might also
 be useful.
 
 For development purposes it is recommended to test the lexers from a lua
 script as described in the
-[Scintillua manual](http://foicica.com/scintillua/manual.html#Using.Scintillua.as.a.Lua.Library).
+[Scintillua manual](https://orbitalquark.github.io/scintillua/manual.html#Using.Scintillua.as.a.Lua.Library).
 
 To enable auto syntax highlighting when opening a file you can associate your
 new lexer with a set of file extensions by adding a corresponding entry into
@@ -42,35 +41,57 @@ Changes to existing lexers should also be sent upstream for consideration.
 
 A template for new lexers:
 
-```
+```lua
+-- Copyright 2006-2020 Mitchell. See LICENSE.
 -- ? LPeg lexer.
 
-local l = require('lexer')
-local token, word_match = l.token, l.word_match
-local P, R, S = lpeg.P, lpeg.R, lpeg.S
+local lexer = require('lexer')
+local token, word_match = lexer.token, lexer.word_match
+local P, S = lpeg.P, lpeg.S
 
-local M = {_NAME = '?'}
+local lex = lexer.new('?')
 
 -- Whitespace.
-local ws = token(l.WHITESPACE, l.space^1)
+lex:add_rule('whitespace', token(lexer.WHITESPACE, lexer.space^1))
 
-M._rules = {
-  {'whitespace', ws},
-}
+-- Keywords.
+lex:add_rule('keyword', token(lexer.KEYWORD, word_match[[
+  keyword1 keyword2 keyword3
+]]))
 
-M._tokenstyles = {
+-- Identifiers.
+lex:add_rule('identifier', token(lexer.IDENTIFIER, lexer.word))
 
-}
+-- Strings.
+local sq_str = lexer.range("'")
+local dq_str = lexer.range('"')
+lex:add_rule('string', token(lexer.STRING, sq_str + dq_str))
 
-return M
+-- Comments.
+lex:add_rule('comment', token(lexer.COMMENT, lexer.to_eol('#')))
+
+-- Numbers.
+lex:add_rule('number', token(lexer.NUMBER, lexer.number))
+
+-- Operators.
+lex:add_rule('operator', token(lexer.OPERATOR, S('+-*/%^=<>,.{}[]()')))
+
+-- Fold points.
+lex:add_fold_point(lexer.KEYWORD, 'start', 'end')
+lex:add_fold_point(lexer.OPERATOR, '{', '}')
+lex:add_fold_point(lexer.COMMENT, lexer.fold_consecutive_lines('#'))
+
+return lex
 ```
 
 # Color Themes
 
-The `../themes` directory contains the color schemes. Depending on the
-number of colors supported by your terminal, vis will start with either
-the `default-16` or `default-256` theme. Symlink it to your prefered
-style or add a command like the following one to your `visrc.lua`:
+The [`../themes directory`](../themes) contains the color
+schemes. Depending on the number of colors supported by your terminal,
+vis will start with either the [`default-16`](../themes/default-16.lua)
+or [`default-256`](../themes/default-256.lua) theme. Symlink it to
+your prefered style or add a command like the following one to your
+`visrc.lua`:
 
 ```
 vis:command("set theme solarized")
@@ -79,4 +100,4 @@ vis:command("set theme solarized")
 # Dependencies
 
  * [Lua](http://www.lua.org/) 5.1 or greater
- * [LPeg](http://www.inf.puc-rio.br/~roberto/lpeg/) 0.12 or greater
+ * [LPeg](http://www.inf.puc-rio.br/~roberto/lpeg/) 1.0.0 or greater


martanne (changes requested):
Thanks for tackling this!

A few general/administrative comments:

- please use one commit per lexer to import the upstream version
- then apply one commit per lexer with our changes on top of it
- start the commit message with `lexers/$lang:` followed by a short description

More concretely, this would result into commit history similiar to this:
- `lexers: update copyright dates`
- `lexers/actionscript: import upstream revision 09e0ff0`
- `lexers/ada: import upstream revision 09e0ff0`
- ...
- `lexers/jq: new upstream lexer revision 09e0ff0`
- ...
- `lexers/elixir: import upstream revision 09e0ff0` in the long description you can mention that the white space fix is already present.
- `lexers/perl: import upstream revision 09e0ff0`
- `lexers/perl: fix long regex patterns` preserve the long description from the original vis commit
- ...

I hope you get the idea. This allows to more easily revert language specific changes and incorporate changes once merged upstream.

Thanks again!

moesasji:
I get the idea and now working on cleaning this up made easier by @orbitalquark already merging the pull-requests that made sense.

lua/plugins/filetype.lua
```diff
@@ -210,6 +210,9 @@ vis.ftdetect.filetypes = {
 	javascript = {
 		ext = { "%.cjs$", "%.js$", "%.jsfl$", "%.mjs$", "%.ts$" },
 	},
+  jq = {
+    ext = { "%.jq$" },
+  },
```
martanne, line 215: The indentation is off here. Please try to match the surrounding style (using tabs in this case).

martanne (commented):
This is moving into the right direction, thanks! I added a minor style related inline comment.

We will eventually have to update the lexer README with updated references to the new API and lexer template.

> There are still a number of lexers to do, but those are not that obvious or the commits I see in vis seems to have removed functionality. I'll add a list of those to the bug-report on this.

Please provide this list. The removed functionality is likely due to performance issues (e.g.  #797, #726), maybe upstream also has suggestions how to improve it. In general the same methodology, i.e. first committing the upstream version followed by our own changes on top, should be applicable.

Thanks again.

lua/plugins/filetype.lua
moesasji: Pushed a commit that fixes this indentation as my default is indeed spaces.

moesasji:
> We will eventually have to update the lexer README with updated references to the new API and lexer template.

Agreed, just left that one for now.

> Please provide this list. The removed functionality is likely due to performance issues (e.g. #797, #726), maybe upstream also has suggestions how to improve it. In general the same methodology, i.e. first committing the upstream version followed by our own changes on top, should be applicable.

Most of the ones that I haven't yet touched are ones that required a bit more thought as I opted to do the easy ones first. The list of files that still need a look with comments added:

- [ ] README.md 
- [ ] ansi_c.lua            (at first glance changes didn't look compatible)
- [ ] bash.lua              (more than just a couple)
- [ ] erlang.lua            (vis commit removes a lot and appears wrong)
- [ ] fennel.lua            (not the same origin - vis version seems more extensive)
- [ ] html.lua               (performance fix)
- [ ] markdown.lua     (not sure how vis changes fit with upstream)
- [ ] php.lua                (more than just a couple, not an easy one)
- [ ] scheme.lua         (more than just a couple, not an easy one)
- [ ] text.lua                (upstream empty, vis whitespace only)
- [ ] toml.lua               (initially didn't see why indent error got removed)
- [ ] wsf.lua                 (performance fix)
- [ ] xml.lua                 (performance fix)
- [ ] yaml.lua               (performance fix)

btw) the ones labeled performance fix should be not be difficult, but I just left them out in my first stab at it.

lepotic:
> * [ ]  fennel.lua            (not the same origin - vis version seems more extensive)

Vis version is not more extensive, it's just that the vis copy of Scintillua does not support inheritance.
Upstream Scintillua made it possible lexers to inherit from each other, in order to avoid duplication:

[fennel.lua:9](https://github.com/orbitalquark/scintillua/blob/d7c687add9288daf0e7060f986e7a923cadd194d/lexers/fennel.lua#L9)

martanne:
> - [ ] bash.lua              (more than just a couple)

The most recent changes are related to here documents, @silasdb can probably make sure they work as expected with the upstream version? The commit message of 21727ae3dce8ccea2e342d3b62b5fa67b1b5a936 also gives an example with variable substitution.

> - [ ] erlang.lua            (vis commit removes a lot and appears wrong)

I am not exactly sure what the intention was in #632, the description is very vague. Maybe @emfa can clarify, until then drop the changes and go with the upstream version.

> - [ ] markdown.lua     (not sure how vis changes fit with upstream)

Check whether #823 still happens with the upstream version. (Not sure whether it actually works correctly in vis either).

> - [ ] php.lua                (more than just a couple, not an easy one)

The issue fixed in 0ff02a70f911327e8a1f92881f5ef43b7cd0e7c8 is that `?>` should [break out of PHP mode](https://www.php.net/manual/en/language.basic-syntax.comments.php). The other changes just add constants and keywords, it is probably best to fetch the latest ones from the the documentation and incorporate them into the upstream version. 

> - [ ] scheme.lua         (more than just a couple, not an easy one)

64554981c786c7a95e64c9f4336d3624d7239672 has descriptive commit message, maybe @ii8 would be willing to update it to the new upstream API

> - [ ] text.lua                (upstream empty, vis whitespace only)

This is such that replacement symbols are visible, see 807ba6f6850bf7dbf86fd65299248f2d9cba7f75. It is not relevant for upstream. We should just convert it to the new API.

> - [ ] toml.lua               (initially didn't see why indent error got removed)

This might have been a misunderstanding on my part.

moesasji:
> > * [ ]  text.lua                (upstream empty, vis whitespace only)
> 
> This is such that replacement symbols are visible, see [807ba6f](https://github.com/martanne/vis/commit/807ba6f6850bf7dbf86fd65299248f2d9cba7f75). It is not relevant for upstream. We should just convert it to the new API.
> 
> > * [ ]  toml.lua               (initially didn't see why indent error got removed)
> 
> This might have been a misunderstanding on my part.

To prevent double work. I've opened pull request on these and several other ones, see: [upstream](https://github.com/orbitalquark/scintillua/pulls) That would tell whether or not they apply to upstream.

silasdb:
Hi! Is it a better practice to send lexers PRs to [upstream](https://github.com/orbitalquark/scintillua) instead of sending them to vis?

martanne:
> Is it a better practice to send lexers PRs to upstream instead of sending them to vis?

In principle sending them upstream is preferable. However, this assumes both projects use the same format which is not yet the case, but is the goal of this PR. Hence, if you could make sure that the upstream `bash.lua` works as expected and contains all the necessary fixes that would be appreciated.

moesasji:
I'm hitting a bit of a glitch with [the upstream markdown lexer](https://github.com/orbitalquark/scintillua/blob/default/lexers/markdown.lua). Out of the box the upstream version fails to render the headers in red and none of the patches that are applied to the legacy version in vis would affect that behavior. 

As the code in the lexer looks fine to me....is this problem caused at this end, i.e. [did we miss something in the changes made in vis.lua](https://github.com/martanne/vis/commit/5ca74646fc8015b705e81f574deeeaf5ce77f278) to be able to switch away from the legacy lexer?

lepotic:
Hi @moesasji!

I think the glitch is due to the Markdown lexer using themeable colors (as opposed to hard-coded ones):
[markdown.lua:19](https://github.com/orbitalquark/scintillua/blob/d7c687add9288daf0e7060f986e7a923cadd194d/lexers/markdown.lua#L19)
[markdown.lua:54](https://github.com/orbitalquark/scintillua/blob/d7c687add9288daf0e7060f986e7a923cadd194d/lexers/markdown.lua#L54)

Some themes have a `colors` table defined and exported - [solarized.lua:23](https://github.com/martanne/vis/blob/40c387aca214f701cd316c63b290ba7f35abbe3a/lua/themes/solarized.lua#L23), but the default one doesn't.
(And even solarized.lua doesn't currently have the `black` color for the `hr` rule above.)
If you do `:set theme solarized`, you'll see that the headers become red.

For a quick test, you can try replacing `lexer.colors.red` with `'red'`.
But in the long term, I think it's better to add a `colors` table to all the themes, so vis could use unmodified lexers from Scintillua.

martanne:
It should be fixed by abee98578c0ea4d18aeb14b1e5ee66b67be2be89?

moesasji:
> It should be fixed by [abee985](https://github.com/martanne/vis/commit/abee98578c0ea4d18aeb14b1e5ee66b67be2be89)?

That indeed fixes the issue for me. Markdown headers have the right color again. Thanks for the quick fix and @lepotic for spotting it!

btw) with this being a subset probably worth keeping fixing the color theme on the todo-list to get the colors one would expect.

lepotic:
> It should be fixed by [abee985](https://github.com/martanne/vis/commit/abee98578c0ea4d18aeb14b1e5ee66b67be2be89)?

Yes, and it's better than what I suggested (duplicating the table in each theme). :)

moesasji:
Not sure what I did wrong to trigger build failures for this recent sets of commits. Any thoughts on where the mistake is would be greatly appreciated?

In any case this should be getting closer. Things left to do:

- [ ] update README.md (already added upstream template to point to)
- [ ] ansi_c.lua (versions are close, but needs a more detailed look; opened upstream issue: https://github.com/orbitalquark/scintillua/issues/21)
- [ ] bash.lua (hopefully @silasdb can take a look at whether the upstream version fixes their issues)
- [ ] scheme.lua (imported upstream version, which should work; opened issue for help merging our patch: https://github.com/orbitalquark/scintillua/issues/24)
- [ ] jq.lua (triggers a deprecated function warning: probably easy to fix, opened issue: https://github.com/orbitalquark/scintillua/issues/23)
- [ ] markdown.lua reintroduces bug #823 with syntax highlighting in code blocks, yet it has the fix included already. Tested with [markdown-test-file](https://github.com/mxstbr/markdown-test-file).

lua/lexers/text.lua
```diff
@@ -1,15 +1,13 @@
--- Copyright 2006-2017 Mitchell mitchell.att.foicica.com. See LICENSE.
+-- Copyright 2006-2020 Mitchell. See LICENSE.
 -- Text LPeg lexer.
 
-local l = require('lexer')
+local lexer = require('lexer')
+local token, word_match = lexer.token, lexer.word_match
+local P, S = lpeg.P, lpeg.S
```
martanne, line 6: Please remove all the unnecessary code here.

martanne (changes requested):
Haven't looked yet what might be wrong, maybe it is related to the text lexer which is used as a default?

Could you please drop your merge commit and instead rebase the changes on top of my latest scintillua branch.

Also some of the latest commit messages related to the performance improvements could probably be improved a bit.

Thanks again!

lua/lexers/text.lua
moesasji: 1) I stuck to the default part of the template, but will take a look at what I can drop from it.

2) I thought I had rebased on top of your scintillua branch, but clearly need to 'fight' a bit more with git.

moesasji: I've made the changes requested and also split the push in two parts, which (after a long wait) indeed shows that modifying the text lexer to the new format is what triggers the build failures.

I can retry without the white space patch, but that will probably make no difference?

lua/lexers/text.lua
```diff
@@ -1,4 +1,10 @@
 -- Copyright 2006-2020 Mitchell. See LICENSE.
 -- Text LPeg lexer.
 
-return require('lexer').new('text')
+local lexer = require('lexer')
+local lex = lexer.new('text')
+
+-- Whitespace
+lex.add_rule('whitespace', lexer.token(lexer.WHITESPACE, lexer.space^1))
```
lepotic, line 8: @moesasji it has to be `lex:add_rule` (a colon, not a dot).

lepotic:
I found where the Markdown lexer goes wrong, but don't have a solution yet.
Actually, it's not the lexer's fault (literally the same lexer works fine in Textadept), more like the vis integration is missing something.

If you do

```diff
-lex:add_style('code', lexer.styles.embedded .. {eolfilled = true})
+lex:add_style('code', {back = 'red'})
```

code blocks will be highlighted.

`lexer.styles.embedded` is a table with a metatable, adding some magic I don't understand.
There are more lexers that use `lexer.styles.*` and are probably subtly broken as well.
I'll see what I can do tomorrow.

lua/lexers/text.lua
moesasji: The joy of being far too used to using C based languages. Thanks for spotting this typo!

moesasji:
> Actually, it's not the lexer's fault (literally the same lexer works fine in Textadept), more like the vis integration is missing something.
> 
> If you do
> 
> ```diff
> -lex:add_style('code', lexer.styles.embedded .. {eolfilled = true})
> +lex:add_style('code', {back = 'red'})
> ```
> 
> code blocks will be highlighted.

It appears to be the same cause as the earlier markdown issues as these styles appear to be defined in the color-themes used for textadept, see for example [this one](https://github.com/shosca/base16-gruvbox/blob/master/textadept/base16-gruvbox-light.lua#L52). 

The fun might be that vis only supports a subset based on the text [here](https://github.com/martanne/vis/wiki/Themes)

martanne:
The issue is indeed related. Not sure what the best/most elegant way to fix it is. [This patch seems to work](http://ix.io/2JeQ) in my *very* brief test.

That also suggest we should probably review the (default) themes.

moesasji:
With the last commits I've pushed we've reached the state I'm pretty much done with the bits I'm confident dealing with. To summarize what is remaining:

- [ ] fix issues with theme templates
- [ ] scheme patch that @ii8  contributed still needs to be applied: upstream issue [here](https://github.com/orbitalquark/scintillua/issues/24)
- [ ] ansi_c had a couple of updates not yet applied: upstream issue [here](https://github.com/orbitalquark/scintillua/issues/21)
- [ ] bash needs fix for nested variables: upstream issue [here](https://github.com/orbitalquark/scintillua/issues/25)
- [ ] bash upstream needs heredoc patches that @silasdb has contributed; our fix actually looks non-optimal as there must be a way to detect either '<< Delimiter' or '<<-Delimiter' with the same LPeg expression instead of catching cases.

martanne:
Thanks. I rebased once again, incorporating the latest upstream changes to the PHP, jq and text lexers.

I also moved the lexer template back into the README for now.

Another TODO item is to port existing lexers using either deprecated functions or the old style:

- [ ] clojure, @Gavlooth
- [ ] elm, @kdsch
- [ ] dmd, upstream
- [ ] dsv, is this lexer used/useful in practice? @eworm-de
- [ ] fantom, @MarSoft 
- [ ] fstab, @eworm-de 
- [ ] gemini, @lanodan 
- [ ] git-rebase, myself?
- [ ] julia, @ckafi
- [ ] meson, @fischerling 
- [ ] networkd, @eworm-de 
- [ ] pony, @ii8 
- [ ] reason, @roguh
- [ ] rest, upstream?
- [ ] routeros, @eworm-de 
- [ ] sml, upstream?
- [ ] snobol, upstream?
- [ ] spin, @TieDyedDevil 
- [ ] strace, myself?
- [ ] systemd, @eworm-de 
- [ ] taskpaper, upstream?
- [ ] vcard, upstream?
- [ ] xs, @TieDyedDevil
- [ ] yaml, upstream?
- [ ] zig, @Karchnu (deleted github profile)

I saw that upstream (@orbitalquark) also has an [issue tracking the progress](https://github.com/orbitalquark/scintillua/issues/1).

eworm-de:
Was this closed by intention?

moesasji:
> Was this closed by intention?

No, this issue is still being worked on. Github for some reason closes an issue when the branches are in sync as I made the mistake to push to my branch after the rebase.

moesasji:
@martanne Did you have any luck figuring out the best way to fix the issue with the new color theming as mentioned [here](https://github.com/martanne/vis/pull/916#issuecomment-750030160)?

mcepl:
According to [upstream](https://github.com/orbitalquark/scintillua/issues/1#issuecomment-775333309) “All lexers except rest have been migrated.”

PR #918, opened 2020-12-23
Luacheck fixes

moesasji:
Running `make luacheck` showed what seem to be two potential bugs. I think I managed to fix the first one in the first commit. How to fix the second one isn't clear. The warning is as follows:

`
Checking lua/plugins/filetype.lua                 1 warning

    lua/plugins/filetype.lua:542:6: accessing undefined variable mime
`

To me this warning triggered by this [line](https://github.com/martanne/vis/blob/master/lua/plugins/filetype.lua#L538) indeed looks like this part of the code wouldn't pick up the mime variable from the defined file-extensions, so I don't think this is a false positive. Unfortunately not at all clear to me how to fix this one. 

While there I also silenced the "unused argument warnings" as one can't pass ignore values to `make luacheck` and not returning false positives is probably the better approach seeing that the developer documentation recommends to run make luacheck [here](https://github.com/martanne/vis/wiki/Developer-Overview#luacheck).

### DIFF
--- a/lua/plugins/complete-filename.lua
+++ b/lua/plugins/complete-filename.lua
@@ -13,8 +13,8 @@ vis:map(vis.modes.INSERT, "<C-x><C-f>", function()
 	local prefix = file:content(range)
 	if not prefix then return end
 	-- Strip leading delimiters for some languages
-	i, j = string.find(prefix, "[[(<'\"]+")
-	if j then prefix = prefix:sub(j + 1) end
+	local i, j = string.find(prefix, "[[(<'\"]+")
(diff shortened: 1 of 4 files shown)


martanne:
In general I don't like to clutter the source with workarounds just to make some diagnostic tools happy. This also recently came up in a C context #893.

Most of the warnings should be fixed by d2d6b29f7f86f9bfedeb6c8050f68cda5196b1e4, in Lua `_` is typically used as a variable name for unused results.

moesasji:
Thanks for fixing those. Note that you still have the warning that I didn't know how to fix, which does look like a genuine bug to me.

PR #928, opened 2021-02-06
add .clang-format code format file

sheeaza:
~~'.cache' and 'compile_commands.json' are used by clangd~~

~~Signed-off-by: sheeaza <winsdenny@outlook.com>~~

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,118 @@
+# SPDX-License-Identifier: GPL-2.0
+#
+# clang-format configuration file. Intended for clang-format >= 4.
+#
+# For more information, see:
+#
+#   Documentation/process/clang-format.rst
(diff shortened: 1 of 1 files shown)


ninewise:
IMHO editor/compiler-specific files belong in your global gitignore/configuration, not in each project you edit with it. On the other hand, there are already vi-specific files in the gitignore.

sheeaza:
so what about the .clang-format commit for code format

ninewise:
That's up to the maintainer, I guess.

martanne:
I agree with @ninewise that editor specific files should be kept out of the repository.

As stated elsewhere I am not a big fan of whitespace only changes. They cause unnecessary merge conflicts and clutter source control history etc. Adding configuration for source formatting and then not enforcing it also doesn't make much sense. Therefore I'm closing this for now.

PR #936, opened 2021-03-08
filetype: Add COMMIT_EDITMSG to diff.

parazyd:
This is useful when writing a commit message in git.

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -96,7 +96,7 @@ vis.ftdetect.filetypes = {
 		ext = { "%.desktop$" },
 	},
 	diff = {
-		ext = { "%.diff$", "%.patch$", "%.rej$" },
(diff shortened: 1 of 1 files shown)


martanne:
Why do you think this is useful? You will have to motivate your changes. Do you typically have diffs in your commit messages?

Also we already have a [section for `COMMIT_EDITMSG`](https://github.com/martanne/vis/blob/0cccd6ef42b99d00c51a865c8fbf72ad9bd02958/lua/plugins/filetype.lua#L163) which sets the column marker to indicate when the commit message is too long. In fact, a proper commit message lexer could ensure that other commit message conventions (e.g. shorter subject line, empty second line etc.) are respected.

parazyd:
If you have git.commit.verbose set to true, then it shows you the entire commit in your $EDITOR. So this is where the diff can be colored.

PR #944, opened 2021-04-21
lexers/scad: Add basic lexer and filetype for openscad

HalosGhost:
Largely cribbed from cpp.lua


ninewise:
Lexers should be submitted to the https://github.com/orbitalquark/scintillua project.

PR #945, opened 2021-04-24
ui: add bell notification (beep)

joukewitteveen:
The bell is sounded when there is nothing to undo/redo.
This is consistent with vim, although vim has many more beeps cases.

---
I wanted to browse the vis code-base a bit and decided that this was a good excuse to do so.
In VIM, the bell can be silenced by configuring a "visual bell" (screen flash) and setting the visual bell command to the empty string. I reckon implementing anything like a "visual bell" is out of scope for vis (it can be configured in your terminal emulator anyway), but we might want an option to silence the bell still.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1875,7 +1875,7 @@ static const char *mark(Vis *vis, const char *keys, const Arg *arg) {
 
 static const char *undo(Vis *vis, const char *keys, const Arg *arg) {
 	size_t pos = text_undo(vis_text(vis));
-	if (pos != EPOS) {
+	if (vis_verify_cursor(vis, pos)) {
 		View *view = vis_view(vis);
 		if (view_selections_count(view) == 1)
 			view_cursor_to(view, pos);
@@ -1887,7 +1887,7 @@ static const char *undo(Vis *vis, const char *keys, const Arg *arg) {
 
 static const char *redo(Vis *vis, const char *keys, const Arg *arg) {
 	size_t pos = text_redo(vis_text(vis));
-	if (pos != EPOS) {
+	if (vis_verify_cursor(vis, pos)) {
 		View *view = vis_view(vis);
 		if (view_selections_count(view) == 1)
 			view_cursor_to(view, pos);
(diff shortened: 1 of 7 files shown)


rnpnr:
I don't see anything wrong with this code but besides this pull request
no one is complaining about vis' lack of a terminal bell. Personally I
really don't like it so I'm not interested in applying this.

But, since this isn't `rnpnr/vis` I'm not closing this and will let
others decide. Most people probably won't want to comment so
you can leave a reaction to this post vote.

:+1:  : I agree DON'T merge this.
:-1: : I disagree DO merge this.

mcepl:
My other problem with this PR is that it does more than what’s in the commit message. It seems to be playing with the colour palette more than just what’s necessary to introducing the bell notification (at least unnecessary formatting changes).

rnpnr:
> the addition of a bell is a matter of accessibility

Maybe I'm ignorant but is it possible to use vis without looking at
the screen? If that's the case there is a possibility that I can be
convinced. Otherwise I think I will close this.

erf:
With an option to silence the bell (defaulted to off), i don't mind this change.

joukewitteveen:
> With an option to silence the bell (defaulted to off), i don't mind this change.

I would consider that useless code that serves as nothing but a maintenance burden. The audible aspect of the terminal bell can be configured in your terminal emulator. The bell is sounded in many more prevalent circumstances than at the end of undo/redo sequences in your editor. Tab completion for example also typically sounds the bell. Have you actually tried `vis` with the change? The change is fairly minimal.

erf:
> > With an option to silence the bell (defaulted to off), i don't mind this change.
> 
> I would consider that useless code that serves as nothing but a maintenance burden. The audible aspect of the terminal bell can be configured in your terminal emulator. The bell is sounded in many more prevalent circumstances than at the end of undo/redo sequences in your editor. Tab completion for example also typically sounds the bell. Have you actually tried `vis` with the change? The change is fairly minimal.

Yes I've tried it and don't mind the default bell sound, especially when i see i can configure my terminal (Wezterm) on how to respond to it. Not sure all users agree though

mcepl:
I think we need a decision about this issue, @rnpnr , either merge this or reject it, but this lingers here.

rnpnr:
It looks to me like most people do not want this patch. For the time being it will have to remain local for anyone who wants it. Feel free to link it in the [wiki](https://github.com/martanne/vis/wiki/Patches) if you want.

Issue #941, opened 2021-04-05
vis-menu can't handle diatrics/accents


jpaulogg:
Hello, thanks for the great work. I'm from Brasil and I miss diatrics and accents in vis-menu prompt. I have no knowledge of C, so I can't help... Is this a difficult task?

Shugyousha:
hi!

I have added a commit to a branch that should make `vis-menu` slightly less broken in regards to UTF-8 handling. You can find it here: https://github.com/Shugyousha/vis/tree/try-to-preserve-codepoints

Let me know if that at least improves things for you! If it does, maybe we can try to get the change merged.

jpaulogg:
Nice! Thanks for the answer.

Unfortunately, it didn't work. My main need is related to word completion. For example, I can't complete the word "atenção", vis-menu shows only "aten".

Shugyousha:
> Unfortunately, it didn't work. My main need is related to word completion. For example, I can't complete the word "atenção", vis-menu shows only "aten".

This is actually a different issue but it turns out that `vis-menu` had to be changed anyways. I have added another fix (for `vis-complete` which is used for word completion) that fixes the issue for me. Can you try again?

jpaulogg:
So the problem was also with the vis-complete script! With your modification now the word completion is working. Thanks!

Shugyousha:
> So the problem was also with the vis-complete script! With your modification now the word completion is working. Thanks a lot!
> 
> Your solution defines words by separation of blank characters, right? I preferred to fix the problem using another pattern like '[a-zA-ZçÇãÃ]' for the tr program.

If you wanted to do it this way, you would have to add any possible combination of diacritics to the pattern. This would be quite a few and rather error prone. My version has its own issues (it won't work for languages without word spaces for example) but I would argue that it's an acceptable 80/20 solution.

jpaulogg:
> > So the problem was also with the vis-complete script! With your modification now the word completion is working. Thanks a lot!
> > Your solution defines words by separation of blank characters, right? I preferred to fix the problem using another pattern like '[a-zA-ZçÇãÃ]' for the tr program.
> 
> If you wanted to do it this way, you would have to add any possible combination of diacritics to the pattern. This would be quite a few and rather error prone. My version has its own issues (it won't work for languages without word spaces for example) but I would argue that it's an acceptable 80/20 solution.

That's true. My pattern works for my situation, but using '[:blank:]' is a wider solution.

PR #946, opened 2021-05-02
Improve word and file name completion with non-ASCII characters

Shugyousha:
These two commits fix this issue https://github.com/martanne/vis/issues/941.

Note that `vis-menu` is still not handling Unicode correctly (it's still not differentiating properly between bytes and "grapheme clusters", i.e. printable glyph-combinations, when calculating lengths). With these changes we are at least able to have a working support for non-ASCII characters in the simple case.

### DIFF
--- a/vis-menu.c
+++ b/vis-menu.c
@@ -84,10 +84,44 @@ appenditem(Item *item, Item **list, Item **last) {
 
 static size_t
 textwn(const char *s, int l) {
-	int b, c; /* bytes and UTF-8 characters */
+	int c;
 
-	for(b=c=0; s && s[b] && (l<0 || b<l); b++) if((s[b] & 0xc0) != 0x80) c++;
+	for(c=0; s && s[c] && (l<0 || c<l); ) c++;
 	return c+4; /* Accomodate for the leading and trailing spaces */
+}
+
+/*
+ * textvalidn returns the highest amount of bytes <= l of string s that
+ * only contains valid Unicode points. This is used to make sure we don't
+ * cut off any valid UTF-8-encoded unicode point in case there is not
+ * enough space to render the whole text string.
+*/
+static ssize_t
+textvalidn(const char *s, int l) {
+  int c, utfcharbytes; /* byte count and UTF-8 codepoint length */
+
+  for (c=0; s && s[c] && (l<0 || c<l); ) {
+		utfcharbytes = 0;
+		if ((s[c] & 0x80) == 0) {
+			utfcharbytes = 1;
+		} else if ((s[c] & 0xf0) == 0xf0) {
+			utfcharbytes = 4;
+		} else if ((s[c] & 0xf0) == 0xe0) {
+			utfcharbytes = 3;
+		} else if ((s[c] & 0xe0) == 0xc0) {
+			utfcharbytes = 2;
+		} else {
+			return -1;
+		}
+
+		if ((l>0 && c + utfcharbytes >= l)) {
+			break;
+		}
+
+		c += utfcharbytes;
+  }
+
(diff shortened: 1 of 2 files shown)


jpaulogg:
Helho again!
Yesterday I was testing a few things with file name completion. I would like to share it here. I am relatively new in the programming world, so I do not know if something will be useful.

**Word completion**:
I am using `tr -s '[:blank:][:punct:]' '\n'` in vis-complete script. Your pattern considers commas, parentheses and other punctuation (and exludes underscores).

**File name completion**:

* I wanted to complete file names after a sign of equal or quotation marks. First I tried to change the  [text object](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/complete-filename.lua#L9) from long words  to short words. This works, but it does not allow longer patterns such as `"directory/pattern*"`.
So I tried to change the `gsub` function [pattern](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/complete-filename.lua#L18) on the complete-filename.lua plugin. I changed it from `gsub("'", "'\\''")` to `gsub("^.-['\"=]", "")`. This is working but I do not now if it is problematic in some way.

* I wanted to use `$HOME` as a pattern. So I made some changes to [these case statement](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/vis-complete#L41-L42) in vis-complete.
from:
```bash
		'~'|'~/'*)
			XPATTERN=$HOME$(echo $PATTERN | tail -c +2)
```
to:
```bash
                \${HOME}/*|~/*)
                        XPATTERN=${HOME}/$(basename $PATTERN)
```
To be honest, I could not understand very well what is happening in these lines of the original script. I noted the '~' case is not working.

I also noted thet when a try to open a file with `:e "~/directory with spaces/file"` vis recognize multiple files, but when I try `:e "/home/user/directory with spaces/file" it works fine. For me this is not a big issue, since I think that file name completion is more useful in text editing than for opening files, which I can easily do from the shell.

mcepl:
Already included in https://git.sr.ht/~mcepl/vis and it seems to work just fine.

mcepl (approved):
Already included in https://git.sr.ht/~mcepl/vis and it seems to work just fine.

vis-menu.c
```diff
@@ -148,8 +183,12 @@ drawtext(const char *t, size_t w, Color col) {
 	memset(buf, ' ', tw);
 	buf[tw] = '\0';
 	memcpy(buf, t, MIN(strlen(t), tw));
-	if (textw(t) > w) /* Remember textw returns the width WITH padding */
-		for (i = MAX((tw-4), 0); i < tw; i++) buf[i] = '.';
+	if (textw(t) > w) {/* Remember textw returns the width WITH padding */
+		valid = textvalidn(t, w-4);
+		if (valid < 0)
+			die("invalid UTF-8 sequence");
+		for (i = MAX(valid, 0); i < tw; i++) buf[i] = '.';
```
rnpnr, line 190: I'm not sure about calling `die()` here. It seems unnecessary.

Shugyousha:
On August 10, 2023 5:59:08 PM GMT+02:00, Randy Palamar ***@***.***> wrote:
>There are definitely still some issues with handling of non-ASCII characters
>after applying this but overall I think this improves the situation.
>
>I'm thinking of applying it apart from the call to `die()`. Any objections?

I'm also for removing the "die()" before applying. Thanks for picking up this PR!

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #142, opened 2016-01-12
Word wrap and line breaks


erf:
Option to wrap word and only line break on enter.

http://vim.wikia.com/wiki/Word_wrap_without_line_breaks

martanne:
Again simply posting a link (describing various different options)  is not that useful. What exact behavior would you like and why?

Personally I do not want that the editor interferes with line wrapping at all. If I want to wrap my text (for example for mails which should not exceed 75 columns) I use an external tool such as `fmt(1)` which is hooked up to the `=` operator. A specialized tool can also take advantage of the hyphenation rules of your language etc.

erf:
Allowing word wrapping makes it easier to read text documents. I would prefer not to figure out how to use an external tool for this, since i'm not a linux super geek.. also this seems to be a feature in most text editors.

Alternatively an option to not break lines, if you don't want to show all text on long lines of text for some reason. (Small terminal window?)

martanne:
Two solutions:

1. Break your lines at reasonable widths
2. Enlarge your terminal window

 `:set nowrap` is difficult to (efficiently) implement given the core data structures underlying vis.

PR #948, opened 2021-05-09
Soft word wrapping

proskur1n:
Hello, this pull request adds soft word wrapping (`linebreak` in vim) support to vis. In addition, it defines an option to change the visual line length instead of always using window width for the maximum line length. The latter was made trivial by my code rewrite.

#### Rationale

It is a lot of pain to write LaTeX and Markdown documents without proper word wrapping support in the editor. This option is something that makes your life much easier and is supported by nearly all other text editors. In fact, a similar functionality was requested by #142, but on the contrary to that proposal, this pull request only affects the way lines are displayed on screen and doesn't modify the underlined character data.

#### New options

- `breakat / brk` specifies a list of characters after which the editor may decide to wrap the line before window width is reached. The default value for this option is an empty string which corresponds to the old behavior of vis. In this case, cells are just printed one after another until window width is reached.
- `wrapcolumn / wc` affects the maximum displayed line length and is especially useful for large monitors. Zero is the default for this option and thus disables this feature. I didn't exactly know how to name this option, but "wrapcolumn" sounds pretty good to me.

I have included a little screenshot with both of the new options enabled below.

```
:set wrapcolumn 80
:set breakat ' .!?'
```

vis_word_wrap

#### Todo

- I have tried this feature with a couple of different text files and couldn't see any problems with my implementation. However, feel free to test this pull request on some strange edge cases, which may cause something to break. Anyway, my intention was to not change the default look of vis, but I had to rewrite the view_addch function in *view.c*, which may have caused some unintentional side effects.
- ~~Right now, `breakat` only supports ASCII characters, but it shouldn't be too difficult to extend my code to handle any UTF-8 characters. In fact, vim implementation of `breakat` also only supports ASCII.~~

### DIFF
--- a/view.c
+++ b/view.c
@@ -80,6 +80,10 @@ struct View {
 	bool need_update;   /* whether view has been redrawn */
 	bool large_file;    /* optimize for displaying large files */
 	int colorcolumn;
+	char *breakat;  /* characters which might cause a word wrap */
+	int wrapcolumn; /* wrap lines at minimum of window width and wrapcolumn (if != 0) */
+	int wrapcol;    /* used while drawing view content, column where word wrap might happen */
+	bool prevch_breakat; /* used while drawing view content, previous char is part of breakat */
 };
 
 static const SyntaxSymbol symbols_none[] = {
@@ -109,6 +113,7 @@ static bool view_viewport_up(View *view, int n);
 static bool view_viewport_down(View *view, int n);
 
 static void view_clear(View *view);
+static bool view_add_cell(View *view, const Cell *cell);
 static bool view_addch(View *view, Cell *cell);
 static void selection_free(Selection*);
 /* set/move current cursor position to a given (line, column) pair */
@@ -156,6 +161,8 @@ static void view_clear(View *view) {
 	view->bottomline->next = NULL;
 	view->line = view->topline;
 	view->col = 0;
+	view->wrapcol = 0;
+	view->prevch_breakat = false;
 	if (view->ui)
 		view->cell_blank.style = view->ui->style_get(view->ui, UI_STYLE_DEFAULT);
 }
@@ -164,24 +171,57 @@ Filerange view_viewport_get(View *view) {
 	return (Filerange){ .start = view->start, .end = view->end };
 }
 
-static bool view_add_cell(View *view, const Cell *cell) {
-	size_t lineno = view->line->lineno;
+static int view_max_text_width(const View *view) {
+	if (view->wrapcolumn > 0)
+		return MIN(view->wrapcolumn, view->width);
+	return view->width;
+}
 
-	if (view->col + cell->width > view->width) {
-		for (int i = view->col; i < view->width; i++)
-			view->line->cells[i] = view->cell_blank;
-		view->line = view->line->next;
-		view->col = 0;
+static void view_wrap_line(View *view) {
+	Line *wrapped_line = view->line;
+	int col = view->col;
+	int wrapcol = (view->wrapcol > 0) ? view->wrapcol : view->col;
+
+	view->line = view->line->next;
+	view->col = 0;
+	view->wrapcol = 0;
+
+	if (view->line) {
+		view->line->lineno = wrapped_line->lineno;
+		/* move extra cells to the next line */
+		for (int i = wrapcol; i < col; ++i) {
+			const Cell *cell = &wrapped_line->cells[i];
+			view->line->width += cell->width;
+			view->line->len += cell->len;
+			view->line->cells[view->col++] = *cell;
+		}
 	}
 
-	if (!view->line)
-		return false;
+	/* clear remaining cells on line */
+	for (int i = wrapcol; i < view->width; ++i) {
+		if (i < col) {
+			wrapped_line->width -= wrapped_line->cells[i].width;
+			wrapped_line->len -= wrapped_line->cells[i].len;
+		}
+		wrapped_line->cells[i] = view->cell_blank;
+	}
+}
+
+static bool view_add_cell(View *view, const Cell *cell) {
+	/* if the terminal is resized to a single (ASCII) char an out
+	 * of bounds write could be performed for a wide char. this can
+	 * be caught by iterating through the lines with view_wrap_line()
+	 * until no lines remain. usually 0 or 1 iterations.
+	 */
+	while (view->col + cell->width > view_max_text_width(view)) {
+		view_wrap_line(view);
+		if (!view->line)
+			return false;
+	}
 
 	view->line->width += cell->width;
 	view->line->len += cell->len;
-	view->line->lineno = lineno;
-	view->line->cells[view->col] = *cell;
-	view->col++;
+	view->line->cells[view->col++] = *cell;
 	/* set cells of a character which uses multiple columns */
 	for (int i = 1; i < cell->width; i++)
 		view->line->cells[view->col++] = cell_unused;
@@ -193,7 +233,6 @@ static bool view_expand_tab(View *view, Cell *cell) {
 
 	int displayed_width = view->tabwidth - (view->col % view->tabwidth);
 	for (int w = 0; w < displayed_width; ++w) {
-
 		int t = (w == 0) ? SYNTAX_SYMBOL_TAB : SYNTAX_SYMBOL_TAB_FILL;
 		const char *symbol = view->symbols[t]->symbol;
 		strncpy(cell->data, symbol, sizeof(cell->data) - 1);
@@ -208,22 +247,18 @@ static bool view_expand_tab(View *view, Cell *cell) {
 }
 
 static bool view_expand_newline(View *view, Cell *cell) {
+	size_t lineno = view->line->lineno;
 	const char *symbol = view->symbols[SYNTAX_SYMBOL_EOL]->symbol;
+
 	strncpy(cell->data, symbol, sizeof(cell->data) - 1);
 	cell->width = 1;
-
 	if (!view_add_cell(view, cell))
 		return false;
 
-	for (int i = view->col; i < view->width; ++i)
-		view->line->cells[i] = view->cell_blank;
-
-	size_t lineno = view->line->lineno;
-	view->line = view->line->next;
-	view->col = 0;
+	view->wrapcol = 0;
+	view_wrap_line(view);
 	if (view->line)
 		view->line->lineno = lineno + 1;
-
 	return true;
 }
 
@@ -232,9 +267,15 @@ static bool view_addch(View *view, Cell *cell) {
 	if (!view->line)
 		return false;
 
-	unsigned char ch = (unsigned char)cell->data[0];
+	bool ch_breakat = (cell->data[0] != 0) && strstr(view->breakat, cell->data);
+	if (view->prevch_breakat && !ch_breakat) {
+		/* this is a good place to wrap line if needed */
+		view->wrapcol = view->col;
+	}
+	view->prevch_breakat = ch_breakat;
 	cell->style = view->cell_blank.style;
 
+	unsigned char ch = (unsigned char)cell->data[0];
 	switch (ch) {
 	case '\t':
 		return view_expand_tab(view, cell);
@@ -493,6 +534,7 @@ void view_free(View *view) {
 		selection_free(view->selections);
 	free(view->textbuf);
 	free(view->lines);
+	free(view->breakat);
 	free(view);
 }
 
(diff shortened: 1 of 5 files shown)


mcepl:
Works for me, just it could be emphasized somewhere, that settings should go to `vis.events.WIN_OPEN` handler, not (more usual, I guess) `vis.events.INIT` one.

gvolk0:
Works great but I found one edge case:

image

The underline styling carries through the wrap-gap.  My guess is this should be easy to clear/restore while rendering the gaps, but it's not a big deal either way (I don't think your PR introduced this behaviour, but is relevant).

erf:
Anything stopping this from being merged? Seem like a good candidate

whiteinge:
I pulled this down locally and it works for me. This would be a nice addition. :+1: 

@proskur1n I'm curious if this technique could also be used to mimic `nowrap`. I often open structured data files (commonly JSON but many others as well) where line lengths are thousands of characters long but I only care to see the beginning of each line/record. (I don't need horizontal scrolling, I just want to toggle wrapping so I can see individual line numbers without the noise.)

I was hoping this PR might let me set `wrapcolumn` to a value larger than `$COLUMNS` as a hack for my use-case. That doesn't work which makes complete sense -- it's a weird idea to wrap text off-screen after all. That said, would it be possible to use the same technique in this PR to disable wrapping altogether and introduce a `nowrap` setting?

proskur1n:
@erf I marked this pull request as a draft because there were some segfaults when scrolling through binary files. Those should be fixed by aa54c376f3b9f36700a0e74d86fa871497c5e698 now. However, it may also have introduced other bugs :)

@whiteinge I don't think this is possible at the moment. As far as I know, if you simply skip some cells when rendering, the navigation may stop working properly. It's actually a miracle that soft wrapping works at all with so few changes to the code.

On a side note, I was thinking about dropping support for `breakat` and simple wrapping on `isspace(...)` instead (With a boolean `:set` option to enable it). Are there any cases when you would want to wrap the lines on something else instead?

rnpnr (changes requested):
Hi, thanks for the patch! I think this is pretty much ready to merge
there are just a few points inline.

> I was thinking about dropping support for breakat and simply wrapping on isspace(...)
> Are there any cases when you would want to wrap the lines on something else instead?

Yes, sometimes I edit Japanese text in Vis and there are no spaces in
Japanese. Instead its better to break on characters such as `。、！？`
or other such wide characters. I'm sure there are other people editing
Japanese or similar languages.

I did test this use case and it works great!

view.c
```diff
@@ -4,6 +4,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <limits.h>
+#include <assert.h>
```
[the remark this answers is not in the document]
proskur1n: I have removed them in my last commit. I don't understand your reasoning though, as asserts are useful for debugging and vis is already configured to remove them from release builds. I originally inserted those asserts there because they would have caught a weird segfault I had to debug earlier.

view.c
```diff
@@ -164,98 +172,136 @@ Filerange view_viewport_get(View *view) {
...
+	}
+}
 
-		view->line->cells[view->col] = *cell;
-		view->line->len += cell->len;
-		view->line->width += cell->width;
-		for (int i = view->col + 1; i < view->width; i++)
-			view->line->cells[i] = view->cell_blank;
+static bool view_add_cell(View *view, const Cell *cell) {
+	/* at most one iteration most of the time */
```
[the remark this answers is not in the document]
proskur1n: It should be "at most", because this loop is only executed if the line has to be wrapped. In most cases, the while-loop body is not entered at all.

view.c
```diff
@@ -164,98 +172,136 @@ Filerange view_viewport_get(View *view) {
...
-				.len = 1,
-				.width = 2,
-				.style = cell->style,
-			};
+	view->line->width += cell->width;
+	view->line->len += cell->len;
+	for (int i = 0; i < cell->width; ++i) {
+		assert(view->col < view->width);
+		if (i == 0) {
+			view->line->cells[view->col++] = *cell;
```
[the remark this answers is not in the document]
proskur1n: Yes, i think so. To my understanding, valid cells cannot have width == 0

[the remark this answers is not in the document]
proskur1n: `cell_unused` actually does have `width == 0` but it is only used inside this function. So it should be fine. Leaving this code as-is would make it a little bit more robust though (no-op for invalid and empty cells).

proskur1n:
Should we rename `breakat` to `wrapat` ? So that they are displayed next to each other in `:help`.

view.c
```diff
@@ -164,98 +171,132 @@ Filerange view_viewport_get(View *view) {
...
 		}
+		wrapped_line->cells[i] = view->cell_blank;
+	}
+}
 
-		strncpy(cell->data, view->symbols[SYNTAX_SYMBOL_EOL]->symbol, sizeof(cell->data)-1);
+static bool view_add_cell(View *view, const Cell *cell) {
+	/* at most one iteration most of the time. we have to use a while-loop
+	 * here because of some pathological edge cases where one unicode char
+	 * may be bigger than the (extremely small) terminal width. */
```
rnpnr, line 213: This still doesn't make sense to me. It can be either of "at most one
iteration" in which case the loop isn't needed OR "one iteration most
of the time" which you are suggesting is the case. But now the comment
makes it seem like a pointless endeavor because no one is using vis in
a terminal that is less than 1 (or even 10) character(s) wide.

view.c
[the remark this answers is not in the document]
rnpnr: gdb can also catch segfaults. The problem with `assert()` is that its
implemented as a macro and can have side effects when used incorrectly.
In general its usage also makes me think the author doesn't understand
the code they have written. There was nothing wrong with your usage but
I don't want to open that can of worms.

rnpnr (commented):
Sorry for being a stickler about that comment but there were a lot of
changes that seemed somewhat arbitrary in the "Fix segfault on some binary
files" commit and that is one of them. I just want to make everything
there is correct before I merge this.

rnpnr:
> Should we rename `breakat` to `wrapat` ? So that they are displayed next to each other in `:help`.

No I think `breakat` makes more sense and is consistent with other editors.

view.c
proskur1n: There are three cases:
1) Zero iterations: We are somewhere in the middle of the line. We have enough space to insert the next character and don't need to wrap the current line.
2) One iteration: Our terminal is 80 characters wide. We are inserting the 81st character. The while-loop runs once and wraps the current line. We can now insert another 80 (single width) characters until next line wrap.
3) Edge case: We have a one character wide terminal window and try to insert a double unicode. In this case the while-loop runs for every remaining line on the screen until no lines are left and we hit the `return false` branch. The editor obviously wont be able to render the text properly in this case, but it also won't crash. Calls to view_wrap_line in a loop make sure that the editor still does all the necessary housekeeping like setting the line number, replacing empty characters with empty cells and etc.

Sadly, I do not remember exactly what the problem was but it was something along those lines:

In my first version view_add_cell and view_wrap_line were implemented as mutually recursive functions. It didn't cause any problems in the most cases. However, sometimes after a line wrap there still wasn't enough space to insert the next character and the two functions were calling each other indefinitely. In aa54c376f3b9f36700a0e74d86fa871497c5e698 I have removed the recursion entirely and added a while-loop to account for the edge case of an extremely small terminal window.

proskur1n: On second thought, I think it wasn't really the recursion itself causing the segfault, but rather a null pointer somewhere as a result of the mutual recursion. Anyway, no recursion -> no problems :)

rnpnr: > recursion

Ahhh, I see now. I didn't catch that that was what you removed. If thats
the case it was probably running out of stack space. Let me think about
how to reword it a bit more and I will update the comment it before
merging.

Also that probably means that it wasn't an issue with your patch and
actually a preexisting issue. I'll make sure to keep it as a separate commit.

rnpnr (approved):
Everything looks good now. Give me a little bit to squash some commits
and reword some comments and I'll merge it.

Also I was just doing some quick research and it seems like not even
neovim supports arbitrary UTF-8 in the `breakat` option so great work!

view.c
proskur1n: > Also that probably means that it wasn't an issue with your patch
No, it was actually a problem I introduced in one of the first commits in this PR and not an preexisting issue. Both functions don't exist in master at all.

Feel free to reword my comments.

rnpnr: Oh the `view_add_cell()` and `view_wrap_line()` mutually calling each
other. Maybe its ok to squash that then. Well I'll make sure whatever
I do makes sense before merging.

rnpnr: So I did some more testing here and I'm not really able to recreate
the segfault even with a 1 ASCII char wide terminal and a file with
multiwide chars. I'm thinking we should just switch the `while` to an
`if` and not worry about the hypothetical edge case.

proskur1n: It doesn't segfault anymore after the recursion got removed, but there is still an out-of-bounds write without the while-loop in `view->line->cells[view->col++] = cell_unused;`

We cannot replace `while` with `if` here.

proskur1n: This doesn't fix the out-of-bounds write. When the terminal window is resized to be one character wide then `view->line->cells` will only have space for a single character.

This line will not produce an out-of-bounds write. `view->col` will be 0 after a resize. So this write is valid.
```
if (cell->width > 0)
	view->line->cells[view->col++] = *cell;
```

This for-loop is the problem in this case. If `cell->width` is for example two then we will write to `view->line->cells[1]` which is out-of-bounds for an array with only 1 element.
```
for (int i = 1; i < cell->width; i++)
	view->line->cells[view->col++] = cell_unused;
```

We should just keep the while-loop instead of removing it for doubtful gains and introducing new bugs in the process...

rnpnr: Ok then the comment should explain this. Right now the code doesn't
make sense and the comment doesn't give any insight into what is going
on and just leads to confusion.

How about this:
```diff
 static bool view_add_cell(View *view, const Cell *cell) {
-	/* at most one iteration most of the time. we have to use a while-loop
-	 * here because of some pathological edge cases where one unicode char
-	 * may be bigger than the (extremely small) terminal width. */
+	/* if the terminal is resized to a single (ASCII) char an out
+	 * of bounds write could be performed for a multiwide char.
+	 * this can be caught by iterating through the lines with
+	 * view_wrap_line() until no lines remain.
+	 */
 	while (view->col + cell->width > view_max_text_width(view)) {
 		view_wrap_line(view);
 		if (!view->line)
```

proskur1n: I am fine with that :+1: . I would just mention that "at most one iteration most of the time". It would make it clear that this is an inexpensive loop.

rnpnr:
The two commits I just pushed are what I will merge shortly. Below is
the diff between what I pushed and what was here before (for posterity):
```diff
diff --git a/view.c b/view.c
index ebdb876..0b7ce36 100644
--- a/view.c
+++ b/view.c
@@ -178,9 +178,9 @@ static int view_max_text_width(const View *view) {
 }
 
 static void view_wrap_line(View *view) {
+	Line *wrapped_line = view->line;
 	int col = view->col;
 	int wrapcol = (view->wrapcol > 0) ? view->wrapcol : view->col;
-	Line *wrapped_line = view->line;
 
 	view->line = view->line->next;
 	view->col = 0;
@@ -197,7 +197,7 @@ static void view_wrap_line(View *view) {
 		}
 	}
 
-	/* clear remaining of line */
+	/* clear remaining cells on line */
 	for (int i = wrapcol; i < view->width; ++i) {
 		if (i < col) {
 			wrapped_line->width -= wrapped_line->cells[i].width;
@@ -208,9 +208,11 @@ static void view_wrap_line(View *view) {
 }
 
 static bool view_add_cell(View *view, const Cell *cell) {
-	/* at most one iteration most of the time. we have to use a while-loop
-	 * here because of some pathological edge cases where one unicode char
-	 * may be bigger than the (extremely small) terminal width. */
+	/* if the terminal is resized to a single (ASCII) char an out
+	 * of bounds write could be performed for a wide char. this can
+	 * be caught by iterating through the lines with view_wrap_line()
+	 * until no lines remain. usually 0 or 1 iterations.
+	 */
 	while (view->col + cell->width > view_max_text_width(view)) {
 		view_wrap_line(view);
 		if (!view->line)
@@ -220,11 +222,9 @@ static bool view_add_cell(View *view, const Cell *cell) {
 	view->line->width += cell->width;
 	view->line->len += cell->len;
 	view->line->cells[view->col++] = *cell;
-	for (int i = 1; i < cell->width; ++i) {
-		/* set cells of a character which uses multiple columns */
+	/* set cells of a character which uses multiple columns */
+	for (int i = 1; i < cell->width; i++)
 		view->line->cells[view->col++] = cell_unused;
-	}
-
 	return true;
 }
```

Issue #949, opened 2021-05-20
INSERT mode autocomplete by <C-n> only completes the primary selection 


sauterp:
**Version**
commit: 1a958f221404b09cb8b0612fb34301e6b9783cf9 (latest master at this time)
`vis -v
vis v0.7-29-g1a958f2 +curses +lua +acl`

**How to reproduce**

1. Open vis.
2. Type any word like "hello".
3. Create more than one selection(<C-j>).
4. Type the word partially like "he".
5. Hit <C-n> while in insert mode.
6. Hit enter when "hello" is selected.

Only the primary selection will be completed. It's not a big problem, but I would find it more convenient and consistent with other features if the autocompletion would apply to all selections, like pasting. Unlike pasting, I wouldn't have a different autocomplete suggestion for each selection. That's not feasible I think. I would implement it such that the suggestions are generated based on the primary selections cursor. What do you think?

ninewise:
I've seen this as well but learned to live with it. It'd be nice to get the same (primary) completion on every selection though.

jpaulogg:
The [Vis:insert](http://martanne.github.io/vis/doc/#Vis:insert) insert keys at all cursor positions of active window. I coppied some comde from [digrpah.lua](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/digraph.lua#L11-L14) and it is working with multipl selections and also in replace mode.

Just replace [this line](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/complete-word.lua#L20) with that:

```lua
	if vis.mode == vis.modes.INSERT then
		vis:insert(out)
	elseif vis.mode == vis.modes.REPLACE then
		vis:replace(out)
	end
```

sauterp:
> The [Vis:insert](http://martanne.github.io/vis/doc/#Vis:insert) function inserts keys at all cursor positions of active window. I copied few lines from [digrpah.lua](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/digraph.lua#L11-L14) and it is working with multiple selections and also in replace mode.
> 
> Just replace [this line](https://github.com/martanne/vis/blob/1a958f221404b09cb8b0612fb34301e6b9783cf9/lua/plugins/complete-word.lua#L20) with that:
> 
> ```lua
[...]

PR #950, opened 2021-05-20
Autocomplete all selections with <C-n>

sauterp:
fixes https://github.com/martanne/vis/issues/949

Autocomplete in INSERT mode by <C-n> completes all selections based on
the suggestions generated by the input at the primary selection's cursor.

### DIFF
--- a/lua/plugins/complete-word.lua
+++ b/lua/plugins/complete-word.lua
@@ -2,17 +2,25 @@
 -- all non-primary selections will be completed aswell if their cursor
 -- is located at the end of the same word
 
+local get_prefix = function (file, pos)
+		if not pos then return "", false end
+
+		local range = file:text_object_word(pos > 0 and pos-1 or pos);
+		if not range then return "", false end
+		if range.finish > pos then range.finish = pos end
+		if range.start == range.finish then return "", false end
+		local prefix = file:content(range)
+		if not prefix then return "", false end
+		return prefix, true
+end
+
 vis:map(vis.modes.INSERT, "<C-n>", function()
 	local win = vis.win
 	local file = win.file
 	local pos = win.selection.pos
-	if not pos then return end
-	local range = file:text_object_word(pos > 0 and pos-1 or pos);
-	if not range then return end
-	if range.finish > pos then range.finish = pos end
-	if range.start == range.finish then return end
-	local prefix = file:content(range)
-	if not prefix then return end
+	prefix, ok = get_prefix(file, pos)
+	if not ok then return end
+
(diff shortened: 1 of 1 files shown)


Shugyousha:
Refactoring out the common parts like this may be nicer:

```
diff --git a/lua/plugins/complete-word.lua b/lua/plugins/complete-word.lua
index f0a6ff5..737cabe 100644
--- a/lua/plugins/complete-word.lua
+++ b/lua/plugins/complete-word.lua
@@ -2,17 +2,25 @@
 -- all non-primary selections will be completed aswell if their cursor
 -- is located at the end of the same word
 
+local get_prefix = function (file, pos)
+		if not pos then return "", false end
+
+		local range = file:text_object_word(pos > 0 and pos-1 or pos);
+		if not range then return "", false end
+		if range.finish > pos then range.finish = pos end
+		if range.start == range.finish then return "", false end
+		local prefix = file:content(range)
+		if not prefix then return "", false end
+		return prefix, true
+end
+
 vis:map(vis.modes.INSERT, "<C-n>", function()
 	local win = vis.win
 	local file = win.file
 	local pos = win.selection.pos
-	if not pos then return end
-	local range = file:text_object_word(pos > 0 and pos-1 or pos);
-	if not range then return end
-	if range.finish > pos then range.finish = pos end
-	if range.start == range.finish then return end
-	local prefix = file:content(range)
-	if not prefix then return end
+	prefix, ok = get_prefix(file, pos)
+	if not ok then return end
+
 	local cmd = string.format("vis-complete --word '%s'", prefix:gsub("'", "'\\''"))
 	local status, out, err = vis:pipe(file, { start = 0, finish = file.size }, cmd)
 	if status ~= 0 or not out then
@@ -21,14 +29,8 @@ vis:map(vis.modes.INSERT, "<C-n>", function()
 	end
 	for selection in win:selections_iterator() do
 		local pos = selection.pos
-		if not pos then goto continue end
-
-		local range = file:text_object_word(pos > 0 and pos-1 or pos);
-		if not range then goto continue end
-		if range.finish > pos then range.finish = pos end
-		if range.start == range.finish then goto continue end
-		local localprefix = file:content(range)
-		if not localprefix then goto continue end
+		localprefix, ok = get_prefix(file, pos)
+		if not ok then goto continue end
 		if localprefix ~= prefix then goto continue end
 
 		file:insert(pos, out)
```

Other than that it looks good to me (with very basic testing)!

Shugyousha (approved):
I hope the way to define the helper function is somewhat Lua-style conform (I am not very familiar with the language). This change looks good to me in any case!

Issue #952, opened 2021-06-04
It's not apparent how highlighting matching pairs works in theme definitions


hyphenrf:
I've gone through the theming wiki page and default theme to try and extract this bit of info but failed without actually experimenting with combinations of theme definitions myself. Would be useful to mention it or at least leave a comment about it on default theme.  
As I mentioned, I've only figured it out with experimentation: `STYLE_DEFAULT` foreground color needs to be set for the matching pairs effect to happen.. Which seems like a strange dependency.

mcepl:
Is this problem with the implementation of https://github.com/orbitalquark/scintillua or an upstream problem? The latter probably should be closed in preference of the scintillua issue tracker.

Issue #927, opened 2021-01-27
Allow different cursor shapes


StavromulaBeta:
I have configured my terminal emulator to use an underline cursor instead of the usual block cursor. One thing that I have noticed is that this does not work in Vis - the cursor simply appears as a block until I exit the editor. I have tested this in Urxvt and Xterm.

 Is there any way that we can have a configuration option to change the cursor, or perhaps an option to use the terminal's default cursor?

3dc1d3:
dup: https://github.com/martanne/vis/issues/781#issuecomment-568176414

Issue #911, opened 2020-12-11
Setting `fore` for vis.lexers.STYLE_CURSOR does nothing


Yodzorah:
# Problem

```lua
vis.lexers.STYLE_CURSOR = 'back:#cccc00,fore:#000000' -- back: acid yellow, fore: black
```

Background changed, foreground stays light. Which means I can't manually set a light background for the cursor without making it impossible to read the symbol under it.

# Expected behaviour

Ability to set `fore` just like with `back`.

Yodzorah:
Sorry. Turns out I confused STYLE_CURSOR and STYLE_CURSOR_PRIMARY. Foreground works.

Issue #462, opened 2017-01-16
change cursor style on focus


aeosynth:
i'd like the cursor to display a block when the containing terminal is focused, and a 'hollow block' when unfocused. can this be scripted with lua?

lafrenierejm:
Where is the cursor shape defined?

josuah:
Cursor shape control is made through [escape sequences](https://gist.github.com/anonymous/8f1e6c22b5213faf8170dcfc2b0f5b93).  There is no API needed, only printing some ANSI characters.

Vis has multiple cursors, and terminals only support 1.  So what vis does is totally disabling the terminal cursor by setting an invisible cursor, and sets some characters to be in reverse video wherever a cursor should be displayed, so vis has no "real" cursor, only reversed blocks here and there.

If ther is any C source to look at, it may be in ui-*.c.

How to debug escape sequence:

``` bash
$ script
$ vis
Editing with vis for not too long as there will be a lot of output.
$ less script
```

ber532k:
Is vis able to detect a lose-of-focus? I frequently work with multiple tmux-panes side by side, quite often having vis run in more than one and it is quite irritating to have two cursors which give no indication as to where the focus lies. Simply removing the cursor-shaped block when vis looses focus would do it for me. Or maybe the cursor-shape(s) could be replaced by some other symbol.

Also: Is it possible to locate the invisible cursor at the primary cursor's position? Or would that have any negative consequences? It would make it more comfortable to work with tmux's yanking functionality. Currently tmux always locates its own cursor for yanking text at the bottom left corner of the pane which – after reading this – it appears to me might be where the invisible cursor actually resides.

PR #953, opened 2021-06-11
use terminal cursor

gotroyb127:
Related issues: #927, #462, #952.

With this, the primary cursor is not drawn via styling the cell it is positioned at, but instead it has the terminal draw it. Consequently, the `STYLE_CURSOR_PRIMARY` is removed, as it is now unused. Also, I added `STYLE_CURSOR_MATCHING` to specify the style for the matching pair of a ``(){}[]`"'``, which is the same for primary and secondary cursors (it is trivial to change it so that one can specify different style for those).

About the implementation:
I have removed all ocurances of `curs_set(.)` and `cursor_visible(.)` in `ui-terminal-{curses,vt100}.c` respectively, which I am unsure if it is the way you would want it.

Also, I just renamed `STYLE_CURSOR_PRIMARY` to `STYLE_CURSOR_MATCHING` in all files (`lua/themes/*.lua` included) to make clear that it is not used anymore and also to have the new option appear somewhere.

Something else to notice is that I tried to prevent drawing of the primary selection when it is only the cell under the cursor so that if one changes the cursor style via terminal-specific escape sequences, such as `\e[6 q`, the underlying cell is not styled specially.

I expect your comments so I can adjust this to something that might be worth merging.

Cheers.

Edit: also related: #911.


MaxGyver83:
Awesome! Thank you for this.
I was trying to figure out how I can get different cursor styles for insert mode and replace mode before I found your pull request.

I change cursor styles like that (on top of your pull request):

```diff
diff --git a/vis-modes.c b/vis-modes.c
index 2e120aa..8dd2803 100644
--- a/vis-modes.c
+++ b/vis-modes.c
@@ -146,6 +146,7 @@ static void vis_mode_normal_enter(Vis *vis, Mode *old) {
                return;
        if (old != mode_get(vis, VIS_MODE_INSERT) && old != mode_get(vis, VIS_MODE_REPLACE))
                return;
+       printf("\e[2 q\n"); /* block cursor */
        if (vis->autoindent && strcmp(vis->key_prev, "<Enter>") == 0) {
                Text *txt = win->file->text;
                for (Selection *s = view_selections(win->view); s; s = view_selections_next(s)) {
@@ -229,6 +230,10 @@ static void vis_mode_visual_leave(Vis *vis, Mode *new) {
 static void vis_mode_insert_replace_enter(Vis *vis, Mode *old) {
        if (!vis->win || vis->win->parent)
                return;
+       if (vis->mode->id == VIS_MODE_REPLACE)
+               printf("\e[4 q\n"); /* underline cursor */
+       else
+               printf("\e[6 q\n"); /* bar cursor */
        if (!vis->action.op) {
                action_reset(&vis->action_prev);
                vis->action_prev.op = &vis_operators[VIS_OP_MODESWITCH];
```

Not sure if this is the correct way to do it.

erf:
Could this perhaps be made optional via lua config? Either use terminal cursor or configure via style.

MaxGyver83:
> Could this perhaps be made optional via lua config? Either use terminal cursor or configure via style.

This would be cool!

> Not sure if this is the correct way to do it.

I think it's not the correct way to do it. I get weird behavior when I enter this after starting vis: `Sabc<Esc>.`

Peek 2022-07-10 12-11

MaxGyver83:
> I think it's not the correct way to do it (just talking about my patch on top of this pull request). I get weird behavior when I enter this after starting vis: `Sabc<Esc>.`

It seems to be working with that patch:

```diff
diff --git a/vis-modes.c b/vis-modes.c
index 2e120aa..d414a2b 100644
--- a/vis-modes.c
+++ b/vis-modes.c
@@ -146,6 +146,8 @@ static void vis_mode_normal_enter(Vis *vis, Mode *old) {
                return;
        if (old != mode_get(vis, VIS_MODE_INSERT) && old != mode_get(vis, VIS_MODE_REPLACE))
                return;
+       printf("\e[2 q"); /* block cursor */
+       fflush(stdout);
        if (vis->autoindent && strcmp(vis->key_prev, "<Enter>") == 0) {
                Text *txt = win->file->text;
                for (Selection *s = view_selections(win->view); s; s = view_selections_next(s)) {
@@ -229,6 +231,11 @@ static void vis_mode_visual_leave(Vis *vis, Mode *new) {
 static void vis_mode_insert_replace_enter(Vis *vis, Mode *old) {
        if (!vis->win || vis->win->parent)
                return;
+       if (vis->mode->id == VIS_MODE_REPLACE)
+               printf("\e[4 q"); /* underline cursor */
+       else
+               printf("\e[6 q"); /* bar cursor */
+       fflush(stdout);
        if (!vis->action.op) {
                action_reset(&vis->action_prev);
                vis->action_prev.op = &vis_operators[VIS_OP_MODESWITCH];
```

Sorry for spamming this pull request. But maybe it's better than no answer at all :-)

gotroyb127:
Hello.

Maybe comments are more related to issue #927, but it seems worth mentioning
that it is possible (I'd guess preferable also) to handle cursor shapes from lua
scripts.

I had been using this for a long time (I prefer the block shape nowadays :D):
````
local lastMode
function updateStatus(win)
	if lastMode ~= vis.mode then
		lastMode = vis.mode
		local n
		if lastMode == vis.modes.NORMAL then
			n = 2
		elseif lastMode == vis.modes.OPERATOR_PENDING then
			n = 4
		elseif lastMode == vis.modes.INSERT then
			n = 4
		elseif lastMode == vis.modes.REPLACE then
			n = 4
		elseif lastMode == vis.modes.VISUAL then
			n = 2
		elseif lastMode == vis.modes.VISUAL_LINE then
			n = 2
		end
		io.stderr:write(strf('\x1b[%d q', n))
	end
	-- status drawing/etc...
end

vis.events.subscribe(vis.events.WIN_STATUS, updateStatus)
````

I guess having an event which is triggered when mode gets changed would be more
straightforward, but this, checking for changes when drawing the status line,
works nicely nevertheless.

Notice that it writes the escape sequences to stderr, which seems to be where
vis writes its drawing (try `vis >/dev/null`).

Cheers.

MaxGyver83:
Cool! That's good to know. I hope your PR will be accepted someday.

hholst80:
> strf

How do I get access to the 'strf' function?

gotroyb127:
> > strf
> 
> How do I get access to the 'strf' function?

`strf` is being used as an alias for `string.format`: `local strf = string.format` (https://github.com/gotroyb127/dotfiles/blob/5d72589137d0218ae8990fe8e9185c02b62f70ce/vis/visUtil.lua#L4)

MaxGyver83:
@martanne, @ninewise , @rnpnr: Is anything wrong with this pull request? Could you please review it and tell if anything is missing or if this feature is unwelcome?

rnpnr (approved):
Hi Petros,

I've tested it with and without curses support enabled and
everything seems fine to me. Personally I see no point in making this
configurable. Also the `STYLE_CURSOR_MATCHING` addition is very welcome.

Thanks for the patch! I will merge it soon.

rnpnr:
@gotroyb127
Actually before I merge it can you give a better description for the
'move cursor to the expected location' commit. Is it for drawing `|`
at the start of the cell instead of the end?

gotroyb127:
Sorry for taking so long to reply.

> Actually before I merge it can you give a better description for the
> 'move cursor to the expected location' commit. Is it for drawing `|`
> at the start of the cell instead of the end?

Without `'move cursor to the expected location'` commit, `vis` instructs the
terminal to not draw its cursor and `vis` makes it appear like there is a cursor
by using appropriate foreground and background colors of the would-be cursor cell.

With the `'move cursor to the expected location'` commit, the terminal's cursor is
no longer hidden and is repositioned on the expected location -- and `vis`'s primary
cursor is no longer drawn.

I don't understand what you mean by drawing `|`.

Cheers.

rnpnr (changes requested):
Hi, so after some more testing I noticed that this seems to break the highlighting of a matching pair when vis is used in a tty. I pushed this branch with the rebased and squashed changes to work off of. I can potentially look into it more if I find some time.

> I don't understand what you mean by drawing `|`.

Sorry I said that because of an added comment in the code. The confusion about the cursor was because it wasn't clear if the commit message was referring to the terminal's cursor or vis' cursor. I understand now, thanks.

mcepl:
Also, this doesn’t apply cleanly (or rather not at all) on the current master branch.

@gotroyb127 Could we ask for the rebase, please?

gotroyb127:
I've reimplemented it (with some improvements) based on current master, and overwritten the previous
branch (I just force-pushed, I don't have any experience with rebase-ing but I'm guessing that'll be enough).

What's different with this new patch is that `STYLE_CURSOR_MATCHING` is added but does not replace `STYLE_CURSOR_PRIMARY` (and thus fixes #952).

`STYLE_CURSOR_PRIMARY` can be set to an empty string (from lua configuration) to not having it styled
specially. This is necessary in cases like #927.

Cheers.

mcepl:
@rnpnr Could you consider and either merge or reject this PR before #1191? This seems to conflict.

rnpnr:
> @rnpnr Could you consider and either merge or reject this PR before #1191? This seems to conflict.

I like this and want to merge it but I think there is a remaining issue. When you do some actions such as `<C-n>` to complete word the cursor remains in place instead of moving to the bottom of the screen where its supposed to. I find that this is a little hard to read because there is no separation between vis-menu and the rest of the editor.

gotroyb127:
Hi.

> I like this and want to merge it but I think there is a remaining issue. When you do some actions such as `<C-n>` to complete word the cursor remains in place instead of moving to the bottom of the screen where its supposed to.

This is, indeed, not the intended behavior, and I am unsure why this happens. In `vis-complete`, `vis-menu` is called with the `-b` option so it should use the last line of the terminal, independently of where the cursor is currently positioned at.

neuromagus:
So, anybody! Who fix this pull request?
For me this is more needed, than other plugins. IMO - ```vis-complete``` or ```vis-menu``` need fix after change core. This pull request change core? Am i right?
Because too complicated look on ugly bottom for check, what is the mode now.

Sorry for my eng.

mcepl:
> So, anybody! Who fix this pull request? For me this is more needed, than other plugins. IMO - `vis-complete` or `vis-menu` need fix after change core. This pull request change core? Am i right? Because too complicated look on ugly bottom for check, what is the mode now.

@gotroyb127 should, it is his idea. And unless, he rebases on the top of the current master branch, I won’t be even looking at it. The good news is that I am now around with right to merge, so after he makes @rnpnr happy, I suppose, I am willing to merge.

> Sorry for my eng.

Don’t apologise, most of us have English as second language here.

acidrums4:
Sorry for bothering, I was wondering if this would make you able to have a blinking cursor in vis if you have set it to in your terminal emulator? Currently I have set it, but in vis it won't blink

MaxGyver83:
@gotroyb127, are you still around? Could you please try to rebase this pull request?

tassmjau:
I wanted this functionality so I re-implemented @gotroyb127 changes manually on top of master.

[My branch "cursor"](https://github.com/martanne/vis/compare/master...tassmjau:vis:cursor) includes this as well as fixes for the `<C-n>` autocomplete issue mentioned by @rnpnr.
When trying with the pipe/bar/unicode snoman cursors in `st(1)` and in a tty I realized some more tweaks were needed to make vis always render the "hardware" cursor (when possible), so that's also in there.

Don't want to create another PR side-by-side to this since @gotryb127 did the heavy lifting, but thought I should mention it if someone is inclined to use it (or if it can be helpful when @gotroyb127 rebases this).

Doodler8888:
@tassmjau i probably found a bug in your patch. if i execute a nonexistent command from a command line, the cursor will disappear. 

https://github.com/user-attachments/assets/a9e75a9d-e894-42b4-85cc-2f564a3103d9

tassmjau:
Aha, thanks for trying it out and reporting that @Doodler8888, I found the cause and pushed a fix if you're still interested!
(edit: I understand this is OT for this PR, sorry about that)

rnpnr:
Closing as this is superseded by #1372. Thanks for the initial work!

PR #957, opened 2021-06-15
build: update alpine in docker build to version 3.14

eworm-de:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Run 'make docker' to build a statically linked vis executable!
-FROM i386/alpine:3.13
```
mcepl, line 2: Wouldn’t it be better to have here `alpine:latest`?

mcepl (changes requested):
Looks completely trivial.

Dockerfile
eworm-de: Not sure... I think the idea here was to have reproducibility. However packages are updated within one branch as well.

Also we had breaking changes with alpine updates, see 0ba252f08e9ef7b62c442eb91b36822d654f5b04 for example. With a commit for every version an issue shows up in bisecting.

mcepl: :+1: then, but it should be the Biblical 3.16 now, shouldn't it?

mcepl (approved):
It wouldn't hurt, right?

Dockerfile
eworm-de: Last commit is for 3.16.
Let's keep everything between, again for bisecting.

deepcube:
On Wed Jun 1, 2022 at 9:32 AM MDT, Christian Hesse wrote:
> Well, you squashed it... So no bisecting between apline versions possible...
> Thanks anyway!

Sorry about that, I missed the comments about bisecting.  Next time!

eworm-de:
I am not aware of any breakage, so let's hope this is fine. 😜

PR #959, opened 2021-07-02
Simpler and more flexible insert completion

jpaulogg:
This PR improves vis insert completion in the following ways:
* allow the shell to expand filename prefixes. This way, tilda and environment variables can be used properly;
* simpler rewriting of vis-completion;
* escape spaces occurrences in vis-open output;
* allow to complete text objects other than short words (for example, functions names in an external file).

More specific details can be found in the commit messages.

PS: I wrote a [plugin](https://github.com/jpaulogg/vis-ins-completion/tree/flex-completion) that benefits from these changes.


mcepl:
@jpaulogg Why did you close this PR? It seems to make sense?

mcepl:

PR #963, opened 2021-07-14
reword the :[v]split and :[v]new descriptions in vis.1

rnpnr:
I've always found this wording a bit confusing.  I think the way I understand it in my mind is opposite to the current wording. By switching to rows and columns it should clear up the confusion.

(horizontally) and (vertically) were kept to be consistent with the 'v'
mnemonic and with the enum labels in the code.

### DIFF
--- a/.github/workflows/man.yml
+++ b/.github/workflows/man.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Manual generation
       run: |
         make man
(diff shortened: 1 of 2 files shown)


mcepl (approved):
LGTM

mcepl:
Merged into https://git.sr.ht/~mcepl/vis/

PR #965, opened 2021-07-22
lexers/html: detect script type "module"

benhormann:
In theory it could be `<script type=module>` (without quotes), hence `(["\']?)`.
Edit: Hmm, `(["\']?)` doesn't seem to work (for unquoted, quoted is fine).

So maybe something like this would be better:
```lua
  local _, _, _, script_type = input:find('^%s+type%s*=%s*(["\'])(^["\']+)%1', index)
  if script_type == 'text/javascript' or script_type == 'module' then
```


mcepl:
Lexers are not maintained as part of vis any more, please, push this ticket to https://github.com/orbitalquark/scintillua/issues. Thank you.

PR #977, opened 2021-08-18
lexers/makefile: Support multiple targets in a single definition

nthirtyone:
Multiple targets are supported by POSIX standard and most commonly used make implementations. I was mildly annoyed by them not being coloured so I fixed it and decided to share it.
This does *not* introduce support for [grouped targets](https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html#Rules-with-Grouped-Targets). If that's desirable, it should be straight-forward to do.


nthirtyone:
@deepcube, There you go.

deepcube:
Applied. Thanks.

PR #983, opened 2021-09-25
filetype: Set `diff` filetype for `COMMIT_EDITMSG` files

krkk:
This filename is set when writing a commit message in git.

If you run git-commit with an `--verbose` option or if you just have set it in your git-config, that shows you the patch diff on bottom of the file.

Unfortunately, now every patch file will set colorcolumn, but overall I think this is better than having duplicated lexers.


mcepl (approved):
Looks exactly correct.

ninewise:
Applied 0cc684f26b136029f4f49d808761d69228f42dff

PR #984, opened 2021-09-27
Redirect stderr of wl-copy to /dev/null

### LINKED ISSUE #929: Can't copy with wl-copy
Using wl-copy results in a weird behavior where it keeps receiving text through the command prompt.
Tested with neovim and acme, it works as expected.
wl-paste works.

vis: v0.7
wlroots: v0.12
sway: v1.15

2021-02-10T14:22:15,566869047-03:00

andreafeletto:
Fixes #929.
This of course ignores possible error messages from wl-copy.
Still, the system clipboard on wayland does not work without this change.

### DIFF
--- a/vis-clipboard
+++ b/vis-clipboard
@@ -81,9 +81,9 @@ vc_paste() {
 
 vc_wlclipboard_copy() {
 	if [ "$sel" = "primary" ]; then
-		wl-copy --primary -t TEXT
+		wl-copy --primary -t TEXT 2>/dev/null
 	else
-		wl-copy -t TEXT
+		wl-copy -t TEXT 2>/dev/null
 	fi
 }
 


mcepl:
> Still, the system clipboard on wayland does not work without this change.

I don’t understand. I have here Wayland-only sway, with vis-0.7+git.1618946717.1a958f2 and wl-clipboard-2.0.0 and I don’t observe any problem.

mcepl (changes requested):
> Still, the system clipboard on wayland does not work without this change.

I don’t understand. I have here Wayland-only sway, with vis-0.7+git.1618946717.1a958f2 and wl-clipboard-2.0.0 and I don’t observe any problem.

PR #991, opened 2021-12-23
Added multilines strings to Zig’s syntax definition

sansfontieres:


Nomarian:
I think this goes in the scintilua project instead.

ninewise:
Yes, lexers are taken from https://github.com/orbitalquark/scintillua, please submit the patch there.

PR #997, opened 2022-01-25
Detect filetype via hashbang

Nomarian:


Nomarian:
I did not, no...


There is still merit in just having a hashbang table, at least for readability I guess (no functions, just a table of strings) I added only 2, but a lot of languages have a pretty well known and unique interpreter in PATH, adding more is rather trivial with a table. and the problem is a bit more complicated than just a simple match, IMO you should also ignore /usr/bin/env.


but its up to yall.

deepcube:
I do like the idea of checking the shebang, and I'm leaning towards accepting this. A few questions/comments:

- Does this offer significantly more functionality than calling file for the mime type?  Is it just to support shebangs that aren't already included in the file command?

- What happens if there's a flag?  e.g.   #!/bin/sh -xe    or    #!/usr/bin/awk -f    It looks like the provided string would have to omit the $ that you used in your example to signify end of string, is that correct?  I wonder if it would be better to only match against the first word, or to leave that open to the user to specify whether or not to care about it.  Can you think of any use case where someone would care about that flag when matching?

- Can you update it to accept a space between the ! and / as that's fairly common e.g.   #! /bin/sh   (if it does already support this and I missed it let me know)

Nomarian:
> Does this offer significantly more functionality than calling file for the mime type? Is it just to support shebangs that aren't already included in the file command?


Well, that's entirely dependent on the /bin/file database/version. It doesn't have to be either this or that. it can try the mime and if it fails it can go for the internal hashbang check. executing file to me is an unnecessary fork. (But loops and shellscript has left me slightly paranoid about them.)


> What happens if there's a flag? e.g. #!/bin/sh -xe or #!/usr/bin/awk -f It looks like the provided string would have to omit the $ that you used in your example to signify end of string, is that correct? I wonder if it would be better to only match against the first word, or to leave that open to the user to specify whether or not to care about it.


The flag gets ignored, it only cares about the last FILENAME of #!/FILEPATH/FILENAME or the first argument to /usr/bin/env (yes, -S is valid because that would require parsing CLI args and identifying where the command arg for env is, which is slightly complex but can be added in) there is a caveat or a bug, which is a false positive because it just wants the last filename (or whatever looks like a filepath)

> Can you think of any use case where someone would care about that flag when matching?


Yes, bash has a --posix flag which mimics sh behavior, zsh also has one. The fix is probably to use the detect function instead, since CLI args can be incredibly complicated.


> Can you update it to accept a space between the ! and / as that's fairly common e.g. #! /bin/sh (if it does already support this and I missed it let me know)


Sure, I haven't programmed in months and I've forgotten everything. Now that I look at it, line 520 and 522 should be changed as well.

lua/plugins/filetype.lua
```diff
@@ -506,6 +508,29 @@ vis.events.subscribe(vis.events.WIN_OPEN, function(win)
...
+
+		-- detect filetype by hashbang
+		local f = io.open(name)
+		if
+			f and f:read(3)=="#!/" and f:read"*l"~="" -- file exists, and has #!/file
+			or
+			f and not f:close() -- close the file if anything fails
+		then
+			f:seek"set"
+			local fullhb = f:read"*l":match"^#!(.+)$"
```
Nomarian, line 520: change to

			local fullhb = f:read"*l":match"^#!%s*(%g+.*)$"

Nomarian:
Updated to use a space. will also not open the file and use the data variable that ft.detect uses. Also added a utility table which matches only the utility directly and the hashbang table matches the entire hashbang (but not the #!). 


Please test.

deepcube:
Thanks for the update! I'll review now. In the meantime can you squash the commits?

deepcube:
Testing now. Can you remove the extra white space on blank lines you added?

PR #1000, opened 2022-04-18
Re-enabled tre by #including <stddef.h>.

somercet:
I noticed libtre.so.5 was not linked to the vis binary; this fixes the test code in ./configure. I think this was caused by compilation on Void Linux's musl libc arch.

Love,

p

### DIFF
--- a/configure
+++ b/configure
@@ -380,6 +380,7 @@ if test "$tre" != "no" ; then
 	printf "checking for libtre... "
 
 cat > "$tmpc" <<EOF
+#include <stddef.h>
 #include <tre/tre.h>
 
 int main() {


mcepl:
Is this just an musl issue? With my Linux I seem to have `libtre.so.5` linked just fine:
```
stitny~$ ldd /usr/bin/vis|grep tre
    libtre.so.5 => /lib64/libtre.so.5 (0x00007f21845ae000)
stitny~$
```

somercet:
I think it is just a musl (or gcc behavior under musl) thing. But it won't break normal builds, it's just auto-included on glibc arches.

ninewise:
Could verify the issue (and fix) on Void. Merging. Thanks for the contribution!

somercet:
Your welcome!

PR #1009, opened 2022-05-28
Fix Python MIME for filetype detection

ilyagr:
On my machine (Linux, file 5.39), the MIME type `file` returns for Python files is `text/x-script.python`, not `text/x-python`.


ilyagr:
I think my system's magic file is from https://github.com/file/file/blob/a6de062b431e1fd1b69bd07d99d157bf73dff460/magic/Magdir/python#L266.

ninewise:
Duplicate of a sourcehut patch, now applied 336f66c

PR #1011, opened 2022-06-08
Add support for the Hare language file type

qiu-x:
https://harelang.org

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -178,6 +178,9 @@ vis.ftdetect.filetypes = {
 	gtkrc = {
 		ext = { "%.gtkrc$" },
 	},
+	hare = {
+		ext = { "%.ha$" }
+	},
 	haskell = {
 		ext = { "%.hs$" },
 		mime = { "text/x-haskell" },


MuhammedZakir:
Lexers are fetched from upstream: https://github.com/orbitalquark/scintillua/. You should open the PR there.

qiu-x:
I see, thanks for the info. And what about the changes in `lua/plugins/filetype.lua`? I think that those are still necessary here.

jgarte:
> Lexers are fetched from upstream: https://github.com/orbitalquark/scintillua/. You should open the PR there.

Would it be possible to install the lexers from [orbitalquark](https://github.com/orbitalquark/scintillua/) instead of vendoring them in?

MuhammedZakir:
> I see, thanks for the info.

You're welcome!

> And what about the changes in `lua/plugins/filetype.lua`? I think that those are still necessary here.

Changes to that should be made here.

> > Lexers are fetched from upstream: https://github.com/orbitalquark/scintillua/. You should open the PR there.
> 
> Would it be possible to install the lexers from [orbitalquark](https://github.com/orbitalquark/scintillua/) instead of vendoring them in?

I think it was done this way for bundling everything together, which makes installing easy. Another important advantage is that, if the upstream made any backward-incompatible changes, you can prevent `vis` from breaking, which is not possible if lexers are directly fetched from upstream.

ninewise:
Applied in eb96e0ce8143804f5a7a37eb76a4b86d8871dd76

PR #1015, opened 2022-06-21
filetype: simplified loop for sanitizedfn, (uses gsub instead of find->sub)

Nomarian:
also ignoresuffix pattern is now `/~+$/` instead of `/~$/`


ninewise:
Applied in 1ca937518ed5ca00e18e5e59e1fef0d9bd628e16, thanks!

PR #1019, opened 2022-06-29
Add filetype detection for SPEC file for RPM packages building.

mcepl:
Working on lexer for vis (which I will probably add to [scintillua](https://github.com/martanne/vis/tree/scintillua) branch), but I would like to add filetype first.


Nomarian:
both at the same time I think, its not like you can accept the filetype first and then the lexer. the lexer must exist.

ninewise:
Applied in f3828efe9dcb51c29e33ff0ee7a3ffbbb8789918

PR #1023, opened 2022-07-04
vis: Compare inodes instead of filenames

tosch42:
Hi,

This fixes a rather old TODO.

I'm unsure if the check for the return value of `stat(2)` should be done in an `if` statement, rather than in a simple `bool` variable. I chose the `if` statement out of personal preference.

Best Regards


ninewise:
Merged in cbeda11801c95687498e2deb6a0144b562581d16.

PR #1018, opened 2022-06-29
Resync the lexers with Scintillua (WIP)

qiu-x:
Related to #908. I fetched the lexers from Scintillua and also added some changes from the `scintillua` branch. 
Please let me know if there are other necessary changes to make this work properly.

PR #1025, opened 2022-07-12
Sync scintilua repo

Nomarian:
It works, but I'm not confident in the branch itself, basically copied every file from scintilua to the lexers folder, copied lexer.lua moved it into plugins/ and changed a bit so it actually works with vis.lua.

lpeg is now a global as well.

Do not merge until MUCH testing.


mcepl:
Despite continuing work upstream on improving our lexers, I don’t see anything stopping us from merging this.

mcepl (approved):
Despite continuing work upstream on improving our lexers, I don’t see anything stopping us from merging this.

Nomarian:
html is broken.

mcepl:
So, closing this in favour of #1018? Is there something which needs to be transferred there?

qiu-x:
Yes, if anything needs to be transferred we can continue the discussion in #1018

mcepl:
So, could you close this PR, please, to make @ninewise happy. You don’t want to make him unhappy, don’t you?

PR #1026, opened 2022-07-14
Update used OSes for GitHub Actions.

mcepl:


ninewise:
Could you disable the (failing) windows checks while you're at it? Or fix them, but from IRC I understood it's not an easy fix.

ninewise:
Applied in f1f15b72878d33923f3744fbf530f5729ae5e8cc

PR #1029, opened 2022-07-25
Actually match filename

Nomarian:
Forgot the $


ninewise:
Applied in cad2b9a9d74c59ac05524f4e747d72ea7a62796c

PR #1038, opened 2022-08-29
build: simplify generating single payload

eworm-de:
We can make `od` skip the address radix, so `sed` does not
need to remove it.


rnpnr (approved):
Looks good to me. You can also drop the extra `-e` for `sed` since there is only one command.

eworm-de:
Updated to drop `-e` from `sed`.

rnpnr (approved):
This looks great to me!

I believe you were given push access so you can merge it if you want. Or I can if you prefer.

Let me know if you have any questions.

eworm-de:
Yes, I was granted push access, but I did not want to misuse after you started a review already. 😉

mcepl:
Don’t forget to push to https://git.sr.ht/~martanne/vis as well.

eworm-de:
Just registered as `eworm`... Can anybody give permissions?

rnpnr:
I think only @martanne can give you permission. I will just push to Sourcehut for now.

PR #1039, opened 2022-08-30
Fix luadoc to use ~= for status

proskur1n:
`vis:pipe` returns an integer status.


mcepl:
@ninewise This is probably not anything super-important, but exactly because of that it could be easily quickly merged, couldn’t it?

rnpnr:
Yes this is correct. Sorry for the wait. I have made this my first merge. Thanks for the patch!

PR #1045, opened 2022-10-22
vis, vis-menu: Add readline Ctrl+G binding

hazen2215:


ninewise:
Looks good, but I don't think we should be adding this by default. I'll be closing the PR, but thank you for your interest in vis!

PR #1050, opened 2022-11-15
vis-clipboard: make xsel honor --selection

njhanley:
vis-clipboard currently ignores the `--selection` option when wrapping xsel because the flag `-b` (use clipboard selection) is passed afterward, overriding the choice. For example:

```
$ echo unset | xsel --clipboard
$ echo unset | xsel --primary
$ echo clipboard | vis-clipboard --copy --selection clipboard
$ echo primary | vis-clipboard --copy --selection primary
$ xsel --clipboard
primary
$ xsel --primary
unset
```


mcepl:
LGTM

rnpnr:
This is correct. `-b` forces `xsel` to copy to the `CLIPBOARD` selection which overrides the previous `--"$sel"` flag.

PR #1051, opened 2022-11-16
vis: remove unused Arg union member

njhanley:
This was missed when pairwise selection combinators were removed in 404bb95..d1d5853.


mcepl:
Everything works for me with this patch applied.

mcepl (approved):
Works for me.

ninewise:
Applied!

PR #1052, opened 2022-11-16
fix miscellaneous spelling mistakes

njhanley:
Ran most of the repo through aspell. It might be nit-picky but I believe this makes the documentation more readable and may help with grepping around.

### DIFF
--- a/array.h
+++ b/array.h
@@ -33,14 +33,14 @@ typedef struct {
 } Array;
 
 /**
- * Initalize an Array object to store pointers.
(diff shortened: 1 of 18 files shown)


mcepl (approved):
@ninewise I think this is completely safe to merge, few variables are (consistently, it seems) renamed but otherwise it is all in comments, so as whole this patch seems to be NOOP.

ninewise:
merged in de8e3bb7b6fedc0554cee20834001dd4a739cd88, thanks!

PR #1054, opened 2022-11-22
build: update alpine in docker build to version 3.17

eworm-de:
This requires one extra package, as static liblzma (liblzma.a) was moved to xz-static.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Run 'make docker' to build a statically linked vis executable!
-FROM i386/alpine:3.16
+FROM i386/alpine:3.17
 RUN apk update && apk add --upgrade --no-cache \
 	acl-dev \
 	acl-static \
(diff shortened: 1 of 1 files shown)


mcepl:
This looks very trivial, please, ninewise, merge.

mcepl:
Wouldn’t it be better to maintain the Dockerfile in some other repo, please? This makes unnecessary churn on questionable technology (how many users of vis use it via Docker?) putting pressure on already overworked maintainers.

@ninewise, @martanne ?

eworm-de:
This is really trivial... Just pressing the merge button is all we need. 😜

But seriously... Mostly this is just bumping the version. If build breaks I will handle this - which makes the above apply.

ninewise:
Applied on good faith: I'm not willing to install docker just to test this.

I do agree with mcepl that it's a bit weird to have a dockerfile for vis in the main repository.

eworm-de:
Thanks!

Well, it goes hand in hand with `vis-single.c` and more bits. No idea how easy it is to put all needed code in a separate repository.

PR #1056, opened 2022-11-23
Improve man lexer, and generalize it to more file extensions

Eolien55:
The current `man` lexer has some issues : it is not a man lexer, but really a troff one; it is very bad for highlighting troff documents that do not use the `man` macros. I thus replaced the old `man` lexer by a more general `troff` one. I also changed the filetypes entry, so it accepts more macro sets.

Note: Some man-specific style was kept (headings), and this is more a groff highlighter than a general troff one because of the groff imperative programming that gets highlighted (`while`s, `break`s, etc.).


ninewise:
Could you send the changes to the lexers upstream instead, to https://github.com/orbitalquark/scintillua? We're going to avoid diverging. The filetype changes are welcome, though.

Eolien55:
Should I wait for https://github.com/orbitalquark/scintillua/pull/84 to be resolved to make the `filetypes` change?

mcepl:
> Should I wait for [orbitalquark/scintillua#84](https://github.com/orbitalquark/scintillua/pull/84) to be resolved to make the `filetypes` change?

Unfortunately, it seems you have to. But you can certainly remove changes to `man.lua` and `troff.lua` from this PR, they belong (if anywhere) upstream.

mcepl:
> Well, scintillua is going through a [re-modernization of its lexers](https://github.com/orbitalquark/scintillua/issues/76), and the `man` one (and all the lexers used in `vis`) are considered legacy. This also means the lexer in the afformentioned PR is using those new features, provided by Scintillua's `lexer.lua`, so using this work from upstream would also require updating the lexer system a bit.

Yup, but that’s the problem of https://github.com/orbitalquark/scintillua/pull/84, it doesn’t belong here, does it?

ninewise:
You could already do the filetype change if you rename the lexer file, too, but I'd rather not merge that to avoid future confusion (us having a troff lexer which will then conflict with the change troff lexer from upstream, ...)

Eolien55:
> Yup, but that’s the problem of [orbitalquark/scintillua#84](https://github.com/orbitalquark/scintillua/pull/84), it doesn’t belong here, does it?

Indeed, but what I was saying is that we cannot use the upstream change without _also_ using the upstream `lexer.lua`. I guess I'll just open another PR for that

Eolien55:
> I'd rather not merge that to avoid future confusion

I agree. Closing this as of now.

mcepl:
> > Yup, but that’s the problem of [orbitalquark/scintillua#84](https://github.com/orbitalquark/scintillua/pull/84), it doesn’t belong here, does it?
> 
> Indeed, but what I was saying is that we cannot use the upstream change without _also_ using the upstream `lexer.lua`. I guess I'll just open another PR for that

I have prepared branch https://git.sr.ht/~mcepl/vis/log/devel_scintillua (hopefully, Soon™ to be merged by @rnpnr ) which upgrades vis lexers (including lexer.lua) to the upstream tag [scintillua_6.2](https://github.com/orbitalquark/scintillua/releases/tag/scintillua_6.2). Does it help you to revive your effort here?

Issue #1032, opened 2022-08-06
color flash upon startup & additional empty char on statusbar


wael444:
https://user-images.githubusercontent.com/47404953/183268280-bf93a004-7f02-4033-84bb-29c0b0565060.mp4

MaxGyver83:
Are you using an alias?
Could you try another terminal emulator? Or a another shell?

wael444:
> Are you using an alias?

No

> Could you try another terminal emulator?

Change: there is an entirely useless bar in stock alacritty settings, also still a blank character:
![image](https://user-images.githubusercontent.com/47404953/183295270-2be7fd22-92a8-4023-b393-1e7b2d88810d.png)


> Or another shell?

No change.

MaxGyver83:
OK, it's strange. Here, it doesn't flicker.

Regarding the "empty char":
Do you mean the space next to the `1`? I think, it's just padding. There is also one blank next to the line and column number in vim.
Or do you mean the black area right to the bar? I think this also some kind of padding. When your window has a width of 170.7 characters, then the terminal uses the space needed for 170 characters and leaves the rest (0.7 character width) black. At least [st](https://st.suckless.org/) does it like that.

wael444:
> OK, it's strange. Here, it doesn't flicker.

on stock config it occurs, here is my config:

```lua
require('vis')

vis.events.subscribe(vis.events.WIN_OPEN, function(win)
        vis:command('set theme base16-gruvbox-dark-hard')
end)
```
if it matters, i compiled vis with these flags: `-O3 -pipe -march=x86-64-v3 -mtune=generic -fno-math-errno -fstack-protector-strong --param ssp-buffer-size=4 -D_FORTIFY_SOURCE=2`

leorosa:
Can you check if the flicker also occurs when setting your theme earlier? I.e.
```
vis.events.subscribe(vis.events.INIT, function()
        vis:command('set theme base16-gruvbox-dark-hard')
end)
```

wael444:
still occurs:
```lua
require('vis')

vis.events.subscribe(vis.events.INIT, function()
        vis:command('set theme base16-gruvbox-dark-hard')
end)
```

wael444:
is there a way i can get around this? i could probably pull of ignoring this but this really throws off my eyes each time i launch vis.

wael444:
solved in 0.8

Issue #327, opened 2016-05-27
Entering command mode causes redraw/flicker of entire window


rwcarlsen:
Everytime I enter command mode (i.e. type ":" from normal mode), the entire window buffer is redrawn and flickers in a distracting way.

martanne:
The problem seems to be related to window resizing, which happens as a result of the new command prompt window.

In general vis' display code assumes that redraws to its internal data structures are cheap and the latest drawn state can then efficiently be "blitted" onto the screen by only updating the damaged screen parts.

For the curses backend this translates into a series of `wnoutrefresh` calls followed by a `doupdate`. Due to the assumed double buffering performed by curses this should do what we want. However the window resizing seems to interfere with this. At first I suspected the [`erase`](https://github.com/martanne/vis/blob/39e089cc057f141ed440635493a041dc321eb2c3/ui-curses.c#L788) call causes the problem, but removing it does not seem to improve things.

I don't have time right now to further look into this, possibly I will have to check how the double buffering in libcurses is actually implemented.

@rgburke has contributed some display related patches, maybe he knows right away what the problem is.

rgburke:
I'm not sure I can offer an explanation for the problem but I can offer an observation.

For the past few months I've played around with vis in Ubuntu using the GNOME Terminal and in FreeBSD using the xfce4-terminal, simply using the default settings in both cases. I've never noticed any flickering before when using vis and I couldn't reproduce it in these environments even when typing :<Enter> repeatedly in quick succession.
Out of interest I installed urxvt on both systems and began playing around with vis. It was at this point I noticed it would occasionally flicker when opening the command prompt.

I'm speculating at this point, but I suspect the issue may be related to the terminal emulator used. It would be interesting to know which terminal emaulators are used by those who have observed this issue.

Screwtapello:
I don't have a lot of terminals handy, but:

- xterm 324 does not flicker when opening the command-line at 80x25, but it does when maximised, and it does flicker a bit when resizing the window
- gnome-terminal 3.20.2 never flickers when opening the command-line, but does when resizing the window
- tmux inside gnome-terminal seems to be much like native gnome-terminal
- pterm 0.67 never flickers when opening the command-line, but flickers heavily during resize
- terminology 0.7.0 never flickers when opening the command-line, but flickers heavily during resize

[...]

PR #1061, opened 2022-12-28
Prevent flickering in curses

mujo-hash:
Reading from curs_refresh(3X) from curses, calling doupdate() repeatedly will cause 'several bursts of output to the screen'. wnoutrefresh() has the smarts to only copy the changed lines to the copied virtual screen, but doupdate() does not.

There have been several bug reports related to flickering but all seems to be inconsistenly reproducible due to different terminal buffering behavior. See #1032, #327

Unfortunately, when I am using a slow display, I still notice flickering, so this commit changes the routines for opening new windows and splitting windows to wait until the last change is finished before calling doupdate().

### DIFF
--- a/ui-terminal-curses.c
+++ b/ui-terminal-curses.c
@@ -230,7 +230,8 @@ static void ui_curses_blit(UiTerm *tui) {
 		}
 	}
 	wnoutrefresh(stdscr);
-	doupdate();
+	if (tui->doupdate)
+		doupdate();
 }
 
 static void ui_curses_clear(UiTerm *tui) {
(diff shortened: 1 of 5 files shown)


mujo-hash:
I've noticed a few rendering bugs with this patch, particularly among themes that use transparency, so I'm not sure it's good enough to merge, but it works well enough for my needs.

rnpnr (approved):
Everything here looks good to me. I have also been using it for a while
and haven't noticed any problems.

I might regret this later but since no one who cares about transparent
terminals has stepped up to test this I am going to assume its fine
to merge.

erf:
I've been using vis with this patch with a transparent terminal and seem to work fine for me, but i can't really notice any different in flickering now.

rnpnr:
Sounds good. I think it also makes sense conceptually to wait until
updates are complete before redrawing the window. I've been working in
vis for days with multiple windows opening and closing and I haven't
noticed any misbehavior so I can't see a reason not to apply this.

rnpnr:
Applied in d1f2c27, thanks for the patch!

Small tip for next time: put any changes on a separate branch so I
can force push to it after rebasing on top of master and Github will
automatically close this as merged.

PR #1066, opened 2023-01-08
expose UI layout and allow it to be set from lua API

### LINKED ISSUE #812: Window layout property
I'd appreciate the addition of a `Vis`, `Window` or `Ui` class property (Lua API) that would expose the current window layout for reading.

I'd use this for conditional split/vsplit and new/vnew command execution in plugins.

I know that I can achieve the same thing through window width and height comparison but having it exposed through a dedicated property sounds neat.

rnpnr:
one use of this is to automatically set the layout on WIN_OPEN based on the width and height of existing windows.

fixes #812

### DIFF
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -672,6 +672,11 @@ err:
 	return false;
 }
 
+enum UiLayout ui_layout_get(Ui *ui) {
+	UiTerm *tui = (UiTerm *)ui;
+	return tui->layout;
+}
+
 Ui *ui_term_new(void) {
 	size_t styles_size = UI_STYLE_MAX * sizeof(CellStyle);
 	CellStyle *styles = calloc(1, styles_size);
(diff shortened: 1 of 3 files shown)


ninewise:
This looks interesting, but I'm not sure we want to expose this. We sort-of want to avoid having powerful window management built into vis, and this allows to develop plugins just for that. On the other hand, this might make people happy until we properly fix it with a client/server architecture.

Regardless, I need to see if I can regenerate and publish the luadoc somehow before applying this.

rnpnr:
Thats a fair point. I would definitely prefer to relegate this to my window manager without needing to write a bunch of helpers. My justification for this though is that this patch simply makes it less tedious to do something that is already possible in vis. For example you could use some global variable in `visrc.lua` and then use `:set layout` with the annoyance that the variable can easily get out of sync.

Re luadoc: I definitely thought is was being automatically regenerated based on the source tree but I see that is not the case. I think it just displays the stuff in the [gh-pages](https://github.com/martanne/vis/tree/gh-pages) branch.

PR #1075, opened 2023-02-11
fix #1074 where secondary selections where considered visible

### LINKED ISSUE #1074: Slave selection strangled by view cliff
**BUG**
`seq 100 | vis - | diff - <(seq 100 | awk '!/9$/')`

The above should yield **no difference** when the following MARTANNE vis keystrokes are done:

```
:x/$9/
V
8k
8j
d
<Esc><Esc>
:wq
```

This is the output I see instead:

```
35a36,44
> 40
> 41
> 42
> 43
> 44
> 45
> 46
> 47
> 48
```

jeremybobbin:


ninewise:
Applied, thanks

Issue #1086, opened 2023-03-24
Cursor jumps to the top of the view pane when navigating past the end of file


rnpnr:
To recreate:
1. open a file
2. navigate to the last character `G$`
3. try to move to the next character `l`
4. cursor jumps to the top of the window rather than staying in place
 
Commit https://github.com/martanne/vis/commit/c22b2c2937d9dd65d0609e57f948f08da7f5c5ea seems to be where this appeared.

ninewise:
Fixed by Jeremy Bobbin in #1087

Issue #1084, opened 2023-03-19
Cursor stays in place when inserting text in a new file


erf:
Cursor now stays in the same place when inserting text in a new file.

How to recreate:

1. open a new file
2. press `i` to insert text 
3. enter some text `hello`
4. cursor stays at first position `h`

This seem to have been introduced in commit: c22b2c2

Disonantemus:
> Cursor now stays in the same place when inserting text in a new file.
> 
> How to recreate:
> 
>     1. open a new file
> 
>     2. press `i` to insert text
> 
>     3. enter some text `hello`
> 
>     4. cursor stays at first position `h`
> 
> 
> This seem to have been introduced in commit: [c22b2c2](https://github.com/martanne/vis/commit/c22b2c2937d9dd65d0609e57f948f08da7f5c5ea)

I'm still experiencing the same, last git:
~~~
> vis -v
vis v0.7-74-gc9737a7 +curses +lua +acl
~~~

Box:
~~~
OS: Arch Linux x86_64
Host: MS-7816 2.0
Kernel: 6.1.14-1-lts
Terminal: tmux
CPU: Intel i7-4790 (8) @ 3.600GHz
GPU: AMD ATI Radeon RX 470/480/570/570X/580/580X/590
Memory: 2897MiB / 15939MiB
~~~

mcepl:
Or you can use my fork at https://git.sr.ht/~mcepl/vis which has it merged in its master branch.

Disonantemus:
> Yes, #1087 hasn't been merged yet. You can apply it locally in the meantime to get back up and running.

I'm no programmer, I just create github account to report issues.
I don't know how to do that. Sometimes I just do: `git clone <asdf>`
but I don't know how to merge with git or understand that.

mcepl:
@Disonantemus you build vis yourself, right? If so then just do `git clone https://git.sr.ht/~mcepl/vis` instead of cloning from this GitHub repo and build the program.

Disonantemus:
> @Disonantemus you build vis yourself, right? If so then just do `git clone https://git.sr.ht/~mcepl/vis` instead of cloning from this GitHub repo and build the program.

Thanks, I'm using your repo now!

ninewise:
#1087 was merged, sorry for the wait. Do reopen should further problems ensue.

mcepl:
Awesome, one commit merged, fourty-two to go!

ninewise:
[...]

PR #1087, opened 2023-03-27
check for EOF before unsetting row, col & line cache in view_coord_get

### LINKED ISSUE #1074: Slave selection strangled by view cliff

jeremybobbin:
fixes #1074 & #1084 & #1086
https://github.com/martanne/vis/issues/1074
https://github.com/martanne/vis/issues/1084
https://github.com/martanne/vis/issues/1086

### DIFF
--- a/view.c
+++ b/view.c
@@ -283,7 +283,8 @@ bool view_coord_get(View *view, size_t pos, Line **retline, int *retrow, int *re
 	size_t cur = view->start;
 	Line *line = view->topline;
 
-	if (pos < view->start || pos >= view->end) {
+	bool eof = view->end == text_size(view->text);
+	if (pos < view->start || (pos >= view->end && !eof)) {
 		if (retline) *retline = NULL;
 		if (retrow) *retrow = -1;
 		if (retcol) *retcol = -1;


mcepl:
After the disaster of https://github.com/martanne/vis/issues/1074, it would be probably prudent to explain slightly more why do you think that this PR works better than the other one.

erf:
@mcepl "disaster" ..? Please try to keep a positive language when commenting . There really isn't much development going on here lately, and we don't want to discourage people who actually wants to contribute !

jeremybobbin:
Personally I do not mind the language. :)
I believe "disaster" is a fair characterization, but then again, the fact that #1074 had gone unnoticed for as long as it did, is also disastrous - especially given vis is "Combining Modal Editing with Structural Regular Expressions".

If I didn't use structural regular expressions, I'd be using vim or neovim.

The change from `pos > view->end` to `pos >= view->end`  made it such that if a selection was beyond the view, `view_coord_get` would return false.
I believe `view_coord_get` is supposed to return whether or not a position in the buffer is visible.

There is a case, however, where the position can be at the view's end & still be visible, and that's when EOF is in the view.
In this patch, we check if the view's end is EOF.

mcepl:
> Personally I do not mind the language. :) I believe "disaster" is a fair characterization, but then again, the fact that #1074 had gone unnoticed for as long as it did, is also disastrous - especially given vis is "Combining Modal Editing with Structural Regular Expressions".

True, and I am sorry if I am too direct, but how else should I call a commit which had to be withdrawn? A shining success? And of course it doesn’t mean anything about yourself personally, I do such disasters a dozen weekly. :(

And, of course, I appreciate a lot that somebody finally started to look at the C part of vis.

> The change from `pos > view->end` to `pos >= view->end` made it such that if a selection was beyond the view, `view_coord_get` would return false. I believe `view_coord_get` is supposed to return whether or not a position in the buffer is visible.
> 
> There is a case, however, where the position can be at the view's end & still be visible, and that's when EOF is in the view. In this patch, we check if the view's end is EOF.

Again, I am not the greatest C programmer myself, but doesn’t `view->end != text_size(view->text))` look like a way too complicated way how to say `!EOF()`? Shouldn’t we have some at least macro for it?

erf:
> True, and I am sorry if I am too direct, but how else should I call a commit which had to be withdrawn? A shining success? And of course it doesn’t mean anything about yourself personally, I do such disasters a dozen weekly. :(

It doesn't have to be either a "disaster" or a "shining success", just try to be somewhat professional and ask for more information, in a polite way. No reason to address a failed attempt at fixing an issue like a "disaster", it doesn't encourage more attempts at fixing stuff, if we don't get it right the first time. 

Also please read the [Github Community Guidelines](https://docs.github.com/articles/github-community-guidelines) if you have not. We want to have a positive community around this project.

jeremybobbin:
> but doesn’t view->end != text_size(view->text)) look like a way too complicated way how to say !EOF()? Shouldn’t we have some at least macro for it?

Makes sense - from another function in view.c:406:

```c
		bool eof = view->end == text_size(view->text);
		if (view->line->len == 0 && eof && view->line->prev)
```

I've changed the patch to match the above.

> No reason to address a failed attempt at fixing an issue like a "disaster"

In mcepl's defense, #1074 introduced a bug which would be easily noticed by users who are opening a vis buffer for the first time, possibly leaving a bad impression.
If I were driven away by the "disaster" characterization, what would that say about the value I see in Vis?

That aside, these sorts of issues should be covered in the test suite. Writing a test for that now.

mcepl:
> ```c
> 		bool eof = view->end == text_size(view->text);
> ```

Isn’t this line a perfect candidate for refactoring (that’s why I was talking about macro)?

> That aside, these sorts of issues should be covered in the test suite. Writing a test for that now.

Yes, a low coverage of our test suite is much larger problem. I think we should include https://github.com/martanne/vis-test to this repo and start to working more intensly on increasing the coverage, but that is aside of this particular issue.

rnpnr:
I've been using this for a day and it seems correct.

It also fixes another unreported issue with the previous commit where starting a visual line selection and moving the cursor to the end of window would make it look like the selection was lost.

I'm in favor of merging this rather than reverting the previous commit.

mcepl:
> It also fixes another unreported issue with the previous commit where starting a visual line selection and moving the cursor to the end of window would make it look like the selection was lost.
> 
> I'm in favor of merging this rather than reverting the previous commit.

Oh, I had https://github.com/martanne/vis/issues/1074 already reverted, you mean, it could be kept around with this commit on the top?

rnpnr:
> Oh, I had #1074 already reverted, you mean, it could be kept around with this commit on the top?

This commit applies on top of c22b2c2937d9dd65d0609e57f948f08da7f5c5ea yes.

mcepl:
Merged into https://git.sr.ht/~mcepl/vis/tree.

mcepl (approved):
Merged into https://git.sr.ht/~mcepl/vis/tree.

ninewise:
Applied here as well, thanks!

PR #1089, opened 2023-04-03
Use /proc/curproc/file in freebsd when adding directories to package.path

Nomarian:
only activates if using ./configure freebsd.

### DIFF
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -2765,7 +2765,13 @@ void vis_lua_init(Vis *vis) {
 		vis_lua_path_add(vis, path);
 	}
 
-	ssize_t len = readlink("/proc/self/exe", path, sizeof(path)-1);
+	ssize_t len = readlink(
+	   #if __BSD_VISIBLE
+	       "/proc/curproc/file"
+	   #else
+	       "/proc/self/exe"
+	   #endif
+	   , path, sizeof(path)-1);
 	if (len > 0) {
 		path[len] = '\0';
 		/* some idiotic dirname(3) implementations return pointers to statically


rnpnr:
This doesn't seem very portable and I don't think its even guaranteed to work on FreeBSD (procfs is entirely optional as far as I know).

A better interface which is available on all the major BSDs is [`getprogname(3)`](https://man.openbsd.org/getprogname.3).

Nomarian:
It's not portable, that's why __BSD_VISIBLE is used because that's only set on ./configure freebsd


And it does work on freebsd. I used it.


getprogname.3 just prints the program name/filename. but not the filepath, which is what readlink /proc/self/exe does. It's why its used, because when you make vis, you will ./vis, which will use ./lua,  Honestly, I would remove the entirety of it and just set VIS_PATH=$PWD/lua ./vis instead or get rid of the entirety of checking for /lua in the build directory and instead would use the equivalent of 

```
realpath $( dirname $( readlink /proc/self/exe ) )/../share/vis/lua
```

rnpnr:
Sorry I should have said "Not portable between BSDs". If you were fixing it on one I thought it should be fixed on all of them.

But yes now looking at the surrounding code I would also rather it just get removed. I don't see the point of having a specific edge case for running out of the build directory. I don't think anyone is putting a `/lua` folder next to the binary elsewhere because they probably already have a `lua` binary there.

Nomarian:
> Sorry I should have said "Not portable between BSDs". If you were fixing it on one I thought it should be fixed on all of them.


That's asking a lot.


Anyway, I guess its good for testing, which is why its necessary... It should be removed and replaced with the whole realpath thing, but that would require a build/ directory, which would require changing make or configure to reflect the install directory.

mcepl:
So, what is the decision about this PR, will we close it or not?

Nomarian:
I will close it, its better to put this in freebsd's ports itself but this raises some issues


- There is no FREEBSD_SOURCE defined
- BSD_SOURCE should be defined for all BSDs in configure
- add a OPENBSD_SOURCE
- /proc/self/exe is linux only

PR #1094, opened 2023-04-25
fix use after free in cmd_files()

### LINKED ISSUE #1090: X/re/q doesn't work
Thanks for the great work, it seems that `X/re-files/q` doesn't work.

rnpnr:
When the cmd closes the window the window pointer gets freed along the way. We can't use win->next to update the loop variable if sam_execute() has been called. Instead we can store win->next early and use that variable to continue the loop.

fixes #1090

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -1564,14 +1564,16 @@ static bool cmd_print(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 
 static bool cmd_files(Vis *vis, Win *win, Command *cmd, const char *argv[], Selection *sel, Filerange *range) {
 	bool ret = true;
-	for (Win *win = vis->windows; win; win = win->next) {
-		if (win->file->internal)
+	for (Win *wn, *w = vis->windows; w; w = wn) {
+		/* w can get freed by sam_execute() so store w->next early */
+		wn = w->next;
+		if (w->file->internal)
 			continue;
 		bool match = !cmd->regex ||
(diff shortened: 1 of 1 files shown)


sam.c
```diff
@@ -1564,14 +1564,16 @@ static bool cmd_print(Vis *vis, Win *win, Command *cmd, const char *argv[], Sele
 
 static bool cmd_files(Vis *vis, Win *win, Command *cmd, const char *argv[], Selection *sel, Filerange *range) {
 	bool ret = true;
-	for (Win *win = vis->windows; win; win = win->next) {
-		if (win->file->internal)
+	for (Win *wn, *w = vis->windows; w; w = wn) {
+		/* w can get freed by sam_execute() so store w->next early */
+		wn = w->next;
```
mcepl, line 1569: I am not much of a C programmer, but isn’t this the canonical situation, where this line should go to the `for()` statement itself?

rnpnr: It looks like it but it can't. That is why I added a comment.

If it goes in the first statement of the for loop it will only get called before the loop actually starts and never updated.
If it goes in the third statement it will run after `sam_execute()` which can free `w` and make the final `w->next` invalid.

mcepl: I see, you are right … I am really not a C programmer, the third clause (which is what I meant) is done AFTER the body, not before. Sorry.

+1 from me then.

mcepl (approved):

mcepl:
+1 from me.

mcepl:

PR #1096, opened 2023-05-02
remove unused vis_message_hide() function

rnpnr:
the message window gets closed like a normal window and nothing calls this function. In fact, it wasn't even used when it was added 8 years ago in 979ab79.


mcepl:

mcepl (approved):

Issue #1033, opened 2022-08-08
Some issues with the man page


ilyagr:
- The key table in the visual mode subsubsection of the `Multiple selections` subsection of the [Vim commands](https://manpages.ubuntu.com/manpages/kinetic/man1/vis.1.html#vi(m)%20key%20bindings) section is messed up. For example, `<C-u>` should correspond to making the previous selection primary and `<C-k>` should probably not be in the table at all.

- Specifically in the HTML version of the man page linked to from the README, http://martanne.github.io/vis/man/vis.1.html#Multiple_Selections, the corresponding rows in this and many other tables are misaligned. It's hard to know which key combination corresponds to which description.  

    This is what it looks like to me on Chrome:
    <details>

    ![image](https://user-images.githubusercontent.com/4123047/183325305-c396c254-6a61-4e9a-b3dc-5645920f74bb.png)
    </details>

ilyagr:
This is strictly speaking a separate issue, but I'll put it here. I must be missing something obvious, but in all documentation, I'm quite confused by the description of the `<C-c>` command as "Remove count selection column" and the similar description of the `<C-l>` command. I think it removes all the selections that overlap the `count` column, or the current column by default?

mcepl:
I guess, I would say “patches welcome”. Somebody has to go and study how to fix those manpages. Could it be you, please?

rnpnr:
For the first issue `<C-k>` and `<C-u>` do the same thing in visual mode. This is the same for the `<C-d>` and `<C-j>` pair. This is slightly more clear when using a manpager in the terminal but I will submit a patch making it more explicit.

For the second issue this seems to be a bug in [`mandoc.css`](https://github.com/martanne/vis/blob/gh-pages/man/mandoc.css) which was just taken from upstream [mandoc](https://mandoc.bsd.lv/). I'm not intimately familiar with CSS so I don't know what the correct fix is but it can be hacked around with the following patch which has minimal effect on the rest of the page:
```diff
diff --git a/man/mandoc.css b/man/mandoc.css
index d75700a..50c54ec 100644
--- a/man/mandoc.css
+++ b/man/mandoc.css
@@ -143,7 +143,7 @@ h2.Ss {		margin-top: 1.2em;
 		overflow: auto; }
 .Bl-compact {	margin-top: 0em; }
 .Bl-compact > dd {
-		margin-bottom: 0em; }
+		margin-bottom: 0.15em; }
 .Bl-compact > dt {
[...]

PR #1097, opened 2023-05-03
vis.1: list equivalent key bindings on the same line

rnpnr:
it is hard to tell which line `<C-u>` and `<C-d>` are supposed to belong to in the current version.

Note that if keeping the pairs on separate lines is preferred I can update the patch to say `same as xxx` instead.

see #1033

### DIFF
--- a/man/vis.1
+++ b/man/vis.1
@@ -1123,12 +1123,10 @@ clear (skip) current selection, but select next matching word
 .It Aq Ic C-p
 remove primary selection
 .
-.It Aq Ic C-u
-.It Aq Ic C-k
+.It Ao Ic C-u Ac Ao Ic C-k Ac
 make the count previous selection primary
 .
-.It Aq Ic C-d
-.It Aq Ic C-j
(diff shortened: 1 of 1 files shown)


mcepl:
I really don’t which version is better.

Old version:

screenshot-2023-05-03_17-05-1683129313

New version:

screenshot-2023-05-03_17-05-1683129337

rnpnr:
Here are two alternatives:
image
or
image
Are either of them better?

mcepl:
> Are either of them better?

I really don't like the second one. So I think it is between this PR and the current state.

tosch42:
The current state is very confusing. I'd expect <C-p> and <C-u> to do the same thing, since one is reading from top to bottom. I prefer the new version because it supports the natural reading flow.

mcepl (approved):
Merged into https://git.sr.ht/~mcepl/vis/tree.

ninewise:
Applied in e6fc48dea1909e916c13ac58a4ff64a379ef30f0

Issue #1067, opened 2023-01-08
Use system clipboard on MacOS


bscuron:
I am currently trying to use the system clipboard by default, but I receive this message when I compile with the changes by @deepcube mentioned [here](https://github.com/martanne/vis/issues/1003#issuecomment-1118627247): "Command failed System clipboard not supported". I also tried just setting the primary register = clipboard register in `vis.h`. This works inside of vis, but when I paste outside of the editor it does not paste the contents of the yanked text.

Please let me know if you have any ideas that may help. Thank you.

QiBaobin:
The `*` doesn't work, but `+` register works for me, you can use that instead.
```
  +   System clipboard register, see vis-clipboard(1)
  *   Primary clipboard register, see vis-clipboard(1)
```

mcepl:
Yes, it is system dependent and Mac OS (and Windows) doesn’t have whole concept of selection, right? When you select text in one application and then middle-click in other one, nothing happens, right?

I think this ticket can be closed, because we can do nothing about your system.

rnpnr:
I'm wondering if this behaviour even makes sense. Wouldn't it be better to just copy to the available clipboard and ignore the `--selection` argument if its not supported by the system?

bscuron:
This worked for me on MacOS:
```lua
    vis.registers['"'] = { io.popen("vis-clipboard --paste"):read("*all") } -- use system clipboard
``` This was suggested by @leorosa in #1003

PR #1101, opened 2023-05-25
vis-clipboard: don't fail when sel is primary on unsupported platforms

rnpnr:
this is mostly useful for the internal vis usage and makes both `*` and `+` ~buffers~ registers work on macOS/cygwin.

fixes: #1067

Note: I can't test this and I want others opinions on it. I don't even think the deleted check is correct right now since who is to say what the user meant when they specified `--selection` on macOS/cygwin.

Edit: I meant registers not buffers

### DIFF
--- a/man/vis-clipboard.1
+++ b/man/vis-clipboard.1
@@ -53,7 +53,8 @@ In this mode,
 reads the content of the system clipboard,
 and writes it to standard output.
 .It Fl -selection Ar selection
-specify which selection to use, options are "primary" or "clipboard"
+specify which selection to use, options are "primary" or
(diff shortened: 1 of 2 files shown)


mcepl:
Yes, I cannot test it either, but I think it shouldn’t do anything wrong, so I have merged this PR to my repo https://git.sr.ht/~mcepl/vis.

mcepl (approved):
Yes, I cannot test it either, but I think it shouldn’t do anything wrong, so I have merged this PR to my repo https://git.sr.ht/~mcepl/vis.

PR #1102, opened 2023-05-30
allow snake case in command names

### LINKED ISSUE #971: underscore in command name prevents them from running
Registering a command with

```lua
vis:command_register("foo_bar", function(argv, force, win, sel, range)
    print("baz")
end, "guu")
```

Works, however, calling `:foo_bar` does nothing.

rnpnr:
fixes #971

I have run into this before and it has always seemed like a weird requirement. Anyone against this?


ninewise (approved):
Sure for user commands, but I wouldn't start doing this for new built-in commands to remain consistent.

rnpnr:
Looking at this again it won't work if `_` is the last character in the command name. I think that case should be fine for user commands as well. Then correct check should be:
```c
return c && !isspace(c) && !isdigit(c) && (!ispunct(c) || c == '_' || (c == '-' && valid_cmdname(s+1)));
```
I will fix this and merge if that still sounds fine.

PR #1105, opened 2023-06-16
Properly restore terminal after running fullscreen program

captaingroove:
When running a fullscreen program (curses) with vis:pipe() the terminal is not restored correctly. There's no way to automatically detect whether a program has to be run in shell or fullscreen mode, so a parameter needs to be added to vis_pipe_*() and also vis:pipe() in the Lua API. This is useful for e.g. running interactive programs like nnn as a file picker, see https://github.com/captaingroove/vis-nnn.

### DIFF
--- a/ui.h
+++ b/ui.h
@@ -99,7 +99,7 @@ struct Ui {
 	void (*suspend)(Ui*);
 	void (*resume)(Ui*);
 	bool (*getkey)(Ui*, TermKeyKey*);
-	void (*terminal_save)(Ui*);
+	void (*terminal_save)(Ui*, bool fscr);
 	void (*terminal_restore)(Ui*);
 	TermKey* (*termkey_get)(Ui*);
 	int (*colors)(Ui*);
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1331,6 +1331,7 @@ static int exit_func(lua_State *L) {
  * @tparam[opt] File file the file to which the range applies
  * @tparam[opt] Range range the range to pipe
  * @tparam string command the command to execute
+ * @tparam[opt] bool fullscreen whether command is a fullscreen program (e.g. curses based)
  * @treturn int code the exit status of the executed command
  * @treturn string stdout the data written to stdout
  * @treturn string stderr the data written to stderr
@@ -1348,7 +1349,8 @@ static int pipe_func(lua_State *L) {
 		range = getrange(L, 3);
 	}
 	const char *cmd = luaL_checkstring(L, cmd_idx);
-	int status = vis_pipe_collect(vis, file, &range, (const char*[]){ cmd, NULL }, &out, &err);
+	bool fullscreen = lua_isboolean(L, cmd_idx + 1) && lua_toboolean(L, cmd_idx + 1);
+	int status = vis_pipe_collect(vis, file, &range, (const char*[]){ cmd, NULL }, &out, &err, fullscreen);
 	lua_pushinteger(L, status);
 	if (out)
 		lua_pushstring(L, out);
@@ -1363,6 +1365,7 @@ static int pipe_func(lua_State *L) {
 	vis_draw(vis);
 	return 3;
 }
+
 /***
  * Redraw complete user interface.
  *
--- a/vis.c
+++ b/vis.c
@@ -1729,7 +1729,8 @@ Regex *vis_regex(Vis *vis, const char *pattern) {
 
 int vis_pipe(Vis *vis, File *file, Filerange *range, const char *argv[],
 	void *stdout_context, ssize_t (*read_stdout)(void *stdout_context, char *data, size_t len),
-	void *stderr_context, ssize_t (*read_stderr)(void *stderr_context, char *data, size_t len)) {
+	void *stderr_context, ssize_t (*read_stderr)(void *stderr_context, char *data, size_t len),
+	bool fullscreen) {
 
 	/* if an invalid range was given, stdin (i.e. key board input) is passed
 	 * through the external command. */
@@ -1754,7 +1755,7 @@ int vis_pipe(Vis *vis, File *file, Filerange *range, const char *argv[],
 		return -1;
 	}
 
-	vis->ui->terminal_save(vis->ui);
+	vis->ui->terminal_save(vis->ui, fullscreen);
 	pid_t pid = fork();
 
 	if (pid == -1) {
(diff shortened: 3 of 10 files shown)


ui.h
```diff
@@ -99,7 +99,7 @@ struct Ui {
 	void (*suspend)(Ui*);
 	void (*resume)(Ui*);
 	bool (*getkey)(Ui*, TermKeyKey*);
-	void (*terminal_save)(Ui*);
+	void (*terminal_save)(Ui*, bool fscr);
```
rnpnr, line 102: Since you have changed the function signature here you need to change it in all the backends. Right now this breaks compilation for the vt100 backend.

vis.c
```diff
@@ -1729,7 +1729,8 @@ Regex *vis_regex(Vis *vis, const char *pattern) {
 
 int vis_pipe(Vis *vis, File *file, Filerange *range, const char *argv[],
 	void *stdout_context, ssize_t (*read_stdout)(void *stdout_context, char *data, size_t len),
-	void *stderr_context, ssize_t (*read_stderr)(void *stderr_context, char *data, size_t len)) {
+	void *stderr_context, ssize_t (*read_stderr)(void *stderr_context, char *data, size_t len),
+	bool fullscreen) {
```
rnpnr, line 1733: This variable should probably be called `interactive` not `fullscreen`.

rnpnr (commented):
Overall I'm not sure I like this patch. `vis_pipe()` is for piping not for running interactive programs.

I'm open to hearing what other people think.

vis.c
rnpnr: Actually scratch that. This is already used for programs like `vis-open` which do require interaction.

ui.h
captaingroove: True, thanks for spotting this!
I updated the patch, accordingly.

I used vis_pipe() for two reasons:
1. I wanted to avoid a separate function in the API
2. I think, you should also be able to pipe through interactive programs, e.g. a fuzzy finder. While there are commands in sam.c for different piping scenarios (cmd_filter(), cmd_pipein(), cmd_pipeout(), cmd_launch()) which boil down to calling vis_pipe(), in the Lua API this can be handled by one function, so no specialization is needed. Current plugins tend to use the Lua standard io.popen() for that purpose (e.g. https://github.com/peaceant/vis-fzf-mru.git). However, this relies on the external command to not change the terminal settings, so it's safer to use vis:pipe(), instead.

rnpnr:
Also can't you just call `vis:redraw()` after your command has run and returned?

captaingroove:
Unfortunately not, the terminal state needs to be saved before running the command. That's in `ui_term_backend_save()`, which for the curses backend is different for fullscreen and shell mode programs:
```C
static void ui_curses_save(UiTerm *tui, bool fscr) {
	curs_set(1);
	if (fscr) {
		def_prog_mode();
		endwin();
	} else {
		reset_shell_mode();
	}
}
```

rnpnr:
So now that I better understand this I think this fix is fine. But I think there should be a way of doing it without the need for an extra function parameter. I'm not super familiar with curses and the documentation is not that helpful. From what I see [here](https://invisible-island.net/ncurses/man/curs_kernel.3x.html) I'm not even sure that calling `reset_shell_mode()` was correct to begin with.

captaingroove:
I think using `reset_shell_mode()` here is non-standard but unique. Then curses mode is not completely left and the vis editor window is scrolled while the subprocess writes to stdout. This is how vis composes other processes with the UI without opening a separate window that has terminal emulation capabilities.

Maybe the following simple program illustrates the two alternatives. By default, it runs the subprocess in shell mode and when providing an arbitrary command line argument, it leaves curses properly before running the subprocess. You might also change the subprocess to either shell or curses commands.
```
#include <stdlib.h>
#include <ncurses.h>

int
main(int argc, char** argv)
{
	initscr();                  /// Start curses mode
	printw("Hello curses!\n");  /// Print Hello World
	refresh();                  /// Print it on to the real screen
	getch();                    /// Pause, to see what happened

	if (argc > 1) {
		def_prog_mode();        /// Save the tty modes
		endwin();               /// End curses mode temporarily
	} else {
		reset_shell_mode();
	}
	// system("/bin/sh");       /// Do whatever you like in cooked mode
	system("ls -l");
	// system("nnn");

	getch();                    /// Pause, to see what happened
	reset_prog_mode();          /// Return to the previous tty mode stored by def_prog_mode()
	refresh();                  /// Do refresh() to restore the Screen contents
	printw("Another string\n"); /// Back to curses, use the full capabilities of curses
	refresh();
	getch();                    /// Pause, to see what happened
	endwin();                   /// End curses mode
	return 0;
}
```

ninewise:
This would be useful to have indeed. The API change seems minor.

rnpnr:
I guess what I want to know is if its possible to achieve this without
the extra parameter. Is there a reason why we can't just always call:
```c
def_prog_mode();
endwin();
```
From some very brief testing I'm not really seeing an issue with this.

captaingroove:
> I guess what I want to know is if its possible to achieve this without the extra parameter. Is there a reason why we can't just always call:
> 
> ```c
> def_prog_mode();
> endwin();
> ```
> 
> From some very brief testing I'm not really seeing an issue with this.

The behavior is different. Using
```c
def_prog_mode();
endwin();
```
prints the output of the subprocess to the terminal buffer while hiding the curses buffer of the parent process.

In contrast
```c
reset_shell_mode();
```
scrolls the curses buffer of the parent process while printing the output of the subprocess. The latter is the default behaviour in vis while the former is desirable for subprocesses that also use curses.

rnpnr:
> The behavior is different.

I understood that much but I thought it didn't matter. But I realized
I didn't try `vis-open`.

I'm still not the biggest fan but @ninewise is correct in saying that
the API change is minor. Can you rebase this onto master? Also I think
you missed a doxygen comment for `vis_pipe_collect()`.

captaingroove:
> Can you rebase this onto master? Also I think you missed a doxygen comment for `vis_pipe_collect()`.

Okay, added the comments and rebased.

vis-lua.c
```diff
@@ -1338,19 +1338,20 @@ static int exit_func(lua_State *L) {
...
 		cmd_idx = 2;
-		fscrn_idx = 3;
+		fscr_idx = 3;
 	} else if (!(lua_isnil(L, 2) && lua_isnil(L, 3))) {
 		file = obj_ref_check(L, 2, VIS_LUA_TYPE_FILE);
 		range = getrange(L, 3);
 	}
 	const char *cmd = luaL_checkstring(L, cmd_idx);
-	bool fullscreen = lua_isboolean(L, 5) && lua_toboolean(L, 5);
+	bool fullscreen = lua_isboolean(L, fscr_idx) && lua_toboolean(L, fscr_idx);
```
rnpnr, line 1354: I don't think you need another variable for this. It's always `cmd_idx + 1`.

rnpnr (changes requested):
I think you made a mistake when rebasing. I can fix it but it will be
easier for you to fix it. Here are some quick steps:

1. use `git reflog` to find and checkout the point before you started.
2. run `git pull master` to make sure master is up to date.
3. from your branch run `git rebase master`. It should fail to apply
   your first commit.
4. make the changes and add the file (add only, don't commit).
5. run `git rebase --continue`.
6. repeat from 4 if necessary.

There shouldn't be any merge commits when you are done. You can check
https://git-rebase.io if you need more detailed explanation.

rnpnr:
Applied, thanks for answering all my questions and addressing the issues!

PR #1108, opened 2023-07-03
vis:pipe(): don't segfault if vis->win isn't present

### LINKED ISSUE #1107: Lua API: vis:pipe() causes a segfault when called before a window is open
Since commit [424b219](https://github.com/martanne/vis/commit/424b2190b9207eea9807eef7132cd95e881cbbbf) vis will segfault if `vis:pipe()` is called when
no windows are open. This can happen if you have some function calling
`vis:pipe()` based on an event such as `FILE_OPEN` which doesn't actually
depend on a window being present. For example my [vis-gpg](https://github.com/rnpnr/vis-gpg) plugin does
this when decrypting a file.

A temporary patch will follow in a pull request.

rnpnr:
fixes #1107

Note: This is a bit of a hack and I don't think it should actually be
applied. It only works when file is changed via the `else` branch of
the check that follows.

I will work on a proper fix later if no one has anything better.

### DIFF
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -1340,7 +1340,7 @@ static int pipe_func(lua_State *L) {
 	Vis *vis = obj_ref_check(L, 1, "vis");
 	int cmd_idx = 4;
 	char *out = NULL, *err = NULL;
-	File *file = vis->win->file;
+	File *file = vis->win ? vis->win->file : NULL;
 	Filerange range = text_range_new(0, 0);
 	if (lua_gettop(L) <= 3) {
 		cmd_idx = 2;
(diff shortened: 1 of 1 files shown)


rnpnr:
The commit I just pushed is a little better. `vis:pipe()` will now throw
an error if you try to call it without specifying a file before a window
is open.

You can test it by adding something like the following to `visrc.lua`:

```lua
vis.events.subscribe(vis.events.FILE_OPEN, function()
	err, out = vis:pipe("echo no explosion")
	if (out) then
		vis:message(out)
	end
end)
```

Issue #1060, opened 2022-12-14
:help doesn't display mappings starting with <Space> correctly 


henricattoire:
If I create a mapping that **starts** with a space, the help page only shows the space (and not the rest of the mapping).

```lua
vis:map(vis.modes.NORMAL, " s", function()
   vis:info("Hello!")
end, "Mapping that starts with <Space>.")
```
```
␣               	Mapping that starts with <Space>.
```
In the code below, vis will try to replace a literal space at the start of a key with `␣`. Doing this, vis disregards all other characters in the key.
- https://github.com/martanne/vis/blob/1abf7eeea6b386c250249ab1ec6259bf0ccbb7c4/vis-cmds.c#L632
- https://github.com/martanne/vis/blob/1abf7eeea6b386c250249ab1ec6259bf0ccbb7c4/vis-cmds.c#L640

Is this expected behaviour? I would expect vis to replace all instances of a literal space (in the key) with `␣`.

PR #1111, opened 2023-07-08
Print keybindings starting with space correctly

MaxGyver83:
Bug fix for #1060.

### DIFF
--- a/vis-cmds.c
+++ b/vis-cmds.c
@@ -632,8 +632,26 @@ static bool cmd_earlier_later(Vis *vis, Win *win, Command *cmd, const char *argv
 	return pos != EPOS;
 }
 
+static size_t space_replace(char *dest, const char *src, size_t dlen) {
+	size_t i, invisiblebytes = 0, size = strlen("␣");
+	for (i = 0; *src && i < dlen; src++) {
+		if (*src == ' ' && i < dlen - size - 1) {
+			memcpy(&dest[i], "␣", size);
+			i += size;
+			invisiblebytes += size - 1;
+		} else {
+			dest[i] = *src;
+			i++;
+		}
+	}
+	dest[i] = '\0';
+	return invisiblebytes;
+}
+
 static bool print_keylayout(const char *key, void *value, void *data) {
-	return text_appendf(data, "  %-18s\t%s\n", key[0] == ' ' ? "␣" : key, (char*)value);
+	char buf[64];
+	size_t invisiblebytes = space_replace(buf, key, sizeof(buf));
+	return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, (char*)value);
 }
 
 static bool print_keybinding(const char *key, void *value, void *data) {
@@ -641,7 +659,9 @@ static bool print_keybinding(const char *key, void *value, void *data) {
 	const char *desc = binding->alias;
 	if (!desc && binding->action)
 		desc = VIS_HELP_USE(binding->action->help);
-	return text_appendf(data, "  %-18s\t%s\n", key[0] == ' ' ? "␣" : key, desc ? desc : "");
+	char buf[64];
+	size_t invisiblebytes = space_replace(buf, key, sizeof(buf));
+	return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, desc ? desc : "");
 }
 
 static void print_mode(Mode *mode, Text *txt) {


rnpnr:
Hi Max, thanks for the patch!

This fixes one of the issues listed in #1060 but doesn't replace any additional spaces with `␣`. Are you able to address that as well? Maybe a local 18 char buffer to copy `key` into while replacing ` ` with `␣`,

MaxGyver83:
Hi Randy, sure. I didn't think about this.

I have updated this pull request. I hope it's not too complicated. The main difficulty is that spaces (1 byte each) get replaced with `␣` which is 3 bytes but only 1 character. I have limited the number of space replacements to 3 because otherwise the destination string might grow in an unpredictable way.

rnpnr:
Definitely a little more complicated then I was expecting but its
partially because I forgot how to count bytes. It's on the right track
though so I'll leave some comments inline.

vis-cmds.c
```diff
@@ -632,16 +632,42 @@ static bool cmd_earlier_later(Vis *vis, Win *win, Command *cmd, const char *argv
 	return pos != EPOS;
 }
 
+static size_t space_replace(char *dest, const char *src, size_t maxlen) {
```
rnpnr, line 635: This has a lot of extra code. How about something like:
```c
static void replace_spaces(char *dest, const char *src, size_t dlen) {
	const size_t size = strlen("␣");
	for (size_t i = 0; *src && i < dlen; src++) {
		if (*src == ' ' && i < dlen - size - 1) {
			memcpy(&dest[i], "␣", size);
			i += size;
		} else {
			dest[i] = *src;
			i++;
		}
	}
	dest[i] = 0;
}
```

vis-cmds.c
```diff
@@ -632,16 +632,42 @@ static bool cmd_earlier_later(Vis *vis, Win *win, Command *cmd, const char *argv
...
+			to++;
+		}
+	}
+	*to = '\0';
+	return count;
+}
+
 static bool print_keylayout(const char *key, void *value, void *data) {
-	return text_appendf(data, "  %-18s\t%s\n", key[0] == ' ' ? "␣" : key, (char*)value);
+	char buf[25];
```
rnpnr, line 658: I would also be fine with `buf` being a little bit too long, say 64 bytes,
since it not hanging around anyways.

vis-cmds.c
```diff
@@ -632,16 +632,42 @@ static bool cmd_earlier_later(Vis *vis, Win *win, Command *cmd, const char *argv
...
+	size_t count = space_replace(buf, key, sizeof(buf));
+	return text_appendf(data, "  %-*s\t%s\n", 18+count*2, buf, (char*)value);
 }
 
 static bool print_keybinding(const char *key, void *value, void *data) {
 	KeyBinding *binding = value;
 	const char *desc = binding->alias;
 	if (!desc && binding->action)
 		desc = VIS_HELP_USE(binding->action->help);
-	return text_appendf(data, "  %-18s\t%s\n", key[0] == ' ' ? "␣" : key, desc ? desc : "");
```
rnpnr, line 644: The `%-18s` here just means that the string length that will be printed
is 18 characters. If key is shorter than 18 chars it will be padded
with spaces. Otherwise 18 chars of key are printed and the rest is
truncated. It can just stay as is.

rnpnr (changes requested):

vis-cmds.c
MaxGyver83: Unfortunately, this does not work. With a constant padding of 18, I get misaligned text, for example:

image

`[vsn]printf` with padding only counts bytes, not characters.

This can be verified with a minimal test:

```c
>> printf("%-4s|\n", " f");
 f  |
>> printf("%-4s|\n", "␣f");
␣f|
```

rnpnr: I see, next time I should test before I comment.

In that case how about returning the corrected width. It should just be
`18 + count * size`. I think that would be better than adding the magic
number `2` in the code.

If you can make those corrections I think I will be fine with the patch.

MaxGyver83: Sounds good to me. I'll change it like that.

MaxGyver83: Please check my last commit. Now I return `invisiblebytecount`. The `2` was indeed magic. But the `18` would be magic, too, when located inside `space_replace`. And I still set `'\0'` (instead of `0`) for terminating the string because this is used in many places in this file.

vis-cmds.c
```diff
@@ -632,42 +632,36 @@ static bool cmd_earlier_later(Vis *vis, Win *win, Command *cmd, const char *argv
...
-			count++;
-			destlen += size;
-			to += size;
+static size_t space_replace(char *dest, const char *src, size_t dlen) {
+	size_t i, invisiblebytecount = 0, size = strlen("␣");
+	for (i = 0; *src && i < dlen; src++) {
+		if (*src == ' ' && i < dlen - size - 1) {
+			memcpy(&dest[i], "␣", size);
+			i += size;
+			invisiblebytecount += 2;
```
rnpnr, line 641: The `2` here should just be `size - 1`. Also the variable could just be called `invisiblebytes`.

rnpnr (approved):
Besides the minor comment above this looks good to me. If you want to
squash the commits I would be OK with merging.

I will just wait a couple days to see if anyone else has any comments.

rnpnr:
Applied! Thanks for your work getting this patch in shape!

erf:
I get the following warning when building the latest version with this change:

```
In file included from sam.c:1813:
./vis-cmds.c:663:34: warning: field width should have type 'int', but argument has type 'unsigned long' [-Wformat]
        return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, (char*)value);
                                     ~~~^         ~~~~~~~~~~~~~~~~~
./vis-cmds.c:673:34: warning: field width should have type 'int', but argument has type 'unsigned long' [-Wformat]
        return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, desc ? desc : "");
```

MaxGyver83:
> I get the following warning when building the latest version with this change:
> 
> ```
> In file included from sam.c:1813:
> ./vis-cmds.c:663:34: warning: field width should have type 'int', but argument has type 'unsigned long' [-Wformat]
>         return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, (char*)value);
>                                      ~~~^         ~~~~~~~~~~~~~~~~~
> ./vis-cmds.c:673:34: warning: field width should have type 'int', but argument has type 'unsigned long' [-Wformat]
>         return text_appendf(data, "  %-*s\t%s\n", 18+invisiblebytes, buf, desc ? desc : "");
> ```

Oh, looks like I missed this. I'll change the type of `invisiblebytes` to `int`!

PR #1113, opened 2023-07-29
build: allow incremental rebuilds

rnpnr:
Its been annoying me for a while that every time I change a single source
file I have to rebuild all of them. Now we rebuild only the files that
were touched.

This should also make the ci marginally faster since the files can be
built in parallel.


TwoF1nger:
This doesn't detect changes in header files (both Vis own, and system ones).
You might want to use the `-MD` or `-MMD` gcc options.

I have a similar commit it in [my fork](https://repo.or.cz/vis/gkirilov.git) - [build: incremental compilation](https://repo.or.cz/vis/gkirilov.git/commit/d1e17fc6ba0a7bf6a31eeefdc20d6df42b2dab14)

rnpnr:
I guess I never really bother tracking headers. I'm not really against
it though. What do other people think? I've added it as a second commit
for now.

mcepl:
PR has been merged to https://git.sr.ht/~mcepl/vis and although only the Makefile part is relevant (and I have also added https://git.sr.ht/~mcepl/vis/commit/25fe086eb427 for build.sr.ht), it seems to be building and working just fine.

erf:
I think it happened after this commit that a bunch of `.d` and `.o` files are generated on `make`. Is it possible to avoid this as it clutters up the directory?

rnpnr:
> Is it possible to avoid this as it clutters up the directory?

Yes you are right, fixed here: 4ca7119.

rnpnr:
> fixed here: https://github.com/martanne/vis/commit/4ca711941423a9bdd206b3efb298aa9793682f10.

Oh oops I guess `|` in make rules is a GNU extension so I broke building
on Open/FreeBSD (macOS uses GNU make). I will have to do a separate fix
for that.

PR #1114, opened 2023-07-30
Lua API: access and set all available options

### LINKED ISSUE #803: Lua API: let plugins read the values of options
Plugins should be able to read the values of options from lua somehow. Currently options can be set with `vis:command`, but they cannot be read. I specifically want to be able to read `expandtab` and `tabwidth`, but I think a more generic mechanism is needed. It's worth noting that vis already exposes the `syntax` option with `vis.syntax` and `vis:set_syntax`.

There's a possible issue with a unified api, in that some options are tied to specific windows and other acting globally.

### LINKED ISSUE #812: Window layout property

rnpnr:
the first point of this commit is to allow all options to be read from lua. this has a number of uses for plugin writers. they are grouped into a couple of tables depending on what they control: `vis.options`: table with global configuration
`win.options`: table with window specific configuration

the second point is to allow you to set all these options as if they were simply lua variables. technically this is already possible by using `vis:command("set ...")` but personally I think this interface is cleaner. Note that this already possible for some things like the current mode (eg. vis.mode = vis.modes.VISUAL). examples: `vis.options.ai = true`
`vis.options.brk = " !?."`

there are a number of related issues and pull requests:
closes #803: Lua API: let plugins read the values of options
closes #812: Window layout property
supersedes/closes #717: Add ability to access tabwidth from Lua
supersedes/closes #1066: expose UI layout and allow it to be set from lua API

### DIFF
--- a/man/vis.1
+++ b/man/vis.1
@@ -1393,16 +1393,16 @@ section.
 .It Cm syntax Op Cm auto
 Syntax highlighting lexer to use, name without file extension.
 .
-.It Cm show-tabs Op Cm off
+.It Cm showtabs Op Cm off
 Whether to display replacement symbol instead of tabs.
 .
-.It Cm show-newlines Op Cm off
+.It Cm shownewlines Op Cm off
 Whether to display replacement symbol instead of newlines.
 .
-.It Cm show-spaces Op Cm off
+.It Cm showspaces Op Cm off
 Whether to display replacement symbol instead of blank cells.
 .
-.It Cm show-eof Op Cm on
+.It Cm showeof Op Cm on
 Whether to display replacement symbol for lines after the end of the file.
 .
 .It Cm savemethod Op Ar auto
--- a/sam.c
+++ b/sam.c
@@ -332,22 +332,22 @@ static const OptionDef options[] = {
 		VIS_HELP("Number of spaces to display (and insert if `expandtab` is enabled) for a tab")
 	},
 	[OPTION_SHOW_SPACES] = {
-		{ "show-spaces" },
+		{ "showspaces" },
 		VIS_OPTION_TYPE_BOOL|VIS_OPTION_NEED_WINDOW,
 		VIS_HELP("Display replacement symbol instead of a space")
 	},
 	[OPTION_SHOW_TABS] = {
-		{ "show-tabs" },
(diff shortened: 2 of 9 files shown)


rnpnr:
Note: I mostly made this patch to familize myself more with how the Lua-C
interface works. For instance this patch could have its size split in
half if people think that `vis:command("set ...")` is sufficient.

I do think that the ability to read the options from Lua is useful and
would like to merge at least that part.

ninewise:
I like this. It's much more ergonomic to have a lua api for these settings. Code looks good on quick read as well.

rnpnr:
In that case I can also add the ability to set the options with a table. For example:
```
vis.options = { expandtab = true, tabwidth = 4 }
```
It will be a very minor modification.

One other thing to note is that I had to rename some options to be valid
in lua (eg. `show-tabs -> showtabs`). I think I should probably rename the
options elsewhere to keep everything consistent.

rnpnr:
I implemented the ability to set options from a table, for example:

```
vis.options = { autoindent = true, expandtab = false }
```

This ended up being slightly more complicated than I originally thought
so I had to move the string parsing to separate functions so that it
could be reused. I am still quite happy with result.

Additionally the options I needed to rename were also changed in `sam.c`
so that everything is consistent.

While this is a big patch I think its overall a net positive. It will also
mean that there will be no more requests for exposing single options that
people need for writing plugins and that we won't need to make a breaking
API change when we get sick of all the options being assigned to the top
level `vis` and `win` tables (ie. `vis.tabwidth`, `vis.autoindent` etc).

mcepl:
I am afraid this doesn't apply at all on the top of #675.

rnpnr:
Its applied fine in my local testing branch. I think the problem is that
#675 needs to be squashed. I aim to apply that one and possibly this one
later tonight.

milhnl:
For those that want this behavior in their current version of vis, see https://github.com/milhnl/vis-options-backport

PR #1116, opened 2023-08-10
vis-lua.c: stop obj_ref_get() from leaving the lua stack modified

rnpnr:
The only place where this behaviour was encountered was in
`file_lines_iterator()` and it was just being worked around.

~~There was also likely a bug in the case that `addr` was `NULL` since 2
values would be popped from the stack instead of just 1. It is possible
that the case is never hit though.~~

Edit: Ignore that last part. It would hit the `lua_argerror()` branch in that case.


mcepl:
I have included it in https://git.sr.ht/~mcepl/vis and it seems to work just fine.

PR #1121, opened 2023-08-25
Added lua-5.4 to list of libluas used to find installed lua version, or

donaldcallen:
the latest if more than one installed. This change is specifically needed for FreeBSD.
modified:   configure


rnpnr:

PR #1125, opened 2023-09-04
vis-lua: raise an error when mapping to invalid handler types

devkarthin:
When passing an invalid handler type (such as nil) to Vis:map/Window:map, the editor would map the key to an empty (zeroed) KeyBinding that would cause itself to become unresponsive when processing the corresponding key.

### DIFF
--- a/vis-lua.c
+++ b/vis-lua.c
@@ -832,6 +832,8 @@ static int keymap(lua_State *L, Vis *vis, Win *win) {
 			goto err;
 	} else if (lua_isuserdata(L, 4)) {
 		binding->action = obj_ref_check(L, 4, VIS_LUA_TYPE_KEYACTION);
+	} else {
+		goto err;
 	}
 
 	if (win) {


mcepl:
Is there some way how to add a test for this?

devkarthin:
> Is there some way how to add a test for this?

Could you clarify what you mean by that?

mcepl:
> > Is there some way how to add a test for this?
> 
> Could you clarify what you mean by that?

You know we have the test suite in https://github.com/martanne/vis-test/, right?

vis-lua.c
```diff
@@ -832,6 +832,9 @@ static int keymap(lua_State *L, Vis *vis, Win *win) {
 			goto err;
 	} else if (lua_isuserdata(L, 4)) {
 		binding->action = obj_ref_check(L, 4, VIS_LUA_TYPE_KEYACTION);
+	} else {
+		vis_binding_free(vis, binding);
+		luaL_argerror(L, 4, "invalid handler");
```
rnpnr, line 837: I think this should just be `goto err;`. If a user ends up confused
as to why their mapping didn't work they can consult the documentation
and see that `vis:map()` and `win:map()` return a boolean indicating
success. We also have `vis:unmap()` for removing a binding so if
that was the goal it should just be called instead.

rnpnr (commented):
Hi Karthin,

Sorry I meant to reply after the build was fixed. Comments inline.

rnpnr:
Thanks for the patch!

PR #1126, opened 2023-09-05
text-io: fix block shrinking when the recent piece is on the block's end

vkochan:
Looks like the original idea was to shrink the block if the text have been deleted on the recent piece which is located at the block's end, so in that case it is enough just to resize the block without memmove.

### DIFF
--- a/text-io.c
+++ b/text-io.c
@@ -160,7 +160,7 @@ bool block_delete(Block *blk, size_t pos, size_t len) {
 	size_t end;
 	if (!addu(pos, len, &end) || end > blk->len)
 		return false;
-	if (blk->len == pos) {
+	if (blk->len == pos + len) {
 		blk->len -= len;
 		return true;
 	}


rnpnr:
So I've looked at this a few times and I still can't quite follow the
logic. Can you explain with a little bit more detail so that my slow
brain can understand it?

vkochan:
In my understanding the original idea was "when we are deleting a text from the recent (cached) piece which
sits on the edge of the block - then we can just shrink the block size without do a memmove". For example we have a
text formed from the following block and pieces:

| [piece1][piece2][piece3 ]| -- this is a block with pieces which holds the inserted chunks
| [Some ][text in][   block] | -- these are the chunks per each piece

so, for example we are deleting the "block" from the piece3 which are at the end of the block, then
we can just re-size piece3 and block sizes.

So, I think that the original logic does not work if for example the "pos" is at the start of the
"block" word and the "len" is length till the end of the block. The real use case might be deleting a text
from the end.

rnpnr:
Ok so I'm still not convinced. I've used this for a while and haven't
noticed any issues but that could just be because this is only in
the cache path.

The main thing confusing me is that the `blk->len` parameter is
used to mean two different things. I can assume this fix is correct
because the check above verifies that `pos + len` does not lie after
the end of the block. The next logical step would be to check if
`blk->len == pos +len`. But on the other hand `blk->len` apparently
also means the insertion position. If you follow the history all
the way back this check was added when the `blk->len` member was
called `blk->pos`. Its possible that this check was wrong then as
well but that was 7+ years ago.

A minor nit: `pos + len` is the value stored in the `end` variable.
That variable should be used in both places where `pos + len` is
being used.

Issue #1001, opened 2022-04-28
vis future?


Sneethe:
Last commit was over a year ago.

realwhz:
I am not sure if this discussion is a bit off-topic.  I started using vis since 2018.  It is never my major editor for programming but always part of my workflow.  I use vis whenever I have bunch of tedious and repetitive changes to make.  Its Sam regex based selection is so convenient.  For example, I can easily step by step build the selection and make the change.  There are more and more plugins developed for vis those days.  I seldom install any as I want to keep vis light and sharp.

erf:
There was a similar issue here #748 some time back . I guess the future is in the hands of the author or if anybody else would like to create a fork and step up to the plate .  I really enjoy **vis** as a light weight vim alternative and its Sam regex based selection and Lua API is great . I think it is an awesome alternative / competitor to other terminal based text editors like vim, nvim, kakoune etc. without the bloat of legacy code . What i would like to see next is the server / client architecture, or non-blocking IO as implemented here #675 which made it possible to implement the Language Server Protocol (LSP) in the issue here #853 . Another exciting feature is Tree-sitter discussed here  #668 for better built in lexers. There is also just the general maintenance - which takes a lot of time and the author does all the work pro bono (!) - so we can't really expect anything . Although some words regarding the status would be nice . I don't have much c skills, but have been enjoying playing around with the Lua API . I think **vis** will still be relevant in years to come <3

acidrums4:
I've been using vis as my main/only 'development' editor (alongside KDE's kwrite, which I use as a scratchpad) for a relative short time (2+ years?) and thus following the repo despite me not being a developer by any means. Though personally I'd disagree with the 'benevolent dictador for life' government model, I won't like vis to be forked - it'll mean a change in its direction and vision, and who knows, in this days people like things to be rewritten in Rust...

Despite at some points it seems not to, its main developer has a clear structure of what he's done and what he wants of it, which actually is one of the greatest perks (if not the greatest) of vis. But certainly the development process has slowed a lot since a few years. The server/client architecture discussion, which I too agree is what I'd love to have next, has been around and unresolved for several years. Even the discussion about the 'vis' name seems halted (#338).

One can only hope its main developer is doing fine and just taking some rest as the last time, but meanwhile it'd be great people could be submitting pull requests, giving ideas and reporting bugs, and even for an development illiterate like me, spreading the word of it so people can discover it and get involved.

deepcube:
[...]

Issue #668, opened 2018-02-10
Better built-in lexers


silversquirl:
The syntax highlighting lexers are currently missing support for various things (eg. labels in C, proper string highlighting for shell scripts, etc.)
I'm wondering whether it would be worth creating a wiki page with a list, so that people who want to contribute can look there.

martanne:
If you feel it would help, after all it is a Wiki. But I'm not sure how many people actually read these pages. For discussion an issue (like this one) is probably more helpful? Maybe we could add a general page regarding syntax highlighting (similar to the [README in `lua/lexers`](https://github.com/martanne/vis/tree/master/lua/lexers#lua-lpeg-lexers-for-vis)). It could then also feature/reference a list of known issues.

There are some known problems/limitations of the current approach to syntax highlighting (e.g. #110, #603). Eventually I would like to address these, but performance and accuracy needs to be weighted against the introduced complexity.

vktec:
If you think an issue is better, then I'm happy with that. I'll try to get acquainted with how LPeg works when I have some time and start working on the issues I mentioned initially.

p-e-w:
Have you considered [tree-sitter](https://github.com/tree-sitter/tree-sitter)?

While there is a lot of infrastructure around them, the parsers themselves are pure C, and designed for integration in text editors. Tree-sitter parsers are probably the highest quality syntax highlighting parsers available today, and are only going to get better because they are maintained by the creators of Atom.

I haven't seen any performance benchmarks yet, but I fully expect tree-sitter parsers to outperform all other incremental parsers of comparable quality eventually. Whether they are faster than vis' current lexers is hard to predict, but then those lexers are quite primitive and don't even come close to the accuracy that tree-sitter provides.

Having parsing of such quality in vis would also make new editing commands possible through accurate scope detection. See the examples in [this Atom PR](https://github.com/atom/atom/pull/16299).

Thus far tree-sitter parsers exist only for the most popular languages, but integrating those seems like a potentially huge win for vis: High-quality parsers that are maintained by someone else.

lanodan:
Disclaimer/Warning: All of this is my opinion and is As Far As I Know. I’m also quite bad at phrasing this kind of stuff but I don’t want to ignore things and I don’t want things to escalate.

* The ~advertised output is DOM-style, while we need something for Terminals
* Another language (C++, JSON, …)
* I’m not sure efficiency is that much needed
* There seems to be no release or tags !
[...]

PR #1127, opened 2023-09-08
Support old option names and don't deprecate them

hazen2215:
- Revert "support old option names but mark as deprecated"
- Support old option names and don't deprecate them

Changing option names is understandable because it may be useful for writing lua plugins but deprecating old names and adding new codes to detect deprecated options seems unnecessary to me.
vis supports alternative option names and the second commit of this pr just adds old names to alternatives.

I will squash two commits if desired.


rnpnr:
Why? If it was up to me I would delete all the aliases. These ones are
particularly useless in my view.

Also only 3 lines of code were added. The rest will be deleted after
0.9 is released.

hazen2215:
Noticed the deprecation message of 'show-eof' option when using vis on master branch.
Until most distros update vis to 0.9 and I can finally use vis.options.xxx, I'm stuck with :set command and seeing such message every time is quite annoying
Also I haven't see config breaking change like this on vis other than scintilua update, but that only broke local lexers and syncing visrc.lua among different versions of vis had previously no problems.

rnpnr:
If you are seeing that message you can use the supported name instead.

> Also I haven't see config breaking change like this on vis other than scintilua update

Hence why it is being marked as deprecated and supported for a
release. Those options were added before vis had a lua interface and so
the names were chosen without support for lua in mind.

hazen2215:
OK I found the plugin (https://github.com/milhnl/vis-options-backport) backporting vis.options to older versions that addresses renaming problem.
So if the maintainers want to I'm ok with closing this PR since the solution is found.
But also I'd like to hear others' opinion if consistency is good reason for deprecation and opinion about deprecating something that's been around for some time.

hazen2215:
also quote
https://github.com/martanne/vis/issues/1001#issuecomment-1116369265
>It will be a very slow, careful process. We don't want to make things worse, and we don't have the intricate knowledge of the internals that martanne had. Please do be patient as we learn the ropes.

TwoF1nger:
> But also I'd like to hear others' opinion if consistency is good reason for deprecation and opinion about deprecating something that's been around for some time.

vis is hardly a household name. I'd say break everything while you can and don't even bother with deprecation notices.

hazen2215:
Whether vis is "hardly" household name with 4K stars is debatable but vis is certainly well known to people who prefer simplicity or "suckless" or "KISS" or whatever
Lately vis is getting lots of commits by incorporating many old PRs since new maintainers are added and while getting the bugs fixed is nice I've seen some projects start making many breaking changes after BDFL(s) disappearance/handover (like anything.el/helm) and I'm concerned about vis going that way.

mcepl:
> Lately vis is getting lots of commits by incorporating many old PRs since new maintainers are added and while getting the bugs fixed is nice I've seen some projects start making many breaking changes after BDFL(s) disappearance/handover (like anything.el/helm) and I'm concerned about vis going that way.

Quite certainly we will make changes which martanne would not make, for one, we don’t have a crystal ball to read his mind. However, there are some reasons I don’t expect any radical changes in the direction different from what vis does right now:

1. We don’t have strength to do much. Current rather strong stream of changes is mostly just processing already existing and neglected in PR queue (there are some PRs created by @rnpnr lately, but these are mostly small fixes of existing issues), and I think we just cannot make any radical changes (which is one of unmentioned reasons behind my negative attitude towards suggested changes to treesitter in https://github.com/martanne/vis/issues/668). See how much time we need just with updating vis to follow upstream released highlighters from scintillua.
2. Given that there is nobody with sufficiently large amount of time, there is nobody who can claim to be BFDL, so there is nobody with large enough authority to decide about radical changes. Decision “to do whatever we were doing before” is usually the only one we can agree upon when dealing with any conflict (a possible conflict, I don’t think we had some great conflict so far).
3. I don’t actually see any reason why to make radical changes in vis. The idea of small C editor with large number of Lua plugins (which anybody can create and modify as they wish them to be) seems like a good design, and I don’t think there is anything which need to be changed.

rnpnr:
+1 to what @mcepl is saying

Since vis no longer has a "BDFL" the review and application of patches
has taken more of a community driven approach. If you are concerned
about the direction vis is going feel free to make your opinion heard
about any patches here or on the mailing list. Any commits that are not
minor are given some amount of time to sit and give people a chance to
comment. That includes major commits by me (the lua option patch sat
around for around a month before I applied it). I have been applying bug
fixes and minor things at a faster rate but they are still being filtered.

In any case this pull request doesn't fix any of this so I think it
should be closed. I don't think keeping such an alias around because
"thats how it used to be" is a good programming practice.

PR #1131, opened 2023-09-18
Add csproj and psm1 extensions to filetype

milhnl:
Hey all, just a small change, so not much to comment.

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -331,7 +331,7 @@ vis.ftdetect.filetypes = {
 		ext = { "%.pony$" },
 	},
 	powershell = {
-		ext = { "%.ps1$" },
+		ext = { "%.ps1$", "%.psm1$" },
 	},
 	prolog = {
 		ext = { "%.pl$", "%.pro$", "%.prolog$" },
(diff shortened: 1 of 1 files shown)


rnpnr (commented):
I'll say yes to the `.psm1` type but no to the `.csproj` type. The
latter seems to be auto-generated and mostly used by vscode. I don't
really think supporting every random auto-generated file from other text
editors makes any sense for vis.

You can easily add support for it locally in `visrc.lua`, just put this
somewhere after `require('vis')`:

```lua
table.insert(vis.ftdetect.filetypes.xml.ext, "%.csproj$")
```

milhnl:
I'm ok with not adding it, but I (and others working with dotnet projects) do need to edit it by hand, so if that makes you reconsider, great. The modification for my visrc you propose is already in there, but yours is more efficient. Thanks :)

mcepl:
On Mon Sep 18, 2023 at 7:16 PM CEST, Randy Palamar wrote:
> I'll say yes to the `.psm1` type but no to the `.csproj` type. The
> latter seems to be auto-generated and mostly used by vscode. I don't
> really think supporting every random auto-generated file from other text
> editors makes any sense for vis.

Just +1 for this decision from me.

-- 
https://matej.ceplovi.cz/blog/, Jabber: ***@***.***
 
I’ve never seen God worried yet.
  -- Wm. Paul Young

milhnl:
Sure. But the reasoning is still incorrect.

rnpnr:
I just pushed 7367ea8 which should also cover your csproj file.

Thanks for the patch!

milhnl:
Smart. Thanks. `file` says it's `text/plain` on my system but that's not vis' problem.

PR #1134, opened 2023-09-22
vis-subprocess.c: remove unnecessary NULL check

tosch42:


rnpnr (approved):
I definitely thought I fixed this when I applied it oh well.

Thanks for the patch!

PR #1135, opened 2023-09-25
vis-menu.c: add space for terminating NUL byte

tosch42:


mcepl:
Hmm, isn’t it cutting off the last character of `text`?

rnpnr:
Thanks for the patch Tom! I will apply it now.

vis-menu inherited a lot of rough code from slmenu and it really needs
a rework. This certainly fixes one issue.

> Hmm, isn’t it cutting off the last character of text?

No, the `sizeof` operator in C has no knowledge of the contents of the
buffer `text`. It simply returns the amount of static memory allocated
for it. If `sel->text` is as long or longer than `sizeof(text)` then
the string copied into `text` will not be terminated by `NUL` (aka `\0`,
aka `0`). `text` is `memset()` to `\0` at the start of this loop.

Hope this helps!

PR #1137, opened 2023-09-26
filetype.lua: add common *roff extensions

tosch42:
In case you're not familiar with *roff, `tmac` is the suffix for macro files and the others are the most popular macro packages. I didn't add preprocessors since Scintilla doesn't support them yet.


mcepl (approved):
What could go wrong here? Applied to the `devel` branch of https://git.sr.ht/~mcepl/vis

rnpnr (approved):
Yes there isn't any problems here.

Thanks for the patch Tom!

PR #1148, opened 2023-10-13
lua: refactor complete-filename plugin

### LINKED ISSUE #1146: Complete file name and file path swapped in doc
Doing `:help` gives me this:

```
<C-x><C-f>            Complete file path                                                                                                                                                                                             
<C-x><C-o>            Complete file name
```

Trying myself, it seems it is the opposite in fact:

```
<C-x><C-f>            Complete file name                                                                                                                                                                                             
<C-x><C-o>            Complete file path
```

I don't know if it is the implementation or doc that is wrong.

rnpnr:
There are probably more things to simplify but at least this makes it easier to see what exactly is different between `<C-x><C-f>` and `<C-x><C-o>`.

closes #1146

Comments on how to simplify this more would be appreciated!

### DIFF
--- a/lua/plugins/complete-filename.lua
+++ b/lua/plugins/complete-filename.lua
@@ -1,56 +1,51 @@
--- complete file path at primary selection location using vis-complete(1)
-
-vis:map(vis.modes.INSERT, "<C-x><C-f>", function()
+local complete_filename = function(expand)
 	local win = vis.win
 	local file = win.file
 	local pos = win.selection.pos
 	if not pos then return end
+
 	-- TODO do something clever here
 	local range = file:text_object_longword(pos > 0 and pos-1 or pos);
 	if not range then return end
 	if range.finish > pos then range.finish = pos end
-	if range.start == range.finish then return end
+
 	local prefix = file:content(range)
 	if not prefix then return end
-	-- Strip leading delimiters for some languages
-	local _, j = string.find(prefix, "[[(<'\"]+")
-	if j then prefix = prefix:sub(j + 1) end
-	local cmd = string.format("vis-complete --file '%s'", prefix:gsub("'", "'\\''"))
-	local status, out, err = vis:pipe(file, { start = 0, finish = 0 }, cmd)
-	if status ~= 0 or not out then
-		if err then vis:info(err) end
-		return
-	end
-	file:insert(pos, out)
-	win.selection.pos = pos + #out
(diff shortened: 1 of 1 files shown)


mcepl:
I have now rebased that patch by João on the top of this PR as https://git.sr.ht/~mcepl/vis/commit/09e25541cace.

rnpnr:
I made a couple changes to remove some differences between the two modes:
* whitespace in range is treated the same for both actions
* empty range will expand to files in PWD for both actions
For the second point `vis-complete` has some edge case that causes the name of the current directory to be added if invoked as `vis-complete --file ''` but that can be fixed separately.

rnpnr:
> I have now rebased that patch by João on the top of this PR as https://git.sr.ht/~mcepl/vis/commit/09e25541cace

Thinking about it again this doesn't belong in `complete-filename.lua` . It should be made in `vis-open` just like how `vis-complete` handles `~` separately.

mcepl:
> > I have now rebased that patch by João on the top of this PR as https://git.sr.ht/~mcepl/vis/commit/09e25541cace
> 
> Thinking about it again this doesn't belong in `complete-filename.lua` . It should be made in `vis-open` just like how `vis-complete` handles `~` separately.

Looking at the code:

```lua
        local cmdfmt = "vis-complete --file '%s'"
 	if expand then cmdfmt = "vis-open -- '%s'*" end
```

Shouldn’t be fixed in both?

rnpnr:
> Shouldn’t be fixed in both?

Sorry I mean in the shell scripts themselves. `vis-complete` already handles it. It just needs to be added to `vis-open`.

mcepl:
> > Shouldn’t be fixed in both?
> 
> Sorry I mean in the shell scripts themselves. `vis-complete` already handles it. It just needs to be added to `vis-open`.

Actually, can you reproduce it? Even without João’s commit, `vis-open ~/.bashrc` does what it should?

mcepl:
> because `~` isn't expanded before being passed to `vis-open`

And why should it matter when I have shown that `vis-open ~/filename` works as expected?

rnpnr:
> why should it matter when I have shown that `vis-open ~/filename` works as expected?

It doesn't if invoked inside vis. Open a new file and do `i~/<C-x><C-f>` and it will expand to your home directory. Now do `i~/<C-x><C-o>` and nothing will happen.

Issue #1147, opened 2023-10-12
`win:style` styles more terminal cells than expected


fischerling:
We noticed in https://github.com/fischerling/vis-lspc/pull/34 that the highlighting seems off around `\t` bytes.

This is reproducible by including something like

```lua
vis.events.subscribe(vis.events.WIN_HIGHLIGHT, function(win)
	win:style_define(43, 'back:red')
	win:style(43, 2, 2)
end)
```

in your `visrc.lua` file and opening a file containing only `\t\tx`.
Instead of highlighting the cell containing `x` (the third byte in the
file) the space after the first `\t` symbol up to and including the x is styled. 

Example Screenshot.
![20231012_19h16m02s_grim](https://github.com/martanne/vis/assets/1218430/e7e176a9-13e4-4789-9514-22274db97263)

rnpnr:
Hi, this patch seems correct. I have it applied locally and will likely apply it to master soon.

fischerling:
> Hi, this patch seems correct. I have it applied locally and will likely apply it to master soon.

What is the most convenient for you? Should I send it to the mailing list or open a PR here on Github?

rnpnr:
> What is the most convenient for you?

Either or is fine. Maybe more people will look at it if its a Github PR but its hard to say.

PR #1149, opened 2023-10-13
skip empty cells before applying a style

fischerling:
Fix empty cells erroneously being styled.

The view_style function is used to apply styles to ranges of text in a view.
It approaches the starting position where the style should be applied by
iterating the columns in the appropriate line using this while loop:

```C
while (pos < start && col < width)
	pos += line->cells[col++].len;
```

The while loop will stop at the last character before the range where the style
should be applied.

This works fine until we encounter "empty" cells between the last cell containing
an actual character and the first cell to be styled. This can happen if the last
character before the range to style is `'\t'` which gets expanded with empty cells
by vis according to the tabwidth option.
Those empty cells get erroneously styled as well.

This is fixed by skipping all empty cells encountered before the range to style.

Fixes: #1147.

### DIFF
--- a/view.c
+++ b/view.c
@@ -1428,6 +1428,10 @@ void view_style(View *view, enum UiStyle style_id, size_t start, size_t end) {
 	while (pos < start && col < width)
 		pos += line->cells[col++].len;
 
+	/* skip empty columns */
+	while (!line->cells[col].len && col < width)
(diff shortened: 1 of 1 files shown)


rnpnr:
Thanks for the patch! Applied in be89a64.

Issue #1103, opened 2023-06-11
terminal no longer has transparency/opacity


plasmoduck:
My terminal is normally transparent/opaque. When I open vis there is no longer transparency/opacity and it is a solid color background. How to keep the transparency/opacity?

rnpnr:
You will need to use a theme that just passes through the terminal background. Something like erf's [minimal clear theme](https://github.com/erf/vis-minimal-theme/blob/main/minimal-clear.lua) or my [term theme](https://github.com/rnpnr/dotfiles/blob/master/.config/vis/themes/term.lua) should work.

If there is a particular theme you like you can modify it to just not set the background.

plasmoduck:
I just use the default-16 theme cause it's the same as my existing terminal
colors (gruvbox), what do I have to change in default-16?

plasmoduck:
Sorry, dark-16.lua is what I'm using.

plasmoduck:
Thanks for your reply. I tried both the themes you included, unfortunatly they both seem to paint some background color, a blushish grey. Not sure what is happening?

esdnm:
I wish the vis came with themes that takes transparent background into account. Or a mechanism by which without modifying a themes file a transparent background can be obtained

rnpnr:
Patches are welcome!

I find transparency annoying in most cases but I will set it up to test any patches.

lobre:
I also think that an editor such as vis should come with a theme that does not define a solid background and that respects the background defined by the terminal. This is how vis looks like by default in my xfce-terminal.

![image](https://github.com/martanne/vis/assets/3764103/1814b6fe-b996-4092-8ed7-cc48684bb3d8)

It is one of the first things that newcomers see, and it is disruptive. Before even thinking about how vis works, it directly distracts me into trying to find a "sane" theme that looks like ok with my terminal colour palette.

ForestOctopus:
I downloaded Gruvbox color scheme by Pavel Pertsev (https://github.com/morhetz),which has this instruction in line 95:
```
-- To use your terminal's default background (e.g. for transparency), set the value below to 'back:default,fore:'..colors.fg1
```
So I set my line 96 as follow:
```
vis.lexers.STYLE_DEFAULT		= 'back:default,fore:'..colors.fg1
```
this make my suckless st terminal opacity function normaly.
I also find this method applie to default. In my case default-256.lua->zenburn.lua.
change  back to default
```
[...]

Issue #1040, opened 2022-09-01
Transparent background and setting options by default


swindlesmccoop:
I know in other ncurses programs that it's possible to implement this by adding
```
use_default_colors();
assume_default_colors(-1,-1);
```
to whatever file/function is responsible for handling drawing, however by adding the latter line under the former in `ui-terminal-curses.c`,  however it did not do anything. I have alpha set on my terminal already, and I would prefer that vis respects that and does not try and draw its own background. I'd prefer to be able to set this option myself in the source code.

Speaking of setting options in source code, is it possible? Say that I always want to show numbers, no matter what file I'm opening - is it possible to do that in source? I want to be able to fully configure it only in source without the need of any Lua.

clarkep:
You don't have to change the code, just use 'back:default' in your style definition. I don't know why there aren't any of the example styles that do this, but it is possible and documented in the Wiki.

clarkep:
Not sure why you want to avoid Lua but you can use 'back:default' in your style definition to avoid drawing a background.

SadielEd:
Hello I am using vis version 0.9 +curses +lua +acl in foot terminal and it respects my theme and transparency by default.
So maybe this issue can be closed.

![vis-screenshot](https://github.com/user-attachments/assets/dfdac7e8-391f-414e-97af-264d3da4cd75)

rnpnr:
Thanks for the info! I thought it should but didn't really set anything up for testing.

PR #1154, opened 2023-11-10
RFC: ui: refactor style handling

### LINKED ISSUE #1151: Theming is sometimes partially applied or ignored
It seems I cannot define the foreground color of a visual selection.

```
lexers.STYLE_SELECTION = 'fore:yellow'
```

This above does not work. Not sure if the problem is in this repo or in https://github.com/orbitalquark/scintillua/

rnpnr:
Hi everyone, this is a pretty major change to the way styles are handled inside vis. I would really appreciate testing to make sure this doesn't cause major breakage with existing themes. Most likely this will cause settings that were silently ignored before to start working.

Also I would appreciate some tests with the default style. I tested it in the Linux virtual console and it seems to work great there. I have not tested with a transparent terminal emulator so I would appreciate if someone could let me know if it respects that or not.

Commit message follows:

The old style handling had a lot edge cases where one of the colours or the attribute wouldn't get applied correctly. This commit adds a new style_set() method to the Ui which should be called instead of manually touching a cell's style. This also means that the Cell struct can be made opaque since all the handling is now done inside the ui-terminal files.

With this it is now viable to combine the light and dark 16 colour themes into a single base-16 theme. This theme works very well with the Linux virtual console and will now be the default theme regardless of if the terminal supports 256 colours or not.  This should address the common complaints about vis not respecting the users default terminal colours.

fixes #1151: Theming is sometimes partially applied or ignored
see #1103: terminal no longer has transparency/opacity
see #1040: Transparent background and setting options by default


lobre:
Nice work! Just tested it rapidly and it seems way better than before!

I would like to try creating a full 16-color simple theme with your branch, to see if I manage to have something meaningful. It could help identify problems here.

For now, the default theme in my terminal is not that great, especially the contrast in selections. If I manage to have something simple but efficient, maybe I could then submit it as default to have something great out of the box, that would generally work great in major terminal emulators (with default palettes). Let's see.

lobre:
Does it make sense to call it a 16-base while it only supports 8 colors? Isn't it the opportunity to bring the missing 8 bright variants?

mcepl:
It didn’t apply completely cleanly. After some work with merge tools I got 
[0001-ui-refactor-style-handling.patch](https://github.com/martanne/vis/files/13335673/0001-ui-refactor-style-handling.patch.txt)

mcepl:
On Fri Nov 10, 2023 at 6:13 PM CET, Randy Palamar wrote:
> Hi everyone, this is a pretty major change to the way styles
> are handled inside vis. I would really appreciate testing to
> make sure this doesn’t cause major breakage with existing
> themes. Most likely this will cause settings that were silently
> ignored before to start working.

Hmm, with my current setup and default theme I get dark yellow
cursor on the light yellow background, so the cursor is more or
less invisible.

Matěj

-- 
 
This conversation can serve no purpose anymore. Good bye.
    -- HAL9000 in 2001: Space Odyssea

mcepl:
> > Hmm, with my current setup and default theme I get dark yellow cursor on the light yellow background, so the cursor is more or less invisible.
> 
> Which theme is that? Some settings related to the cursor and selection used to be silently ignored so it could be an issue with the theme.

Light theme in my `foot` terminal and no theme in `~/.config/vis/visrc.lua` (which I guess means `default.lua` theme, right?).

rnpnr:
> Light theme in my `foot` terminal and no theme in `~/.config/vis/visrc.lua` (which I guess means `default.lua` theme, right?).

It looks like [foot](https://codeberg.org/dnkl/foot/src/branch/master/themes) has a handful of different light themes, some with poor contrast between regular and bright. I'm not sure we can get it right for all possible combinations but I do think the theme can be modified. Can you send a screenshot of your terminal without vis loaded (but with an active cursor) and with vis loaded? Maybe I can at least infer what should be changed.

mcepl:
> It looks like [foot](https://codeberg.org/dnkl/foot/src/branch/master/themes) has a handful of different light themes, some with poor contrast between regular and bright. I'm not sure we can get it right for all possible combinations but I do think the theme can be modified. Can you send a screenshot of your terminal without vis loaded (but with an active cursor) and with vis loaded? Maybe I can at least infer what should be changed.

Taking this back, I had some really [weird (most likely my own work) theme](https://paste.sr.ht/~mcepl/861a9acb885c506b33d923e109269f26664ab637), which is most likely buggy. Unfortunately, I cannot find anything I would like (I am a friend of Plan9/ACME-like very plain styles), but it is not fair to blame vis for it.

mcepl:
> > Light theme in my `foot` terminal and no theme in `~/.config/vis/visrc.lua` (which I guess means `default.lua` theme, right?).
> 
> It looks like [foot](https://codeberg.org/dnkl/foot/src/branch/master/themes) has a handful of different light themes, some with poor contrast between regular and bright. I'm not sure we can get it right for all possible combinations but I do think the theme can be modified. Can you send a screenshot of your terminal without vis loaded (but with an active cursor) and with vis loaded? Maybe I can at least infer what should be changed.

screenshot-2023-11-23_23-11-1700779388

(left is vis with no theme set, right is just plain foot with bash running; both with 
[acme](https://github.com/martanne/vis/files/13453727/acme.txt) foot style)

Actually, when using [vis-minimal-theme/minimal-clear](https://git.sr.ht/~mcepl/vis-minimal-theme/tree/devel/item/minimal-clear.lua), it is much better:

screenshot-2023-11-23_23-11-1700779891

I am not sure why the default theme fails so miserably, but at least I can work now.

mcepl:
To conclude my rambling:

Tested-by: Matěj Cepl <mcepl@cepl.eu>

rnpnr:
One other issue I noticed, also related to dimming, is that when you have the cursor somewhere within a dimmed section the line to the right of the cursor gets reset to the normal attribute. See attached: 
image

mcepl:
Couldn't we just merge this and #1133? I was running this for the last couple of weeks, and it seems to be working. I would hope that with merging these, we could get more testing and bigger chance to fix any remaining issues.

rnpnr:
I would be ok with that. I've been using both of them for as long has they have been open and while I know of a couple of bugs in both of them they don't get in the way of editing in any way. I've been too busy with other things to track down them down. Most of them (like the screenshots above) are probably quite trivial.

I'll hold off for a couple days for any other opinions but I think its time.

mcepl:
How long is “a couple of days” in your part of the world?

PR #1161, opened 2024-01-23
Update deprecated change-256colors in vis-std.lua

milhnl:
This caused a warning on startup when using Apple Terminal. I have no clue what this does and whether it is still necessary, but at least now it is silent.


rnpnr:
Thanks for the patch Michiel!

I plan on looking into that option after I fix the theme handling. I believe a few open issues are related to what that option does.

milhnl:
Welcome! Thanks for the merge, and let me know if you need someone to test something in Apple Terminal.

PR #1163, opened 2024-01-30
use correct apostrophe in error message

erf:
Use correct apostrophe in error message. E.g. when trying to open a directory.

BTW: I noticed that the error messages varies a bit, e.g. we use both "Can not", "Can't", "Could not", "Failed to". Perhaps these could be expressed in one way? I suggest "Can not" (as it is the most used). BTW "Cannot" is also a normally used word to express this.


rnpnr:
Applied, thanks!

Issue #1123, opened 2023-08-30
disabling status line


andreisou:
Is there any way to disable rendering of the status line via runtime configuration? If not maybe I'll try to implement this feature.

rnpnr:
Something like the following maybe:
```lua
vis.events.subscribe(vis.events.WIN_STATUS, function(win)
	win:status("")
end)
```

You also might want to modify your theme to not colour the status bar:

```lua
vis.lexers.STYLE_STATUS = ''
vis.lexers.STYLE_STATUS_FOCUSED = ''
```

See the [Lua API documentation](http://martanne.github.io/vis/doc/index.html) for more info.

andreisou:
That solution does stop rendering the status line contents, however, the status line itself is still present. I don't think there's a built-in way to disable the status line yet, so I might go ahead and implement that sometime.

PR #1168, opened 2024-02-01
allow disabling of statusbar

apprehensions:
implements #1123

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -292,6 +292,7 @@ enum {
 	OPTION_SHOW_TABS,
 	OPTION_SHOW_NEWLINES,
 	OPTION_SHOW_EOF,
+	OPTION_STATUSBAR,
 	OPTION_NUMBER,
 	OPTION_NUMBER_RELATIVE,
 	OPTION_CURSOR_LINE,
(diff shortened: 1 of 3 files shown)


rnpnr (commented):
I'm a bit on the fence about this one. For a future vis where there is no internal windowing I think this makes sense. As of right now its a bit clunky. For an example open two windows using `:split`. If the statusbar is off there is no way of distinguishing between the two windows.

I do like that the code is simple and I'm not strictly against the idea.

apprehensions:
> If the statusbar is off there is no way of distinguishing between the two windows.

The statusbar disablement is only valid for the selected window. If the user so chooses to disable it on the window, it will be gone, but not on all others.

mcepl:
I am not happy about window-managering vis either, but this seems like such a small thing, that I wouldn't bother. Yes, boiling of a frog, I know.

rnpnr (approved):
I thought about some more and unless someone else is extremely objected to this I think its fine. It's a useful feature if you only have 1 window open.

Hopefully we don't get too many reports complaining that they disabled the status bar with two or more windows open and now its not clear which window is focused or where the boundary is.

rnpnr:

PR #1170, opened 2024-02-27
Update Hare lexer

wackbyte:
Updates keywords to match the current syntax, sorts everything alphabetically, and renames "at rules" to "attributes" as they are named in the specification. Also styles `enum` and `union` as keywords rather than types to be consistent with `struct`, and the same for `void` because it has a dual purpose as both a type and a constant.


rnpnr (commented):
Thanks for the patch but can you send it to [upstream scintillua](https://github.com/orbitalquark/scintillua) instead? I will pull it into #1133 once its accepted in scintillua.

wackbyte:
Ah, okay

PR #1171, opened 2024-03-10
lua: filetype: add JSX and Typescript extensions

Stonks3141:


rnpnr:
Thanks for the patch!

PR #1172, opened 2024-03-13
Typst lexer

Disonantemus:
Bad, but better than nothing.

### DIFF
--- a/lua/lexers/typst.lua
+++ b/lua/lexers/typst.lua
@@ -1,0 +1,69 @@
+-- Copyright 2006-2021 Mitchell. See LICENSE.
+-- ? LPeg lexer.
+
+local lexer = require('lexer')
+local token, word_match = lexer.token, lexer.word_match
+local P, S = lpeg.P, lpeg.S
+
+local lex = lexer.new('?')
+
+lex:add_rule('variable', token(lexer.VARIABLE, lexer.to_eol('#')))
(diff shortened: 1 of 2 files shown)


mcepl:
Please, submit the new lexer to the upstream https://github.com/orbitalquark/scintillua, who maintains them (and who have much better experience with their design). We are just taking them from there.

rnpnr:
Exactly as mcepl says. I will happily merge it if its accepted there.

Its too much of a maintenance burden for us to manage changes to the lexers here. We already accidentally deleted local changes last time we synced with upstream.

mcepl:
On Thu Mar 14, 2024 at 2:37 PM CET, Randy Palamar wrote:
> Exactly as mcepl says. I will happily merge it if its accepted there.
>
> Its too much of a maintenance burden for us to manage changes to the lexers here. We already accidentally deleted local changes last time we synced with upstream.

That was slightly too fast, wasn’t it? Changes to plugins/filetype.lua should be accepted.

-- 
 
Every developer’s hunt for the best editor ends up with Vim,
Emacs or a management position.
  -- Rolf Bjaanes

rnpnr:
> That was slightly too fast, wasn’t it? Changes to plugins/filetype.lua should be accepted.

Oops you are right. I applied that part here 8fccd1b.

Issue #100, opened 2015-11-06
Command line completion with CTRL-D and TAB


erf:
Show a list of commands with CTRL-D
Command completion with TAB

martanne:
As stated elsewhere I would like to avoid it, if possible. The current approach is to have only few options/commands with aliases/unique prefixes such that completion is not needed. For file related tab completion the integration with `vis-open`/`slmenu` could probably be improved a bit.

This issue is a duplicate of #25 and #70.

erf:
What do you think about using <code>CTRL-j</code> and <code>CTRL-k</code> as aliases of left and right arrow when selecting file with slmenu?

martanne:
Yes this might be an option. However this can't be changed without forking slmenu which then raises the question whether vis should ship its own menu.

Also <kbd>Tab</kbd> in prompt mode could launch vis-open with the current prompt content followed by an asterisk as file pattern. That is `:edit vis`<kbd>Tab</kbd> would be expanded to something  like `ls -1 vis* 2>/dev/null | slmenu vis` where the argument to slmenu indicates the default/starting prompt value.

PR #1173, opened 2024-03-18
Command completion with tab

MaxGyver83:
In the command prompt, press `<tab>` to get a list of all available commands and pick one (using `vis-menu`). This works also after typing the first letters of a command (p.e. `:la<tab>`).

vis-command-completion

Implements #100 (more or less).

I'm aware that @martanne [wants to avoid this feature](https://github.com/martanne/vis/issues/100#issuecomment-154757305). But I want to share it anyway for those who want to apply this patch to their local vis build.


mcepl:
See also https://lists.sr.ht/~martanne/devel/patches/50426 for a discussion how to remove a warning caused by this patch.

mcepl:
I have a deep suspicion that 4970df09cea1e427b36469cc66340906cb5a0cb1 breaks this patch.

fischerling:
> I have a deep suspicion that [4970df0](https://github.com/martanne/vis/commit/4970df09cea1e427b36469cc66340906cb5a0cb1) breaks this patch.

It does indeed. But it is easily resolveable (a742dc367779dbf4a8703d4f59afe958cbc2eefe).

MaxGyver83:
> It does indeed. But it is easily resolveable ([a742dc3](https://github.com/martanne/vis/commit/a742dc367779dbf4a8703d4f59afe958cbc2eefe)).

Cool, thank you!

fischerling:
> > It does indeed. But it is easily resolveable ([a742dc3](https://github.com/martanne/vis/commit/a742dc367779dbf4a8703d4f59afe958cbc2eefe)).
> 
> Cool, thank you!

You are welcome. Please double check it. But I use it on my main branch for some time now.

MaxGyver83:
> You are welcome. Please double check it. But I use it on my main branch for some time now.

I have rebased and force-pushed it. It seems to work.

rnpnr:
I'm honestly on the fence about this. I know martanne wants to avoid this but the code is trivial and it aids in discoverability which in my opinion is one of the biggest marks of good piece of software (I don't mean via tutorials/help pages but those are nice too).

Of course there are more efficient ways of doing this, especially if vis stored the available commands in a [trie](https://en.wikipedia.org/wiki/Trie), but with the code that is already here this is pretty good.

mcepl:
> > > It does indeed. But it is easily resolveable ([a742dc3](https://github.com/martanne/vis/commit/a742dc367779dbf4a8703d4f59afe958cbc2eefe)).

This didn’t work for me (`:t<TAB>` does nothing), but I will try the latest commit yet.

machinedgod:
Just as a side-observer, would this get merged or it will forever remain an optional patch (because Marc didn't think its the right way forward 10y ago - not being sarcastic, 10y is a long time for software evolution and change of opinion)

mcepl:
@MaxGyver83 FYI, https://lists.sr.ht/~martanne/devel/patches/60055

mcepl:
@rnpnr What do you think about this PR? Should I merge it?

rnpnr:
> > I think if we want to merge it there should be a few revisions.
> > 
> > 2. Instead of passing around a stack variable and assuming it has a capacity of `VIS_COMMAND_BUFFER_MAX` I would prefer if a `Buffer` was used instead. Especially since the size of `prefix` is not being checked.
> 
> I still need to figure out how to do this. I guess I need two `Buffer`s to replace `cmd_list` and `cmd`!? Or should I use a single `Buffer`, append `echo '`, then all commands, and then `' | grep` and so on?

I was mostly referring to the one that gets passed through a void pointer into `print_cmd_name()`. For the `cmd` buffer I think you can use one on the stack, 32 or 64 bytes would do, and use `lua_tolstring` to get both the string and it's length. Then clamp the string's length to remainder of the buffer after subtracting the other characters. You can use:
```c
snprintf(cmd, sizeof(cmd), "grep '^%.*s' | vis-menu -b", (int)clamped_length, prefix);
```

I don't think you need the `echo` right? The `cmd_list` is piped directly into the `cmd` in `argv`.

> > By the way, were you also given push access to sourcehut?
> 
> Are you asking me? I don't think so.

Sorry I mostly meant @mcepl

mcepl:
> By the way, were you also given push access to sourcehut?

Yes, https://git.sr.ht/~martanne/vis/commit/bb94e22 is in.

PR #1177, opened 2024-04-06
vis_pipe*: take a Text object, vis-lua: add pipe_buffer

git-bruh:
This adds a pipe_buffer helper to the lua api to allow piping arbritary buffers to an external command, as the current way of constructing a `printf` command is quite hacky and is prone to shell injection

As a workaround, I'd written a wrapper like this but it messes up the scroll position, so made this PR instead:

```lua
vis.pipe_buffer = function(vis, buffer, cmd, fullscreen)
	vis:command("open")
	vis.win.file:insert(0, buffer)

	local status, stdout, stderr = vis:pipe(vis.win.file, {start = 0, finish = vis.win.file.size}, cmd, fullscreen)

	vis.win:close(true)

	return status, stdout, stderr
end
```

### DIFF
--- a/lua/plugins/complete-word.lua
+++ b/lua/plugins/complete-word.lua
@@ -24,8 +24,7 @@ vis:map(vis.modes.INSERT, "<C-n>", function()
 	if #candidates == 1 and candidates[1] == "\n" then return end
 	candidates = table.concat(candidates, "\n")
 
-	local cmd = "printf '" .. candidates .. "' | sort -u | vis-menu"
-	local status, out, err = vis:pipe(cmd)
+	local status, out, err = vis:pipe_buffer(candidates, "sort -u | vis-menu")
 	if status ~= 0 or not out then
 		if err then vis:info(err) end
 		return
--- a/sam.c
+++ b/sam.c
@@ -1767,7 +1767,7 @@ static bool cmd_filter(Vis *vis, Win *win, Command *cmd, const char *argv[], Sel
 	buffer_init(&bufout);
 	buffer_init(&buferr);
 
-	int status = vis_pipe(vis, win->file, range, &argv[1], &bufout, read_buffer, &buferr, read_buffer, false);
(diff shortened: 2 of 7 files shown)


rnpnr:
I'm not sure about this one. Can you provide some real example of where `printf` is creating a problem?

(Ignore the failing macOS CI, github broke something.)

git-bruh:
> I'm not sure about this one. Can you provide some real example of where `printf` is creating a problem?
> 
> (Ignore the failing macOS CI, github broke something.)

In my case I had to display a list in the format of `file:line: contents` with a fuzzy search program, and it broke in certain cases as file contents can have embedded quotes as well that might not be terminated, so the constructed command could look something like this, causing the command to fail
 
```sh
printf '
./file1:0:line zero
./file1:1:'unterminated quote
'
```

rnpnr:
Is there reason why you don't pipe the output of the fuzzy finder to the next stage in one go? Why do you need to save the intermediate output?

git-bruh:
The contents I shared are generated by my lua code, I'm using it for displaying all the lines for the goto references in LSP. I get the file and line numbers from the LSP server, and then construct a buffer in lua that is piped to a fuzzy finder

fischerling:
I wanted something like this too, but for different reasons.

Sometimes you do not have a window or a file to use `vis:pipe` with.

For example during the creation of the `vis-lockfile` plugin it would have been beneficial to be able to prompt for a user selection during a `FILE_OPEN` event (we solved it by subscribing to both `FILE_OPEN` and `WIN_OPEN`).
During the first `FILE_OPEN` event there does not exist a window or a file yet.

git-bruh:
Yeah we can definitely come up with more efficient ways of doing this, but for the PR just wanted to re-use as much of the existing code as possible

mcepl:
Sorry, just abusing this ticket for a test.

mcepl:
@git-bruh Hi, what is the status of this PR, please? Your last comment seems to indicate that there will be some new version of this PR coming.

fischerling:
I played with the idea and came up with #1204.
Any suggestions are welcome :)

rnpnr:
Superseded by #1204

Issue #1128, opened 2023-09-17
Customizing (a bit) the status line


acidrums4:
Hearing about #1123 made me realize something (maybe really silly for you) I miss from vim - customizing stuff in the status line.

This is how it currently looks like in several modes:

![1](https://github.com/martanne/vis/assets/513421/501cafa6-289f-4c3e-b781-282e50401a06)

Though not wanting a status line full of sparkles and stuff, something that I found nice and useful back in those days was that one could make it so each mode had its own (background) color. So what I wish is being able to make possible something like this:

![2](https://github.com/martanne/vis/assets/513421/0360a150-8a7a-464c-9e9c-d237de3a8240)

Of course that should work only in active views, where as in inactive views it could look something like this:

![3](https://github.com/martanne/vis/assets/513421/643cb162-afab-48ea-9e90-05fdb781bae1)

dther:
I decided to give this a try while messing with the UI code, and was surprised to discover that it was completely impossible in the current version of vis. That being said, if you're willing to compile your own vis, I've included a working example for how PR #1180 can be used to implement this!

rnpnr:
I will close this for now as it is achievable with the changes made in #1180.

PR #1180, opened 2024-04-21
RFC: Direct access to cell styling via API

dther:
This pull request essentially exposes the new style handling from PR #1154 to the Lua API, allowing plugins to make direct changes to cells using a new `Win` method, `Win.style_pos`.

It also adds a new style constant, `STYLE_LEXER_MAX`, which provides a reliable way of defining new user styles without needing to interact with the lexer.

Finally, a new event in `ui-terminal.c`, UI_DRAW, allows plugin developers the opportunity to override all previous styling. This allows for the creation of new UI elements entirely within the Lua API. This event is emitted at quite a low level (outside of vis.c and in ui_draw, right before the backend blit function), so I'd like opinions on whether it's even a good idea to have events outside of vis.c. In my opinion, it's a reasonably UI-specific event which has different ramifications based on the GUI toolkit being used, so aside from perhaps an extra argument to indicate more information to the user (i.e. "terminal/curses/ansi", "tk", "gnome", w/e). I don't foresee many changes to it.

By exposing a low-level styling function, many feature requests become implementable using the Lua API alone. For example, the following snippet can be placed into visrc.lua in order to implement status bar styling (#1128):
```lua
-- Example usage for style_pos: styling the status bar
local modes = {
	[vis.modes.NORMAL] = 'NORMAL',
	[vis.modes.OPERATOR_PENDING] = '???',
	[vis.modes.VISUAL] = 'VISUAL',
	[vis.modes.VISUAL_LINE] = 'VISUAL-LINE',
	[vis.modes.INSERT] = 'INSERT',
	[vis.modes.REPLACE] = 'REPLACE',
}

vis.events.subscribe(vis.events.WIN_STATUS, function(win)
	local left_parts = {}
	local right_parts = {}
	local file = win.file
	local selection = win.selection

	local mode = modes[vis.mode]
	if mode ~= '' and vis.win == win then
		table.insert(left_parts, mode)
	end
-- insert the rest of the code from vis-std.lua here...
	local left = ' ' .. table.concat(left_parts, " » ") .. ' '
	local right = ' ' .. table.concat(right_parts, " « ") .. ' '
	win:status(left, right);

	-- this must be done after calling win:status()
	-- mode change styling
	win.STYLE_RED = win.STYLE_LEXER_MAX - 1
	win.STYLE_GREEN = win.STYLE_LEXER_MAX - 2
	win.STYLE_YELLOW = win.STYLE_LEXER_MAX - 3
	win.STYLE_BLUE = win.STYLE_LEXER_MAX - 4
	win.STYLE_CYAN = win.STYLE_LEXER_MAX - 5
	win.STYLE_MAGENTA = win.STYLE_LEXER_MAX - 6
	win:style_define(win.STYLE_RED, "fore:red,reverse,bold")
	win:style_define(win.STYLE_GREEN, "fore:green,reverse,bold")
	win:style_define(win.STYLE_YELLOW, "fore:yellow,reverse,bold")
	win:style_define(win.STYLE_BLUE, "fore:blue,reverse,bold")
	win:style_define(win.STYLE_CYAN, "fore:cyan,reverse,bold")
	win:style_define(win.STYLE_MAGENTA, "fore:magenta,reverse,bold")

	local modestyle = {
		[vis.modes.NORMAL] = win.STYLE_BLUE,
		[vis.modes.OPERATOR_PENDING] = win.STYLE_YELLOW,
		[vis.modes.VISUAL] = win.STYLE_GREEN,
		[vis.modes.VISUAL_LINE] = win.STYLE_CYAN,
		[vis.modes.INSERT] = win.STYLE_RED,
		[vis.modes.REPLACE] = win.STYLE_MAGENTA,
	}

	local style = modestyle[vis.mode]
	if style and vis.win == win then
		for c=0,#modes[vis.mode]+1 do
			win:style_pos(style, c, win.height - 1)
		end
	end
end)
```

And UI_DRAW can be used, for example, to allow tighter integration with debugging tools by visually marking important lines:
```lua
-- UI_DRAW example usecase: highlighting important line numbers
-- using scraped info from cc or gdb or whatever
local importantfile = "vis.c"
local linesinview = {3, 23, 6, 15}

vis.events.subscribe(vis.events.UI_DRAW, function()
	local win = vis.win
	if not win then
		return
	end
	local file = win.file.name
	if not file or file ~= importantfile then
		return
	end
	if not win.STYLE_RED then
		return
	end

	-- highlight the line number and the rightmost column
	for line=1,#linesinview do
		if win.options.numbers then
			win:style_pos(win.STYLE_RED, 0, linesinview[line] - 1)
			win:style_pos(win.STYLE_RED, 1, linesinview[line] - 1)
		end
		win:style_pos(win.STYLE_RED, win.width - 1, linesinview[line] - 1)
	end
end)
```
And the end result:
image

### DIFF
--- a/lua/vis.lua
+++ b/lua/vis.lua
@@ -167,6 +167,7 @@ local events = {
 	WIN_STATUS = "Event::WIN_STATUS", -- see @{win_status}
 	TERM_CSI = "Event::TERM_CSI", -- see @{term_csi}
 	PROCESS_RESPONSE = "Event::PROCESS_RESPONSE", -- see @{process_response}
+	UI_DRAW = "Event::UI_DRAW", -- see @{ui_draw}
 }
 
 events.file_close = function(...) events.emit(events.FILE_CLOSE, ...) end
@@ -183,6 +184,7 @@ events.win_open = function(...) events.emit(events.WIN_OPEN, ...) end
 events.win_status = function(...) events.emit(events.WIN_STATUS, ...) end
 events.term_csi = function(...) events.emit(events.TERM_CSI, ...) end
 events.process_response = function(...) events.emit(events.PROCESS_RESPONSE, ...) end
+events.ui_draw = function(...) events.emit(events.UI_DRAW, ...) end
 
 local handlers = {}
 
--- a/main.c
+++ b/main.c
@@ -2229,6 +2229,7 @@ int main(int argc, char *argv[]) {
 		.win_highlight = vis_lua_win_highlight,
 		.win_status = vis_lua_win_status,
 		.term_csi = vis_lua_term_csi,
+		.ui_draw = vis_lua_ui_draw,
 	};
 
 	vis = vis_new(ui_term_new(), &event);
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -392,6 +392,7 @@ static void ui_draw(Ui *ui) {
 		ui_window_draw((UiWin*)win);
 	if (tui->info[0])
 		ui_draw_string(tui, 0, tui->height-1, tui->info, NULL, UI_STYLE_INFO);
+	vis_event_emit(tui->vis, VIS_EVENT_UI_DRAW);
 	ui_term_backend_blit(tui);
 }
 
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -69,6 +69,9 @@ struct UiTermWin {
 #include "ui-terminal-vt100.c"
 #endif
 
+/* helper macro for handling UiTerm.cells */
+#define CELL_AT_POS(UI, X, Y) (((UI)->cells) + (X) + ((Y) * (UI)->width));
+
 __attribute__((noreturn)) static void ui_die(Ui *ui, const char *msg, va_list ap) {
 	UiTerm *tui = (UiTerm*)ui;
 	ui_term_backend_free(tui);
@@ -305,6 +308,17 @@ static void ui_window_style_set(UiWin *w, Cell *cell, enum UiStyle id) {
 	memcpy(&cell->style, &set, sizeof(CellStyle));
 }
 
+static bool ui_window_style_set_pos(UiWin *w, int x, int y, enum UiStyle id) {
+	UiTermWin *win = (UiTermWin*)w;
+	UiTerm *tui = win->ui;
+	if (x < 0 || y < 0 || y >= win->height || x >= win->width) {
+		return false;
+	}
+	Cell *cell = CELL_AT_POS(tui, win->x + x, win->y + y)
+	ui_window_style_set(w, cell, id);
+	return true;
+}
+
 static void ui_window_status(UiWin *w, const char *status) {
 	UiTermWin *win = (UiTermWin*)w;
 	if (!(win->options & UI_OPTION_STATUSBAR))
@@ -534,6 +548,7 @@ static UiWin *ui_window_new(Ui *ui, Win *w, enum UiOption options) {
 
 	win->uiwin = (UiWin) {
 		.style_set = ui_window_style_set,
+		.style_set_pos = ui_window_style_set_pos,
 		.status = ui_window_status,
 		.options_set = ui_window_options_set,
 		.options_get = ui_window_options_get,
--- a/ui.h
+++ b/ui.h
@@ -97,6 +97,7 @@ struct Ui {
 
 struct UiWin {
 	void (*style_set)(UiWin*, Cell*, enum UiStyle);
+	bool (*style_set_pos)(UiWin*, int x, int y, enum UiStyle);
 	void (*status)(UiWin*, const char *txt);
 	void (*options_set)(UiWin*, enum UiOption);
 	enum UiOption (*options_get)(UiWin*);
--- a/vis-core.h
+++ b/vis-core.h
@@ -236,6 +236,7 @@ enum VisEvents {
 	VIS_EVENT_WIN_HIGHLIGHT,
 	VIS_EVENT_WIN_STATUS,
 	VIS_EVENT_TERM_CSI,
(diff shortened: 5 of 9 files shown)


rnpnr:
Very interesting! I definitely like this more than the line numbers styling patch from the wiki. This obviously has some footguns like the `WIN_HIGHLIGHT` event but I don't really think that's a problem.

I'm going to cherry pick the first two commits since they are unrelated, and leave you to fix the compilation issues. I also want to here what other people think!

dther:
Thanks for checking! It seems I missed a stub function in vis-lua.c. Should be fine now, I've tested it on my system with `./configure --disable-lua && make clean vis`.

fischerling:
> Very interesting! I definitely like this more than the line numbers styling patch from the wiki. This obviously has some footguns like the `WIN_HIGHLIGHT` event but I don't really think that's a problem.
> 
> I also want to hear what other people think!

I really like the simplicity and universal approach of this API.

However, the line numbers patch has the advantage that it allows to easily style lines not currently visible in the viewport.

In LSP land the server provides the information that in the file at line 200 there is a certain diagnostic.
vis-lspc can simply tell vis to highlight the line 200, and it will be styled if visible.
With this API vis-lspc needs to map the line number in the file to the viewport.

If anyone has a good idea how to efficiently map file line numbers to viewport line number I would gladly support this API in vis-lspc.

Maybe exporting the line range of the viewport to Lua could be a simple solution?

dther:
> If anyone has a good idea how to efficiently map file line numbers to viewport line number I would gladly support this API in vis-lspc.

This is something I've wanted for a *long time* too. It would greatly simplify a lot of the code I've written for the experimental mouse support, which relies heavily on translating visual lines to absolute byte position. Granted, I now plan to implement that logic directly in C instead, but I'm glad to see that I'm not the only one who sees value in exposing this data to the Lua API.

> Maybe exporting the line range of the viewport to Lua could be a simple solution?

I'm thinking of something along these lines as well. My initial thoughts are to output a simple table/list indexed by viewport lines, with member variables being the actual logical file lines, and perhaps special keys for "first" and "last", since this is what's exposed in `view.h`. In the future this might lead to a Lua class that exposes `struct View` with more power, but in my opinion it's best we keep things simple for the first prototype.

@rnpnr Should I add the commits for this proposed feature onto this PR? On the one hand, it's a function that's primarily useful due to the new UI_DRAW event, but on the other, I'd like this PR to be merged sooner rather than later, so that others can experiment with what's already possible.

rnpnr:
> My initial thoughts are to output a simple table/list indexed by viewport lines, with member variables being the actual logical file lines
I think we would want something simpler than that,  for example just a `Range` with the first and last line in view. We already have `File.lines` in lua which lets you get the lines contents by line number. On the C side the first and last line are already easily accessible via `win->view`.

> Should I add the commits for this proposed feature onto this PR?
It should be a separate submission. From a brief look at what is already available I don't really expect such a feature to need more than ~10 lines of code.

dther:
> I think we would want something simpler than that, for example just a `Range` with the first and last line in view. We already have `File.lines` in lua which lets you get the lines contents by line number. On the C side the first and last line are already easily accessible via `win->view`.

Apologies for the confusion- when I mean mapping visual lines to logical lines, I meant by line number, not content. While simply returning the first and last line numbers is much simpler than what I've implemented, I didn't have much trouble, and still believe that, exposing `lineno` values stored in `win->view->lines` should make for much simpler Lua code. I still shudder when I think about how large `guess_file_pos` (my lua function for translating terminal x, y coordinates to file lines and columns) ended up being.

rnpnr:
> Sorry, I guess I forgot that the viewport height and width isn't visible from lua right now, only the height/width of windows. Personally I would be ok if those were exposed separately.

Continuing this train of thought, I think long term the useful dimensions are the full UI width/height and the View width/height. With these two you can determine the sidebar width and status line height (if for example the info window is open). The window width and height which includes the sidebar and status line are only useful now while we allow separate windows.

dther:
As requested, I've moved the code for view height and width from #1181 over here. I imagine it would be best to make them fields within the new `viewport` table, so I plan to do that if no one objects- but I'll hold off on committing until #1181 is merged so that I can fast-forward cleanly.

dther:
Well, so much for a clean fast-forward. I'll rebase and force-push later.

rnpnr:
Nope, that is actually preferred. Github doesn't really play nice with it but everyone maintaining vis is used to the patch file/linear history approach.

dther:
> Nope, that is actually preferred. Github doesn't really play nice with it but everyone maintaining vis is used to the patch file/linear history approach.

Fair enough, I'll leave the merge as-is. The `git log` output looks funny on my end, but I'll live. @fischerling Sorry for changing up the field name just as you posted that...

fischerling:
> @fischerling Sorry for changing up the field name just as you posted that...

No problem I just updated the snippet ;)

rnpnr:
Alright I'm going to merge this, thanks for the patch!

There are a couple unrelated issue that this brings to light though:

 First, I was doing some experimenting to add custom lexer keywords ([here][] but it's very hacky) and I noticed that some lexers (eg. markdown) are very close to reaching LEXER_MAX.  So that probably needs to be increased.

Second it would be good to actually know from lua what the first available user style id is.

Third, as per your status line example, the function that fills in the status line (in `vis-std.lua`) should probably be made 
configurable so that it doesn't have to be run twice (i.e. store it as a variable in the vis table so that it can get reassigned to your custom function).

Also I'm thinking of a more robust version of `win:style_pos` that isn't limited to just the window. But that could be added later and `win:style_pos` would just do the bounds check and call that one.

[here]: https://github.com/rnpnr/dotfiles/blob/master/.config/vis/highlight.lua#L50-L62

PR #1185, opened 2024-04-30
release 0.9

rnpnr:
I tried to do a summary for the changelog instead of listing every last detail. Am I missing anything?

PR #1187, opened 2024-04-30
lua: complete-filename: use biggest possible prefix

fischerling:
This allows for example to complete file names in markdown images.

E.g.:    `![](pic<C-x><C-f>`    will complete the file name starting with pic.

Previously it would have detected the `'['` as prefix and looked for files starting with `](pic`.

Maybe I can sneek this in before #1185. This is sitting on top of my tree for quite some time now and I have not encountered any regression using the biggest possible match with our regex.

Apparently, my email setup is currently broken hence the github pull request.


rnpnr:
Looked fine to me, thanks!

PR #1188, opened 2024-05-03
configure: check for 'lua5.4-lpeg' too

mralusw:
Enables lpeg when building on Debian 12 with `apt install lua5.4 lua-lpeg`.


rnpnr:
Applied; thanks!

PR #1189, opened 2024-05-03
lua: lexers: use style `CODE_BLOCK` in favor of `CODE`

thimc:
Hi,

Congrats on a new release, here's a PR to fix code highlighting which I noticed while I was migrating from 0.8 to 0.9.

The lexer specs for gemini, markdown and txt2tags use a style named `CODE` to highlight some form of code (a code block in most situations) but the default theme (base-16) only sets the `CODE_BLOCK` style so I suppose `STYLE_CODE` is a bit redundant.


rnpnr:
Hi! Can you to add a style for `CODE` to the default theme instead? The lexers are vendored in from [scintillua](https://github.com/orbitalquark/scintillua) so we would prefer to not modify any of those files if it can be avoided.

rnpnr:
Applied in [a9a4f5c](https://github.com/martanne/vis/commit/a9a4f5cabb766e3dadd504df4fd9583192c22315). I swapped some things around because after looking at it again I think CODE_BLOCK will probably get removed if the ReST lexer is updated.

Thanks for the patch!

PR #1192, opened 2024-05-20
fix primary cursor color display

terminalzeiro:
Since the last release (v0.9), my theme configuration has stopped working: the color of the primary selection cursor no longer matches the ``lexer.STYLE_CURSOR_PRIMARY`` setting. 

This patch fixes this issue.

### DIFF
--- a/vis.c
+++ b/vis.c
@@ -398,7 +398,8 @@ static void window_draw_cursor(Win *win, Selection *cur) {
 	int col = view_cursors_cell_get(cur);
 	if (!line || col == -1)
 		return;
-	win->ui->style_set(win->ui, &line->cells[col], UI_STYLE_CURSOR);
+	Selection *primary = view_selections_primary_get(win->view);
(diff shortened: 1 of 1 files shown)


mcepl:
Could you please make it work with #1191 ?

rnpnr:
Hi! Can you give some example of the PRIMARY_CURSOR theming that is broken? Its possible that since I never use a different colour between cursors that it was missed.

> Could you please make it work with https://github.com/martanne/vis/pull/1191 ?

I could do that too. Its not that important.

wcsxfrm:
> Hi! Can you give some example of the PRIMARY_CURSOR theming that is broken? Its possible that since I never use a different colour between cursors that it was missed.

sure, here is a screenshot...

vis-with-without-patch-primary-cursor-color

rnpnr:
Yep, this seems correct so I applied it. Thanks!

PR #1202, opened 2024-08-25
Add lexer for the Odin programming language

brett-mahar:
(I've never done a pull request on github before, hope this is right. I based this Odin lexer on the Vis lexer for Golang)


brett-mahar:
Ok, I will submit it there :-)

halfbit0:
Scintillua seems to contain the odin lexer, but it isn't present in vis/lua/lexers

Nomarian:
Seems like it was added 3 months back, scintillua must be updated I guess

PR #1204, opened 2024-09-08
supporting piping a buffer to an external process

fischerling:
Alternative to #1177.

There are two major differences:
* No new lua function is introduced. `vis:pipe("string", cmd)` is used instead.
* No memory allocations are required.

`vis_pipe` is extended to support both file and range as well as buffer arguments.
Wrapper functions are provided to use the old behavior and the new one with a clean interface.


rnpnr:
Sound good to me! I won't have time to review this until Friday but it seems reasonable given that lua already has the string allocated.

Also thanks for adding a test! I need to remember to install `busted` in the github workflow so that the lua tests actually run.

fischerling:
> Also thanks for adding a test! I need to remember to install `busted` in the github workflow so that the lua tests actually run.

I added busted to the workflows in #1205.

mcepl:
BTW, does this fix also https://todo.sr.ht/~martanne/vis/5 ?

fischerling:
> BTW, does this fix also https://todo.sr.ht/~martanne/vis/5 ?

Probably yes.

PR #1207, opened 2024-09-10
Add support for C23 attributes

samuelmarquis:
Minor change to the `ansi_c` parser, adds highlighting support for every [c23 attribute](https://en.cppreference.com/w/c/language/attributes) as well as vendor-specific attributes of the form `[[a:b]]`


mcepl:
Please, send this PR to https://github.com/orbitalquark/scintillua/, which is where we get our lexers from.

PR #1210, opened 2024-09-23
lua: add mail lexer

eliobtl:
Adds mail lexer wich colors quotes in mails.


mcepl:
Also, I found somewhere this [mail.lua](https://github.com/user-attachments/files/17102164/mail.lua.txt), which seems infinitesimally better than yours. You may want to incorporate it.

eliobtl:
ok thanks, I'll do this

PR #1219, opened 2024-12-06
build: update alpine in docker build to version 3.21

eworm-de:
Just a version bump, no changes required.


rnpnr:
Applied Thanks!

Feel free to commit these directly yourself in the future. I don't think they really need review.

eworm-de:
Me neither. But we have copies of this repository that I do not have push access to, and I want to avoid making these out of sync.

rnpnr:
I think you mentioned this before my bad; I will try to remember next time.

Issue #1209, opened 2024-09-20
Syntax highlight on highlighted line


machinedgod:
### Problem

Hello,
since the lexer update, highlighted line doesn't have any syntax colouring on it any more. My memory might be playing tricks on me, but i can remember it being there with 0.7 or 0.8?

I've tested with several themes - various base16 ones, solarized and gruvbox, and its all the same.

### Steps to reproduce

1. `set cursorline`

### vis version (vis -v)

vis 0.9 +curses +lua +tre +acl

### Terminal name/version

rxvt-unicode 9.31

### $TERM environment variable

rxvt-unicode-256color

PR #1227, opened 2025-01-04
curses ui: fix default color detection

rnpnr:
closes: #1209


fischerling:
`Tested-By: Florian Fischer <florian.fischer@muhq.space>`

rnpnr:
This is the wrong fix; it breaks setting the background for line numbers (or perhaps this just exposed a different bug). I will need to fix that before applying this.

fischerling:
`Tested-By: Florian Fischer <florian.fischer@muhq.space>`

With 75892aad7939b40354be7ea71bd55c651956b27c both colored line numbers and syntax highlighting of the cursor line seems to work.

PR #1229, opened 2025-01-19
Change ui_window_style_set and change the way cursorline is drawn

infastin:
Changes:
- `ui_window_style_set` now accepts `keep` flag, that allows to not override non-default style values (fg and bg);
- `ui_window_style_set_pos` and `win_style` also accept the same flag;
- `win:style` and `win:style_pos` both accept the same `keep` flag, but it is optional;
- Now when drawing cursorline the CURSOR_LINE style won't override non-default values of other styles on the line.

This was mainly done so that CURSOR_LINE wouldn't override other styles, but plugins can potentially benefit from this as well.

### DIFF
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -276,8 +276,8 @@ void ui_window_style_set(Ui *tui, int win_id, Cell *cell, enum UiStyle id, bool 
 		return;
 	}
 
-	CellStyle default_style = tui->styles[win_id * UI_STYLE_MAX + UI_STYLE_DEFAULT];
 	if (keep_non_default) {
+		CellStyle default_style = tui->styles[win_id * UI_STYLE_MAX + UI_STYLE_DEFAULT];
 		if (!cell_color_equal(cell->style.fg, default_style.fg)) {
 			set.fg = cell->style.fg;
 		}
@@ -285,12 +285,9 @@ void ui_window_style_set(Ui *tui, int win_id, Cell *cell, enum UiStyle id, bool 
 			set.bg = cell->style.bg;
 		}
 	}
-	if (is_default_fg(set.fg)) {
-		set.fg = cell->style.fg;
-	}
-	if (is_default_bg(set.bg)) {
-		set.bg = cell->style.bg;
-	}
+
+	set.fg = is_default_fg(set.fg) ? cell->style.fg : set.fg;
+	set.bg = is_default_bg(set.bg) ? cell->style.bg : set.bg;
 	set.attr = cell->style.attr | set.attr;
 
 	cell->style = set;
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -278,19 +278,18 @@ void ui_window_style_set(Ui *tui, int win_id, Cell *cell, enum UiStyle id, bool 
 
 	CellStyle default_style = tui->styles[win_id * UI_STYLE_MAX + UI_STYLE_DEFAULT];
 	if (keep_non_default) {
-		if (!is_default_fg(cell->style.fg) && !cell_color_equal(cell->style.fg, default_style.fg)) {
+		if (!cell_color_equal(cell->style.fg, default_style.fg)) {
 			set.fg = cell->style.fg;
 		}
-		if (!is_default_bg(cell->style.bg) && !cell_color_equal(cell->style.bg, default_style.bg)) {
+		if (!cell_color_equal(cell->style.bg, default_style.bg)) {
 			set.bg = cell->style.bg;
 		}
-	} else {
-		if (is_default_fg(set.fg) || cell_color_equal(set.fg, default_style.fg)) {
-			set.fg = cell->style.fg;
-		}
-		if (is_default_bg(set.bg) || cell_color_equal(set.bg, default_style.bg)) {
-			set.bg = cell->style.bg;
-		}
+	}
+	if (is_default_fg(set.fg)) {
+		set.fg = cell->style.fg;
+	}
+	if (is_default_bg(set.bg)) {
+		set.bg = cell->style.bg;
 	}
 	set.attr = cell->style.attr | set.attr;
 
--- a/ui-terminal.c
+++ b/ui-terminal.c
@@ -40,12 +40,12 @@
 
 #define CELL_STYLE_DEFAULT (CellStyle){.fg = CELL_COLOR_DEFAULT, .bg = CELL_COLOR_DEFAULT, .attr = CELL_ATTR_NORMAL}
 
-static bool is_default_fg(CellColor c) {
-	return is_default_color(c);
+static bool is_default_fg(CellColor c, CellColor default_fg) {
+	return is_default_color(c) || cell_color_equal(c, default_fg);
 }
 
-static bool is_default_bg(CellColor c) {
-	return is_default_color(c);
+static bool is_default_bg(CellColor c, CellColor default_bg) {
+	return is_default_color(c) || cell_color_equal(c, default_bg);
 }
 
 void ui_die(Ui *tui, const char *msg, va_list ap) {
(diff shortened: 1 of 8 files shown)


rnpnr:
Hi! I'm not sure I understand the point? CURSOR_LINE already doesn't override non-default styles if the user didn't request it.  For example the [zenburn](https://github.com/martanne/vis/blob/master/lua/themes/zenburn.lua#L32) theme only specifies the background for CURSOR_LINE so whatever foreground was set by the other rules remains unchanged when CURSOR_LINE is applied.

infastin:
> Hi! I'm not sure I understand the point? CURSOR_LINE already doesn't override non-default styles if the user didn't request it. For example the [zenburn](https://github.com/martanne/vis/blob/master/lua/themes/zenburn.lua#L32) theme only specifies the background for CURSOR_LINE so whatever foreground was set by the other rules remains unchanged when CURSOR_LINE is applied.

If other styles on the same line has non-default background, then CURSOR_LINE will override them. That's is kind of the problem when using `vis-sneak` plugin, which sets CURSOR style on matches, which are then "overriden" by CURSOR_LINE when you are on the same line with the match.

<details>
Before:

image

After:

image
<details>

ui-terminal.c
```diff
@@ -269,25 +269,29 @@ static void ui_window_draw(Win *win) {
 	}
 }
 
-void ui_window_style_set(Ui *tui, int win_id, Cell *cell, enum UiStyle id) {
+void ui_window_style_set(Ui *tui, int win_id, Cell *cell, enum UiStyle id, bool keep) {
```
rnpnr, line 272: Please use a more descriptive name than `keep`. What is being kept? It should be immediately clear.

ui-terminal.c
```diff
@@ -269,25 +269,29 @@ static void ui_window_draw(Win *win) {
...
 	CellStyle set = tui->styles[win_id * UI_STYLE_MAX + id];
-
 	if (id != UI_STYLE_DEFAULT) {
-		set.fg = is_default_fg(set.fg)? cell->style.fg : set.fg;
-		set.bg = is_default_bg(set.bg)? cell->style.bg : set.bg;
+		CellStyle def = tui->styles[win_id * UI_STYLE_MAX + UI_STYLE_DEFAULT];
+		if (keep) {
+			set.fg = is_default_fg(cell->style.fg, def.fg) ? set.fg : cell->style.fg;
```
rnpnr, line 277:
```suggestion
			if (!is_default_fg(cell->style.fg) && !cell_color_equal(cell->style.fg, def.fg))
				set.fg = cell->style.fg;
```

These ternary operations are too difficult to follow with the proposed changes.

vis-lua.c
```diff
@@ -2038,7 +2039,8 @@ static int window_style(lua_State *L) {
 	enum UiStyle style = luaL_checkunsigned(L, 2);
 	size_t start = checkpos(L, 3);
 	size_t end = checkpos(L, 4);
-	win_style(win, style, start, end);
+	bool keep = lua_isboolean(L, 5) && lua_toboolean(L, 5);
```
rnpnr, line 2042:
```suggestion
	if (!lua_isboolean(L, 5))
		return luaL_argerror(L, 5, "boolean expected");
```
I think we should raise an error rather than assume false when the caller passes something weird.

vis-lua.c
```diff
@@ -2063,7 +2066,8 @@ static int window_style_pos(lua_State *L) {
 	enum UiStyle style = luaL_checkunsigned(L, 2);
 	size_t x = checkpos(L, 3);
 	size_t y = checkpos(L, 4);
-	bool ret = ui_window_style_set_pos(win, (int)x, (int)y, style);
+	bool keep = lua_isboolean(L, 5) && lua_toboolean(L, 5);
```
rnpnr, line 2069: Same as above.

rnpnr (changes requested):
Sorry for the delayed review. I'm fine with this if you address the comments I left inline.

ui-terminal.c
```diff
@@ -40,12 +40,12 @@
...
-	return is_default_color(c);
+static bool is_default_fg(CellColor c, CellColor default_fg) {
+	return is_default_color(c) || cell_color_equal(c, default_fg);
 }
 
-static bool is_default_bg(CellColor c) {
-	return is_default_color(c);
+static bool is_default_bg(CellColor c, CellColor default_bg) {
+	return is_default_color(c) || cell_color_equal(c, default_bg);
 }
```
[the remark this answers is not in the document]
infastin: Okay, reverted the change.

ui-terminal.c
infastin: Changed it to `keep_non_default`, but I don't know if this name is more descriptive or not.

vis-lua.c
infastin: All other functions use `lua_isboolean(L, 5) && lua_toboolean(L, 5)` approach.
If we raise an error in this function it would be appropriate to do the same in all of the others.

rnpnr:
Thanks for the updates, doesn't this break the zenburn theme with cursor line? Start vis with the zenburn theme, enable cursorline and place the cursor over a function prototype.

infastin:
> Thanks for the updates, doesn't this break the zenburn theme with cursor line? Start vis with the zenburn theme, enable cursorline and place the cursor over a function prototype.

Looks fine to me:
<details>

image

</details>

rnpnr:
> > Thanks for the updates, doesn't this break the zenburn theme with cursor line? Start vis with the zenburn theme, enable cursorline and place the cursor over a function prototype.
> 
> Looks fine to me:
> 
> image

Which terminal emulator are you using (`echo $TERM`)? I'm testing on vis master with this change applied and the following diff:
```diff
diff --git a/lua/vis-std.lua b/lua/vis-std.lua
index 1fe93a9..8285762 100644
--- a/lua/vis-std.lua
+++ b/lua/vis-std.lua
@@ -4,7 +4,7 @@ vis.events.subscribe(vis.events.INIT, function()
 	if os.getenv("TERM_PROGRAM") == "Apple_Terminal" then
 		vis:command("set change256colors false")
 	end
-	vis:command("set theme default")
+	vis:command("set theme zenburn")
 end)
 
 vis:option_register("theme", "string", function(name)
```
Then I'm running directly out of the main code directory: `./vis vis.c`.

infastin:
> > Looks fine to me: [...]
> 
> Which terminal emulator are you using (`echo $TERM`)? I'm testing on vis master with this change applied and the following diff:
> 
> ```diff
> diff --git a/lua/vis-std.lua b/lua/vis-std.lua
> index 1fe93a9..8285762 100644
> --- a/lua/vis-std.lua
> +++ b/lua/vis-std.lua
> @@ -4,7 +4,7 @@ vis.events.subscribe(vis.events.INIT, function()
>  	if os.getenv("TERM_PROGRAM") == "Apple_Terminal" then
>  		vis:command("set change256colors false")
>  	end
> -	vis:command("set theme default")
> +	vis:command("set theme zenburn")
>  end)
>  
>  vis:option_register("theme", "string", function(name)
> ```
> 
> Then I'm running directly out of the main code directory: `./vis vis.c`.

I've tested it on [st](https://st.suckless.org/) and xterm.

rnpnr:
I'm seeing the following in both st and xterm (this is an unconfigured xterm):
image

infastin:
At least I can see that it breaks solarized theme:
<details>

image

</details>

infastin:
I probably misunderstood what `ui_window_style_set` did originally. Works fine now.

rnpnr:
Squashed and applied in b7f4057 thanks!

I made a small tweak to not return early. This is less code and more readable IMO.

> I probably misunderstood what `ui_window_style_set` did originally. Works fine now.

The way the default style is being determined is really janky and that code is fragile. Every time I've touched it I've also broken it so its not really your fault. I've been intending to fix it but there are other parts of the code that I think need more attention.

infastin:
Thanks for your time!

PR #1240, opened 2025-04-23
vis-menu: support SIGWINCH resizing

youbitchoc:
leftover text wraps when resizing smaller, not sure how to fix that


mcepl:
Could we get back to the basics, please? What’s the actual bug this PR is trying to fix?

PR #1243, opened 2025-05-18
filetype: detect yash_profile and yashrc as shell scripts

unrealapex:


lua/plugins/filetype.lua
```diff
@@ -46,7 +46,7 @@ vis.ftdetect.filetypes = {
 	},
 	bash = {
 		utility = { "^[db]ash$", "^sh$","^t?csh$","^zsh$" },
-		ext = { "%.bash$", "%.csh$", "%.sh$", "%.zsh$" ,"^APKBUILD$", "%.ebuild$", "^.bashrc$", "^.bash_profile$" },
+		ext = { "%.bash$", "%.csh$", "%.sh$", "%.zsh$" ,"^APKBUILD$", "%.ebuild$", "^.bashrc$", "^.bash_profile$", "^.yashrc$", "^.yash_profile$" },
```
apprehensions, line 49: i would argue to change this to `^.*sh_profile$`, vice versa for .yashrc; and add `^.profile$`

lua/plugins/filetype.lua
```diff
@@ -46,7 +46,7 @@ vis.ftdetect.filetypes = {
 	},
 	bash = {
 		utility = { "^[db]ash$", "^sh$","^t?csh$","^zsh$" },
-		ext = { "%.bash$", "%.csh$", "%.sh$", "%.zsh$" ,"^APKBUILD$", "%.ebuild$", "^.bashrc$", "^.bash_profile$" },
+		ext = { "%.bash$", "%.csh$", "%.sh$", "%.zsh$" ,"^APKBUILD$", "%.ebuild$", "^.*shrc$", "^.sh_profile$" },
 		mime = { "text/x-shellscript", "application/x-shellscript" },
```
mcepl, line 50: And what about the good old `~/.profile`? Also, aren’t you missing `*` in last item? Shouldn’t it be `^.*sh_profile$`?

unrealapex: I think instead of using a wildcard, the extensions would be explicitly matched. I'll update them.

PR #1256, opened 2025-07-12
docs: remove $ from environment variable name

unrealapex:
Remove $ from environment variable name because environment variables are referenced by name without it. For example, in man pages.


mcepl:
Idiotic GitHub doesn’t recognise that I have already merged it manually as https://github.com/martanne/vis/commit/d8a26a9aee012f198038cbd2add99c1a945dd6b5. Useless.

PR #1260, opened 2025-08-18
Fix waycopy clipboard hanging 

runitclean:
A simple modification to fix waycopy hanging due to unclosed stderr fd (see [here](https://github.com/martanne/vis/issues/929))


ninewise:
That would be ideal, yes, but this can already prevent unneeded annoyance until the underlying issue is fixed. Applied, thanks!

PR #1263, opened 2025-09-01
make install: Copy from .CURDIR

falsifian:
Running `make install` on NetBSD currently gives

```
cp: vis: No such file or directory
cp: vis-menu: No such file or directory
cp: vis-digraph: No such file or directory
cp: vis-clipboard: No such file or directory
cp: vis-complete: No such file or directory
cp: vis-open: No such file or directory
```

(I configured with `--path=$HOME/sw/vis` in case that matters.)

Apparently this is because BSD make (at least on NetBSD and OpenBSD) runs commands from `$.OBJDIR`, which defaults to `./obj` if it exists, otherwise the current working directly. I don't know why this ever worked.

NB: I have not successfully tested this results in a working `vis` installation. I can confirm that `make install` installs stuff, but `vis` currently just hangs on NetBSD and I don't currently have time to investigate why.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -187,25 +187,25 @@ install: $(ELF)
 	@echo installing executable files to ${DESTDIR}${PREFIX}/bin
 	@mkdir -p ${DESTDIR}${PREFIX}/bin
 	@for e in ${EXECUTABLES}; do \
-		cp -f "$$e" ${DESTDIR}${PREFIX}/bin && \
+		cp -f "${.CURDIR}/$$e" ${DESTDIR}${PREFIX}/bin && \
 		chmod 755 ${DESTDIR}${PREFIX}/bin/"$$e"; \
 	done
 	@test ${CONFIG_LUA} -eq 0 || { \
 		echo installing support files to ${DESTDIR}${SHAREPREFIX}/vis; \
 		mkdir -p ${DESTDIR}${SHAREPREFIX}/vis; \
-		cp -r lua/* ${DESTDIR}${SHAREPREFIX}/vis; \
+		cp -r "${.CURDIR}"/lua/* ${DESTDIR}${SHAREPREFIX}/vis; \
 		rm -rf "${DESTDIR}${SHAREPREFIX}/vis/doc"; \
 	}
 	@echo installing documentation to ${DESTDIR}${DOCPREFIX}/vis
 	@mkdir -p ${DESTDIR}${DOCPREFIX}/vis
 	@for d in ${DOCUMENTATION}; do \
-		cp "$$d" ${DESTDIR}${DOCPREFIX}/vis && \
+		cp "${.CURDIR}/$$d" ${DESTDIR}${DOCPREFIX}/vis && \
 		chmod 644 "${DESTDIR}${DOCPREFIX}/vis/$$d"; \
 	done
 	@echo installing manual pages to ${DESTDIR}${MANPREFIX}/man1
 	@mkdir -p ${DESTDIR}${MANPREFIX}/man1
 	@for m in ${MANUALS}; do \
-		sed -e "s/VERSION/${VERSION}/" < "man/$$m" >  "${DESTDIR}${MANPREFIX}/man1/$$m" && \
+		sed -e "s/VERSION/${VERSION}/" < "${.CURDIR}/man/$$m" >  "${DESTDIR}${MANPREFIX}/man1/$$m" && \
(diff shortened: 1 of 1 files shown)


falsifian:
I should also mention I have not yet checked that `make install` is broken on OpenBSD, for example, or that this patch fixes it. But if I have time later I may do that.

mcepl:
>  Edit: oops, it actually does work. I was just accidentally running the base system's vis filter instead of the editor.

So, should this be closed or not?

Makefile
```diff
@@ -187,25 +187,25 @@ install: $(ELF)
 	@echo installing executable files to ${DESTDIR}${PREFIX}/bin
 	@mkdir -p ${DESTDIR}${PREFIX}/bin
 	@for e in ${EXECUTABLES}; do \
-		cp -f "$$e" ${DESTDIR}${PREFIX}/bin && \
+		cp -f "${.CURDIR}/$$e" ${DESTDIR}${PREFIX}/bin && \
```
mcepl: This doesn't make sense: how is this supposed to work on platforms which don’t set this weirdly named variable (and with all due respect to NetBSD, I am afraid, it is almost all of them)? Isn’t Makefile supposed to work as if `set -u` was set? Also, wouldn’t it be better to check whether `${.CURDIR}` is set and then just `cd` there?

falsifian: (For clarity: I mean `${.CURDIR}` is set on {Open,Net,Free}BSD's makes. Also I think the bug will affect all three of those platforms, when using base system `make` instead of GNU make, but I haven't tried.)

mcepl:
On Fri Nov 14, 2025 at 9:12 PM CET, James Cook wrote:
> I would be happy to try updating this PR to do that, but first,
> do you have any other make implementations in mind?

Looking at
https://github.com/martanne/vis/wiki/Distribution-Packages I
think we are really talking about various Linuxes (which I
suppose use all GNU Make), *BSD which use BSD Make, right? and
Mac OS X with Homebrew (what do they use? Gemini told me that
it is an ancient, and I mean ancient, from 2006, GNU Make, but
Homebrew installs something more recent for its packages).

Matěj

-- 
 
No matter what happens in the kitchen, never apologize.
  -- Julia Child

mcepl:
On Fri Nov 14, 2025 at 9:14 PM CET, James Cook wrote:
> (For clarity: I mean `${.CURDIR}` is set on
> {Open,Net,Free}BSD's makes. Also I think the bug will affect all
> three of those platforms, when using base system `make` instead
> of GNU make, but I haven't tried.)

Another thought which I have had on back burner for a long time
… do we really need two Makefiles? Couldn’t we somehow join them
together?

Best,

Matěj

-- 
 
Afraid to die alone?
Become a bus driver.
  -- alleged easter egg in notepad++

mcepl:
Also, what do you think about https://github.com/martanne/vis/pull/1270?

mcepl:
On Fri Nov 28, 2025 at 4:44 PM CET, James Cook wrote:
> falsifian left a comment (martanne/vis#1263)
>
>> Also, what do you think about #1270?

Ignore that reference, it was a moment of temporary insanity to even consider CMake. Of course, it is complete nonsense to even consider it (see my comments on #1270). If we want to ever move to anything more fancy (and I generally don’t), then we could think about `mk-configure`, but as long as we can live with nothing, I would stay with it.

Makefile
```diff
@@ -31,6 +31,10 @@ SRC = array.c \
 	$(REGEX_SRC)
 OBJ = $(SRC:%.c=obj/%.o)
 
+# On BSD make, set CURDIR to .CURDIR. On GNU make, CURDIR was already set in
+# GNUmakefile.
+CURDIR ?= $(.CURDIR)
```
mcepl: Shouldn’t it be `CURDIR ?= $(pwd -P)`?

GNUmakefile
```diff
@@ -1,5 +1,8 @@
 include Makefile
 
+# Part of a workaround for $(.CURDIR) being set in BSD make but not GNU make.
+CURDIR = .
```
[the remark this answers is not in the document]
falsifian: If it isn't set, I don't think the `install` rule in `Makefile` will work, since it uses the `$(CURDIR)` variable.

I'm not sure whether `.` or the value of `pwd` is better. As far as I know, both will have the same effect, but I don't know a lot about GNU make.

Issue #1198, opened 2024-07-26
Not all colors from the current theme are being used


acidrums4:
### Problem

There's something since 0.9 going on and it's bugging me out a bit: some (text) colors set in the current theme are not being used by vis.

I've noticed this specially with HTML files - for example, the HTML attributes are being painted with a color set for the shell/terminal emulator, green, rather than the colors set in the vis theme. No matter what theme is used, those are always being painted with the green color set for the shell/terminal emulator.

On the other hand, tabs are being painted with the default text color set in the vis theme rather than the color set by lexers.STYLE_WHITESPACE.

### Steps to reproduce

1. Use vis-0.9
2. Open an HTML file
3. Show tabs with `set showtabs`
4. Set whatever theme for vis, even a default theme like solarized or zenburn

### vis version (vis -v)

vis 0.9 +curses +lua +tre +acl

### Terminal name/version

konsole 24.05.2

### $TERM environment variable

xterm-256color

rnpnr:
I've known about this for a while but I haven't invested the time to write a proper fix. There is an old assumption here: https://github.com/martanne/vis/blob/2d87fdc0d088b12571b0c695549106b766585a92/ui.h#L37
that the number of lexer styles is small but in new scintillua that does not seem to be the case. The function for defining styles returns a bool status but that is never checked in the lua code. HTML probably has more styles than the maximum and so a lot of them are just not getting stored and failing silently.

(Below is an outline of what I think should be done, I'm leaving it here as a good first issue for someone else to solve).
`UI_STYLE_LEXER_MAX` should be increased and the lua code should check if it actually succeeds in defining the style. The maximum style id that gets defined should also be stored somewhere accessible from lua so that users/plugins that want to define their own styles know which slots are available. Also a test should be written that loops through the lexers and checks that they can successfully define all their styles (plus some headroom for users/plugin authors).

dcallea:
> I've known about this for a while but I haven't invested the time to write a proper fix. There is an old assumption here:
> 
> https://github.com/martanne/vis/blob/2d87fdc0d088b12571b0c695549106b766585a92/ui.h#L37
> 
> that the number of lexer styles is small but in new scintillua that does not seem to be the case. The function for defining styles returns a bool status but that is never checked in the lua code. HTML probably has more styles than the maximum and so a lot of them are just not getting stored and failing silently.
> 
[...]

PR #1266, opened 2025-09-19
Use `STYLE_WHITESPACE` for spaces, newlines, and tabs

lfdebrux:
I wanted to be able to change the colour of the replacement characters shown for showspaces, showtabs, and/or shownewlines, and was expecting `STYLE_WHITESPACE` to do this, but that didn't have any effect...

Issue #1198 suggested `STYLE_WHITESPACE` wasn't being used for anything, and I couldn't see it being used in any of the lexers that comes with vis, so this PR adds some lines to style the cells used for the replacement characters with the style in `STYLE_WHITESPACE`.


lfdebrux:
@mcepl ah, I see, you're creating a fork on sourcehut. I think in that case it makes most sense to keep the commits in this branch as they are and submit a different patch for your fork; I can put one together. Good luck with your fork!

PR #1285, opened 2025-12-02
doc/Makefile: add an empty clean target...

eworm-de:
... now that this is made with 12fc09a442939d0af09d700c7a8074cca9b1694e.

This unbreaks `make docker`, which triggered the catch-all target and failed.


fischerling:
We could also simply revert 12fc09a442939d0af09d700c7a8074cca9b1694e.
@mcepl What was the reason for this commit?

mcepl:
Merged as 1baf9352ae760f576f2ad213256649bdb266ddfb

PR #1289, opened 2025-12-09
show "pattern not found: foo"

eworm-de:


eworm-de:
Found this in an old local branch... 🙃

rnpnr (changes requested):
I think this is fine, but I guess the same should be done for `search_backward` below.  You can probably just make a separate `search_common` function that takes a `bool backward` and let the compiler do the inlining.

eworm-de:
> I think this is fine, but I guess the same should be done for `search_backward` below.

Indeed.

> You can probably just make a separate `search_common` function that takes a `bool backward` and let the compiler do the inlining.

Hmm, there are a lot of `*_forward()` and `*_backward()` functions. If we start this it should be de-duplicated for all functions, no?

PR #1290, opened 2025-12-14
Unity/Single File Build

rnpnr:
Hi all!

This is a big diff but nothing in here should cause any behavioural change of the program.

The motivation behind this change is that it allows the compiler to produce better code and it greatly simplifies the build process.

Notes:
* Optimized builds will be slightly larger now as many things can be inlined.
* `-Os` builds will be significantly smaller as the compiler can eliminate lots of useless stack management and other things.
* `Makefile`/`configure` were not removed yet but they were simplified.
* Most of the removed lines are in `main.c`, 3 parallel lists were combined into a [X-macro] list.

Commits are split up with extra info in each commit message. If you have any questions please ask!

As an example the following is now sufficient to build vis with this patchset:

```sh
#!/bin/sh

#CFLAGS="-O0 -gdwarf-4"
CFLAGS="-O3"

clang -march=native ${CFLAGS} -Wall -std=c99 \
      -DVIS_EXPORT=static \
      -DNCURSES_WIDECHAR -I/usr/include/ncursesw \
      -I/usr/include/lua5.3 \
      -DCONFIG_HELP=1 -DCONFIG_LUA=1 \
      -DCONFIG_CURSES=1 \
      -Wno-initializer-overrides \
      main.c -o vis -llua5.3 -lncursesw -ltinfow -ltermkey
```

[X-macro]: https://en.wikipedia.org/wiki/X_macro

### DIFF
--- a/text-objects.c
+++ b/text-objects.c
@@ -176,7 +176,7 @@ static bool text_line_blank(Text *txt, size_t pos) {
 }
 
 Filerange text_object_paragraph(Text *txt, size_t pos) {
-	char c;
+	char c = 0;
 	Filerange r;
 	if (text_line_blank(txt, pos)) {
 		Iterator it = text_iterator_get(txt, pos), rit = it;
@@ -314,6 +314,10 @@ Filerange text_object_indentation(Text *txt, size_t pos) {
 	size_t end = text_line_next(txt, bol);
 	size_t line_indent = sol - bol;
 	bool line_empty = text_byte_get(txt, bol, &c) && c == '\n';
+
+	/* NOTE: gcc spits a warning otherwise, this is probably not possible but something
+	 * about how text_line_{begin,start} are implemented makes gcc think otherwise */
+	line_indent = MIN(line_indent, PTRDIFF_MAX);
 
 	char *buf = text_bytes_alloc0(txt, bol, line_indent);
 	char *tmp = malloc(line_indent);
--- a/vis-operators.c
+++ b/vis-operators.c
@@ -99,7 +99,7 @@ static size_t op_put(Vis *vis, Text *txt, OperatorContext *c) {
 }
 
 static size_t op_shift_right(Vis *vis, Text *txt, OperatorContext *c) {
-	char spaces[9] = "         ";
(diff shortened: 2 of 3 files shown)


rnpnr:
> @rnpnr Why the single file build? Doesn't it make incremental builds impossible?

Yes thats the point. Incremental builds add a bunch of complexity for no good reason. The outcome with incremental builds is slightly faster recompilation if you have not touched something that is used in many places at the cost of worse codegen. 2 things: First, the majority of vis users are not recompiling vis over and over again. This is only something that we do as people working on this vis codebase. The users care about a fast (or small) application, they will experience the cost of the full build regardless of if it is single file or not.

Second, because vis was not designed with single file build in mind its actually much slower to build than it could be. With single file builds its very easy to count the lines of code being compiled:
```sh
$ cpp main.c | wc -l
25561
```
I have projects with 3x the amount of code, that compile about 13% faster than vis does, simply because I knew the code was compiling in one file so I made sure full function definitions always came before usage sites (no prototypes).

Here are the binary sizes of `vis` given various options. I did not include LTO for `-O3` builds as the optimizations are generally worse than what the compiler can do because the linker does not have the same information available to it. I also don't think LTO is being used by the majority of people packaging vis, so non LTO sizes are more representative of the current state.
```
-------------------
Unity Build:
-Os -march=native
gcc   15.2.1: 317.2K
clang 20.1.8: 324.5K

-Os -march=native -flto
gcc   15.2.1: 320.5K
clang 20.1.8: 322.9K

-O3 -march=native
gcc   15.2.1: 442.0K
clang 20.1.8: 433.0K
--------------------
master:
-Os -march=native
gcc   15.2.1: 344.2K
clang 20.1.8: 349.5K

-Os -march=native -flto
gcc   15.2.1: 317.8K
clang 20.1.8: 327.4K

-O3 -march=native
gcc   15.2.1: 438.0K
clang 20.1.8: 400.2K
```

> (I have an implementation in my vis copy over at repo.or.cz that I thought I could beat into shape and propose, but never got around to doing it.)

I'm not sure what you are referring to here. Vis without this patch has incremental builds.

PR #1293, opened 2025-12-22
Simplify and improve usability of jump list

rnpnr:
Hi!

For those who don't know there is set of cached selections in vis called the jumplist. It can be navigated with `<vis-jumplist-next>` and `<vis-jumplist-prev>`, bound to `g>` and `g<` by default, and saved to with `<vis-jumplist-save>`, bound to `gs` by default. It also gets saved to any time the cursor jumps.

Currently this is completely unusable because the editor mode is not being saved with the selections. The selections will get restored but there is no way to act on them so they just get removed on the next action. This fixes that issue.

Furthermore, from what I could tell the jumplist was supposed to be implemented as a least-recently-used (LRU) cache, obviously with a fixed size. However the code for doing this was vastly over complicated and the size was not enforced at all. I replaced the existing mess with a simple fixed size circular buffer. With this implementation all handling could be trivially collapsed into a single function.

I think this should behave the same as the existing implementation but since I wasn't using it there may be some edge cases I don't know about. I also think the cursor navigation feels a little janky but that part is the same as the existing code. I think its because other parts of the editor save to the jumplist in inappropriate places. Any suggestions there are welcome.

### DIFF
--- a/array.c
+++ b/array.c
@@ -141,21 +141,3 @@ void array_sort(Array *arr, int (*compar)(const void*, const void*)) {
 	if (arr->items)
 		qsort(arr->items, arr->len, arr->elem_size, compar);
 }
-
-bool array_push(Array *arr, void *item) {
-	return array_add(arr, item);
-}
-
-void *array_pop(Array *arr) {
-	void *item = array_peek(arr);
-	if (!item)
-		return NULL;
-	arr->len--;
-	return item;
-}
-
-void *array_peek(const Array *arr) {
-	if (arr->len == 0)
-		return NULL;
-	return array_get(arr, arr->len - 1);
-}
--- a/array.h
+++ b/array.h
@@ -101,26 +101,5 @@ VIS_INTERNAL bool array_resize(Array*, size_t length);
  * Sort array, the comparision function works as for `qsort(3)`.
  */
 VIS_INTERNAL void array_sort(Array*, int (*compar)(const void*, const void*));
-/**
- * Push item onto the top of the stack.
- * @rst
- * .. note:: Is equivalent to ``array_add(arr, item)``.
- * @endrst
- */
-VIS_INTERNAL bool array_push(Array*, void *item);
-/**
- * Get and remove item at the top of the stack.
- * @rst
- * .. warning:: The same ownership rules as for ``array_get`` apply.
- * @endrst
- */
-VIS_INTERNAL void *array_pop(Array*);
-/**
- * Get item at the top of the stack without removing it.
- * @rst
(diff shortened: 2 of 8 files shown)


machinedgod:
Hello,
I thought this would've been an appropriate place to ask, as I recall this PR.

I am not sure if its because of it, but it seems that the old functionality, where going backwards in the jump list doesn't work any more for previous selections.

It is possible I'm doing something wrong, but previously I was able to use `g<` to mimic `C-o` from Vim, ie. jumping back to the last cursor position, but I am working with the git version (commit c71a6c515655dda63a941c5706783dc54f442472) and its not doing it any more. If I hit `gs` then its all fine, but I don't always remember to do it, and I liked jumplist being populated automatically as well.

If this PR has at all touched that - was this an intended change?

rnpnr:
> It is possible I'm doing something wrong, but previously I was able to use `g<` to mimic `C-o` from Vim, ie. jumping back to the last cursor position, but I am working with the git version (commit [c71a6c5](https://github.com/martanne/vis/commit/c71a6c515655dda63a941c5706783dc54f442472)) and its not doing it any more.

Can you confirm exactly what the expected behaviour is? What motion did you use for a jump prior to `g<`? I have a completely vanilla install of vim with no configuration and `C-o` seems quite janky. If I jump via certain motions (eg. `{`) `C-o` will reverse those actions, but other jumps don't apply (eg. `10j`).

> If I hit `gs` then its all fine, but I don't always remember to do it, and I liked jumplist being populated automatically as well.

The intended behaviour in my view is that anytime there is a motion that could be considered a jump (basically anything that is not a single character motion) all selections should be pushed onto the jumplist. This seemed to be what the code was trying to achieve prior to this PR as well but it was doing it in a way that was completely useless in practice. This PR was meant to simply make it possible to save and restore selections to the jumplist.  As far as I remember I did not intentionally change anything related to when stuff gets pushed onto the jumplist. I definitely think vis needs some improvements around when the jumplist is pushed.

One thing I notice currently, which I thought was present before as well, is that you need to do `g<` twice to go back. That should be pretty trivial to fix (but its also not clear what should happen in the `g>` direction).

machinedgod:
Hey! My apologies for responding so late - I haven't gotten notified of your reply at all!

Everything you explained is my view too, and in last stable version I was using (I think it was 0.7), this worked exactly like that for me. Heck, I (ab)use that functionality for `vis-ultisnips`, to store anchor locations.

In all honesty I can't recall now if I just had a bit of a borked HEAD locally that I used, or I forgot what the difference was - but meanwhile I got used to it, so, all good from my side 👍

Thank you for the work!

PR #1299, opened 2026-01-07
Replace array.c with an intrusive implementation

rnpnr:
The existing dynamic array implementation in `array.c` has a number of key limitations. First it results in very poor performing code since any use needs to go through a layer of indirection to determine the element size. Second, it has no type safety whatsoever. There have been a couple of times where I wanted to refactor something and got completely lost because it was not obvious what the array was even holding. This version fixes those issues by using an intrusive data structure - the `da_*` helpers know that the structure they have a pointer to holds a `data`, a `count`, and a `capacity` member.

Key details:
* Full type safety. This is critical to ensuring correct code is being written, I cannot stress this enough. It is generally not a good idea to change C into a weakly typed language except for in a number of very specific circumstances (and in those cases type safety is ensured through stronger means than just the element size).
* Some DAs which were passed around got their own `*List` type and others which only exist in one place were just defined inline.
* Callers no longer check for successful resize. If a realloc ever fails, indicating we have drained the entirety of our massive virtual address space, the code `longjmp(3p)`s back to the start of the `vis_run()` function where we can attempt to save a backup of any open files (not yet implemented). With the way memory is just thrown around randomly in vis I highly doubt there was any hope of recovering from a failed push. In the future I hope to further reduce the amount of dynamic allocations which vis performs.
* Currently `count` and `capacity` have the type `VisDACount` which is a signed 32 bit integer. I don't think any of the dynamic arrays should ever be holding more than 2 billion elements, if they do that is a major bug. This way a standalone `*List` is 16 bytes which fits comfortably in a small vector register on any modern CPU (`xmm*` on x86, `v*` on ARM).
* Array tests were removed. The types of bugs they were checking for simply do not exist in this implementation.

I have been using this for a couple weeks already in various forms and I believe I fixed any bugs I introduced when trying to untangle the untyped mess. Please test and ask any questions you might have!


rnpnr:
Oh and I forgot:
* Coalesces a number of allocations that were scattered through memory into blocks. Operators, Motions, and TextObjects were previously stored as arrays of pointers meaning each one was stored in some disjoint location in memory. Now they are stored directly in their respective dynamic arrays which is much better for cache coherency.

machinedgod:
Fantastic under-the-hood upgrade, thank you for all this work!
I especially love that callers don't handle successful/unsuccessful resizes any more, and close second is cacheability improvements.

PR #1303, opened 2026-01-11
Makefile: docker: get date for reproducibility from file...

eworm-de:
... and drop the static date from 12 years ago.

This will change with every Alpine release, but resulting binary will be different anyway. 😜


eworm-de:
The idea here is to build a self-contained and statically linked executable that works everywhere. Well, as long as the architecture matches...

For the build environment we use *Alpine* in `docker`, which is small but ships all the dependencies required, including the static libraries. As *Alpine* is the only system supported here we should be fine using that file... Actually we could use any other one available.

PR #1326, opened 2026-03-22
number-inc-dec: fix wrong increment in a word

### LINKED ISSUE #1324: increment/decrement wrong number
### Problem

I use the *increment*/*decrement* functionality (`Ctrl`+`a` & `Ctrl`+`x`) a lot, often with version numbers. But it fails with specific versions.

Let's consider we have a version string "`1.2rc3`". I place the cursor above the "`3`" and press the shortcut to increment the number. I would expect to have "`1.2rc4`" - but this misbehaves: instead the cursor is moved forward, and the wrong number is incremented, the result is "`1.3rc3`".

Any chance to fix this? 

### Steps to reproduce

* type "`1.2rc3`"
* move the cursor to "`3`"
* press `Ctrl`+`a`

### vis version (vis -v)

vis v0.9-396-g2c27addf +curses +lua +tre +acl

### Terminal name/version

xfce4-terminal 1.2.0-1

### TERM environment variable

xterm-256color

fischerling:
Only the number under the cursor or the next number after the cursor should be changed.
However, if we unconditionally use the text object to detect were the number to increment is, words like `1rc3` could result in the number to the left of the curser to be changed.

This is fixed by only looking at the word under the cursor if the cursor is actually on a number.

Fixes #1324.

### DIFF
--- a/lua/plugins/number-inc-dec.lua
+++ b/lua/plugins/number-inc-dec.lua
@@ -18,17 +18,29 @@ local change = function(delta)
 	for selection in win:selections_iterator() do
 		local pos = selection.pos
 		if not pos then goto continue end
-		local word = file:text_object_word(pos);
-		if not word then goto continue end
-		local data = file:content(word.start, 1024)
+		local data = file:content(pos, 1024)
 		if not data then goto continue end
 		local s, e = pattern:match(data)
 		if not s then goto continue end
+		if s == 1 then
+			-- detect if the cursor is in the middle of a number
+			local word = file:text_object_word(pos)
+			-- the cursor is in the middle of a word
+			if word.start < pos then
+				local word_prefix = pos-word.start
+				data = file:content(word.start, word_prefix)..data
+				local e_old = e + word_prefix - 1
+				pos = pos - word_prefix
+				while e_old > e do
+					s, e = pattern:match(data, e)
+				end
+			end
+		end
 		data = string.sub(data, s, e-1)
 		if #data == 0 then goto continue end
 		-- align start and end for fileindex
-		s = word.start + s - 1
-		e = word.start + e - 1
+		s = pos + s - 1
+		e = pos + e - 1
 		local base, format, padding = 10, 'd', 0
 		if lexer.oct_num:match(data) then
 			base = 8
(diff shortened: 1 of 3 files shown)


eworm-de:
This looks mostly good, and fixes my use case. Thanks a lot!

Test some more I managed to break it, though. With "`1foo2bar3`" you can increment and decrement the first and second number, but the code fails on the third one with:
```
lua/plugins/number-inc-dec.lua:36: attempt to compare nil with number
```

fischerling:
Thanks for testing!
The first version of the algorithm was brittle and stupid. Using the `init` parameter of `lepg.match(data, pattern, [init])` solves the bookkeeping issue what parts of the word we already searched.

eworm-de:
This one is fine. 👍 Again, thanks a lot!

fischerling:
> This one is fine. 👍 Again, thanks a lot!

You are welcome :)

Do you have other interesting test cases worth adding to the tests?

eworm-de:
No, that looks very similar to what I typed into `vis` for testing. 😉

fischerling:
@rnpnr What do you think?
If you think it is ok I would apply this.

rnpnr (approved):
I don't see any issues here. Thanks for the test as well!

fischerling:
applied as da84fe700880eb988ea0819f9c823547287f28a9.

eworm-de:
Damn, this still misbehaves. 🫣 Sorry I did not notice before.

I we have a number with two or more digits the position for cursor is important. Let's assume we have "`11`", depending on cursor position the number is incremented to "`12`" or "`21`".

fischerling:
> Damn, this still misbehaves. 🫣 Sorry I did not notice before.
> 
> I we have a number with two or more digits the position for cursor is important. Let's assume we have "`11`", depending on cursor position the number is incremented to "`12`" or "`21`".

Sorry about that. Should be fixed by 13c577a1b3802562dc43235a5a4a8e5acdd331a9.

eworm-de:
Ok, found even more... A "`19`" can increment to "`20`" or "`110`"...

fischerling:
> Ok, found even more... A "`19`" can increment to "`20`" or "`110`"...

Damit! You are right. I will make a step back and will only push changes that are properly tested.

fischerling:
Sorry for all that noice and my rushed changes!
Not sure why I can't reopen/reuse this pull-request I will open a follow-up.

PR #1329, opened 2026-04-03
Fix UI lag: revert markdown lexer to 0208d7a (Issue #1314)

### LINKED ISSUE #1314: 100% cpu spikes and slow response with md syntax on
### Problem

With some markdown files and syntax on, moving through the file gets _**really-really slow**_.

### Config

Minimal `visrc.lua`, default file with no plugins.

### Workaround

- Disable syntax highlighting: `:set syntax off`.
- Process the file with a markdown format checker/fixer, example: `:%|mdformat -` or `:|pandoc -t gfm`.

---

### System

> distro: Arch Linux x86_64  
> kernel: 6.12.33-1-lts  
> shell: bash 5.3.9  
> term: tmux 3.6a

### Steps to reproduce

Compile last git and run `vis`:

~~~ocaml
git clone https://github.com/martanne/vis
cd vis/
./configure
make
sudo make install
vis p3_slow.md
~~~

When `vis` is launched:

1. launch a monitor of system resources, like: `btop`.
1. press and hold `j` to move down (line by line) until the end of the file; watch for CPU spikes, including 100% in a core, and slow redraw.
2. press `:set syntax off`; watch for slow redraw/reaction for key presses.
3. press `Enter` key to change syntax and `gg` to go to start of file.
4. repeat from `1.`; but now you get fast/instant redraw as usual/expected.

[p3_slow.md](https://github.com/user-attachments/files/25218432/p3_slow.md)

### vis version (vis -v)

vis v0.9-372-gc71a6c5 +curses +lua +tre +acl

### Terminal name/version

foot version: 1.25.0 +pgo +ime +graphemes -assertions

### TERM environment variable

screen-256color

4d30:
### Title
Fix UI lag in markdown lexer by reverting to 0208d7a (Fixes #1314)

### Description
Upon investigation, it was determined that changes made to the markdown lexer `/lua/lexers/markdown.lua` after commit 0208d7a introduced nested loops during the highlighting process. This increased lexing complexity, causing the UI thread to hang or stutter when scrolling through or editing markdown files.

### Proposed Change
- Revert `lua/lexers/markdown.lua` to 0208d7a

### DIFF
--- a/lua/lexers/markdown.lua
+++ b/lua/lexers/markdown.lua
@@ -29,54 +29,22 @@ lex:add_rule('list', lex:tag(lexer.LIST,
 local hspace = lexer.space - '\n'
 local blank_line = '\n' * hspace^0 * ('\n' + P(-1))
 
-local code_line = lexer.starts_line((B('    ') + B('\t')) * lpeg.P(function(input, index)
-	-- Backtrack to the start of the current paragraph, which is either after a blank line,
-	-- at the start of a higher level of indentation, or at the start of the buffer.
-	local line, blank_line = lexer.line_from_position(index), false
-	while line > 0 do
-		local s, e = lexer.line_start[line], lexer.line_end[line]
-		blank_line = s == e or lexer.text_range(s, e - s + 1):find('^%s+$')
-		if blank_line then break end
-		local indent_amount = lexer.indent_amount[line]
-		line = line - 1
-		if line > 0 and lexer.indent_amount[line] > indent_amount then break end
-	end
-
-	-- If the start of the paragraph does not being with a '    ' or '\t', then this line
-	-- is a continuation of the current paragraph, not a code block.
-	local text = lexer.text_range(lexer.line_start[line + 1], 4)
-	if not text:find('^\t') and text ~= '    ' then return false end
-
-	-- If the current paragraph is a code block, then so is this line.
-	if line <= 1 then return true end
-
-	-- Backtrack to see if this line is in a list item. If so, it is not a code block.
-	while line > 1 do
-		line = line - 1
-		local s, e = lexer.line_start[line], lexer.line_end[line]
-		local blank = s == e or lexer.text_range(s, e - s + 1):find('^%s+$')
-		if not blank and lexer.indent_amount[line] == 0 then break end
-	end
-	text = lexer.text_range(lexer.line_start[line], 8) -- note: only 2 is needed for unordered lists
-	if text:find('^[*+-][ \t]') then return false end
-	if text:find('^%d+%.[ \t]') then return false end
-
-	return true -- if all else fails, it is probably a code block
-end) * lexer.to_eol(), true)
-
-local code_block = lexer.range(lexer.starts_line('```', true),
-	'\n' * hspace^0 * '```' * hspace^0 * ('\n' + P(-1))) +
-	lexer.range(lexer.starts_line('~~~', true), '\n' * hspace^0 * '~~~' * hspace^0 * ('\n' + P(-1)))
-
+local code_line = lexer.starts_line((B('    ') + B('\t')) * lexer.to_eol(), true)
+local code_block =
+	lexer.range(lexer.starts_line('```', true), '\n```' * hspace^0 * ('\n' + P(-1))) +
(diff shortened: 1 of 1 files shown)


lua/lexers/markdown.lua
```diff
@@ -29,54 +29,22 @@ lex:add_rule('list', lex:tag(lexer.LIST,
...
 end)
-
 lex:add_rule('block_code', lex:tag(lexer.CODE, code_line + code_block + code_inline))
 
-lex:add_rule('blockquote', lex:tag(lexer.STRING, lexer.starts_line('>', true)))
+lex:add_rule('blockquote',
+	lex:tag(lexer.STRING, lpeg.Cmt(lexer.starts_line('>', true), function(input, index)
+		local _, e = input:find('\n[ \t]*\r?\n', index) -- the next blank line (possibly with indentation)
+		return (e or #input) + 1
+	end)))
```
rnpnr, line 47: This can only perform worse. It does more than the existing line.

lua/lexers/markdown.lua
```diff
@@ -124,28 +92,4 @@ local start_rule = lexer.starts_line(P(' ')^-3) * #P('<') * html:get_rule('tag')
 local end_rule = #blank_line * ws
 lex:embed(html, start_rule, end_rule)
 
-local FOLD_HEADER, FOLD_BASE = lexer.FOLD_HEADER, lexer.FOLD_BASE
--- Fold '#' headers.
-function lex:fold(text, start_line, start_level)
```
rnpnr, line 129: We don't support folding so there is no reason to diverge from upstream here.

lua/lexers/markdown.lua
```diff
@@ -29,54 +29,22 @@ lex:add_rule('list', lex:tag(lexer.LIST,
...
-	if text:find('^%d+%.[ \t]') then return false end
-
-	return true -- if all else fails, it is probably a code block
-end) * lexer.to_eol(), true)
-
-local code_block = lexer.range(lexer.starts_line('```', true),
-	'\n' * hspace^0 * '```' * hspace^0 * ('\n' + P(-1))) +
-	lexer.range(lexer.starts_line('~~~', true), '\n' * hspace^0 * '~~~' * hspace^0 * ('\n' + P(-1)))
-
+local code_line = lexer.starts_line((B('    ') + B('\t')) * lexer.to_eol(), true)
```
rnpnr, line 32: This seems to be the only relevant change.

rnpnr (commented):
There are certainly other things we could be doing so that this is not necessary. I will discuss the code line change with upstream though. 

Interestingly when testing I couldn't recreate the issue because my installed vis uses a statically linked LPEG. It seems that embedding LPEG directly in vis provides a much more significant performance boost than I would normally expect. It may be worth making that the default when lua is enabled.

lua/lexers/markdown.lua
4d30: Disagree, see benchmark in #1314. Reverting decreases runtime by 96%.

lua/lexers/markdown.lua
4d30: Agree, love to cut out what's not required.

lua/lexers/markdown.lua
4d30: Im not sure what this means, would you please clarify? Do you mean only the removal of 68-70 and the addion of 32?

When I diff the files, I see a significant addition added in 16e31ce lines 32-70.

lua/lexers/markdown.lua
rnpnr: You didn't run the benchmark with only this change. The difference from this hunk alone is insignificant.

4d30: You're right. Working from a functional state like 0208d7a and then benchmarking and improving from there would be a productive path.

lua/lexers/markdown.lua
rnpnr: Sorry, I didn't mean to be unclear. What I meant was that changing `code_line` rule back to its old form was the only part that was meaningfully changing performance. That is why I pointed out to upstream that it shouldn't have been running on every line and it was fixed. But as shown by your benchmark that was not the only issue.

PR #1333, opened 2026-04-10
vis-clipboard: update manpage and add termux support

candrew66:
Two commits:
- man/vis-clipboard: document wayland and wsl clipboard tools
- vis-clipboard: add support for termux

The termux commit builds on top of the manpage commit.


rnpnr:
Thanks for the patch!

Issue #1311, opened 2026-02-09
[Feature Request]: Display syntax/lexer on statusbar


Nomarian:

Nomarian:
Sure, I'm just wondering about making it the default behavior

Nomarian:
I suppose in comparison, it is more wasteful than C code but from the options I've seen, C status bars are not dynamic and if they allow for it then you get dwm which is just a pipe (its usually shellscript and about the most wasteful thing you can do) or you end up with tmux, which implements its own scripting language mixture (a giant mess) and you still have to execute shellscript anyway. i3status is cleaner but ehh now I need a cronjob to create a file so path_exists picks it up and formats the bar correctly. just a PITA.

Using any scripting language for a status bar is a better choice than that IMO. the code in question is just

```
table.insert(left_parts, win.file.syntax or "NOSYNTAX") -- or maybe right_parts?
```

in vis-std.lua

rnpnr:
Oh I wasn't suggesting we remove the option to set the status bar from lua, only that the existing code just be commented and left as an example. Then for the majority of users who are using the default configuration they get a more efficient version implemented in C. Note we already have code for doing this when vis is compiled without support for lua. I would like to see that tidied up and become the default case regardless of whether lua is enabled or not.

For the record I don't find any issue writing robust, scriptable, C interfaces. But I may be a bit biased having written entire graphical applications in C using only a 3D graphics API to draw things.

PR #1334, opened 2026-04-10
[RFC] Add a plugin namespace

### LINKED ISSUE #1332: [Feature Request]: Extend `Ctrl+N `completions
### What feature would you like to see?

Currently `Ctrl+N` can only complete words within the current file. I request that we make it so that plugins can add additional words to the list without completely overriding the binding. For example, in the completions PR I wrote for [vis-ctags](https://github.com/kupospelov/vis-ctags/pull/24), I frankensteined `complete-word.lua` so that `Ctrl+N` would also complete tags along with words.

fischerling:
Multiple times we needed a way for plugins to communicate with each other.
Currently the user has to do the plumbing in their `visrc.lua` file.

This proposal simply adds a global table `vis.plugins` as central namespace
for all plugins to expose their APIs.

Each plugin would insert its own table `vis.plugins['plugin-name']`.

This allows plugins to be easily extended by other plugins without
duplicating a lot of user configuration code.

### Usecases:

* The status line could be extendable
* The word-completion could be extended
* The lspc configuration could be changed by an editorconf plugin
* ...

### Questions

* Do we want this in general?
    * In my opinion it is really simple but powerful, therefore a good addition.
* Should the plugin namespace be userdata exposed by C to prevent it from be deleted `vis.plugins=nil`?
* Should the table be read-only and use registration functions to add new plugin subtables? 

### Related

* https://github.com/martanne/vis/issues/1332
* https://github.com/martanne/vis/issues/1311
* https://gitlab.com/muhq/vis-lspc/-/merge_requests/54

Fixes #1332.

### DIFF
--- a/lua/plugins/complete-word.lua
+++ b/lua/plugins/complete-word.lua
@@ -1,5 +1,14 @@
 -- complete word at primary selection location using vis-complete(1)
 
+-- Table of hooks other plugins may register to insert their completion
+-- candidates.
+-- Hooks are called with the current word prefix and MUST return a table
+-- containing their completion candidates.
+local candidate_hooks = {}
+if vis.plugins then
+	vis.plugins["complete-word"] = {candidate_hooks = candidate_hooks}
+end
+
 vis:map(vis.modes.INSERT, "<C-n>", function()
 	local win = vis.win
 	local file = win.file
@@ -17,6 +26,13 @@ vis:map(vis.modes.INSERT, "<C-n>", function()
 	-- collect words starting with prefix
 	vis:command("x/\\b" .. prefix .. "\\w+/")
 	local candidates = {}
+
+	for _, candidate_hook in ipairs(candidate_hooks) do
+		for _, candidate in ipairs(candidate_hook(prefix)) do
+		  table.insert(candidates, candidate)
+		end
+	end
+
 	for sel in win:selections_iterator() do
 		table.insert(candidates, file:content(sel.range))
 	end
--- a/lua/vis-std.lua
+++ b/lua/vis-std.lua
@@ -89,6 +89,11 @@ local modes = {
 	[vis.modes.REPLACE] = 'REPLACE',
 }
 
+-- Namespace for lua plugins to expose their API and communicate with each other.
+-- Each plugin can use its own namespace vis.plugins['plugin-name'].
+local plugins = {}
+vis.plugins = plugins
+
 vis.events.subscribe(vis.events.WIN_STATUS, function(win)
 	local left_parts = {}
 	local right_parts = {}


TwoF1nger:
> Multiple times we needed a way for plugins to communicate with each other.
Currently the user has to do the plumbing in their visrc.lua file.

Is there an example of this plumbing that I could take a look at?

I'm asking because a few of my plugins do customize each other via hooks, but no action from the user is necessary. The user doesn't even need to know about it.

The thing is, my hooks are just functions in the global namespace. On one hand, this runs the risk that another plugin, or the user, defines a symbol with that same name. On the other hand, this keeps everything working even if the user renames the plugins to their liking. Which is not possible if you use per-plugin tables.

Another thing - if per-plugin tables are what you want, can't you use `package.loaded['name']` instead, since it already exists. The same API exposed by returning a module table can be accessed directly by other plugins, without any plumbing by the user (IIUC what you mean by plumbing).

fischerling:
> > Multiple times we needed a way for plugins to communicate with each other.
> > Currently the user has to do the plumbing in their visrc.lua file.
> 
> Is there an example of this plumbing that I could take a look at?

My `visrc.lua` has no example because the plugins I use are not integrated.
For example `lspc` exposes its `select` implementation via the returned module table.
If a second plugin should use this function, without using the global namespace the user has to register the `select` implementation for use by the second plugin.

```lua
local lspc = require('vis-lspc')
local plugin2 = require("plugin2")
plugin2.use_select_impl(lspc.select)
```

> I'm asking because a few of my plugins do customize each other via hooks, but no action from the user is necessary. The user doesn't even need to know about it.
> 
> The thing is, my hooks are just functions in the global namespace. On one hand, this runs the risk that another plugin, or the user, defines a symbol with that same name. On the other hand, this keeps everything working even if the user renames the plugins to their liking. Which is not possible if you use per-plugin tables.

In general I dislike using globals, but for our use case the "C approach" (globals with a safe prefix (`vis_plugin_foo`)) should be totally fine.

Using a per plugin table under a namespace **controlled** by the plugin is what I propose.
Plugin `foo` would register itself with the chosen name `foo`.
Therefore, it does not matter how the user is requiring the plugin.
This eliminates the use of globals and has the same effect as your approach.

> Another thing - if per-plugin tables are what you want, can't you use `package.loaded['name']` instead, since it already exists. The same API exposed by returning a module table can be accessed directly by other plugins, without any plumbing by the user (IIUC what you mean by plumbing).

I did not know about `package.loaded` indeed this is similar to the more narrow suggested `vis.plugins`.
However, it has the disadvantage that the names are not stable.
If I require `vis-lspc` as `lspc` how should another plugin know the local name I used to require `vis-lspc`?

rnpnr:
I do like the idea of having a coherent way of allowing plugins to interact with each other. Some additional thoughts:

* I think the plugins table should be managed on the C side. I am still considering experimenting with plugins written in C and having the table exposed from C would allow for C plugins to be interacted with from lua.
  * I need to think a little bit more about his because it should be possible to allow interaction in both directions.
* This can be addressed separately but for debug purposes I would like a list of loaded plugins from the `package.loaded` table, and from this new `vis.plugins` table.
* I think that registration should require that a version number be provided. Other plugins can rely on always being able to access a version number for the plugin they are trying to interact with.
  * This would imply that we can't just allow plugins to directly modify the `vis.plugins` table.
  * I'm thinking that plugin will call a registration function which will add an entry to the table with a version field and after that the plugin is allowed to add and modify other fields as it sees fit.

PR #1337, opened 2026-04-20
lexers: update for later RouterOS releases

eworm-de:


fischerling:
Did we reach a conses how to handle lexer updates?

I know that Matej did favor pulling all changes from Scintillua at once.

@rnpnr Should we merge individual lexer changes?

eworm-de:
What ever you like... I can send it there if you prefer.

fischerling:
> What ever you like... I can send it there if you prefer.

If you did not sent the patches upstream, then yes please create a pull request at https://github.com/orbitalquark/scintillua.

eworm-de:
Opened https://github.com/orbitalquark/scintillua/pull/200 ...

Do you want to close this now?

fischerling:
> Do you want to close this now?

I think we should wait for consensus on how we handle our downstream lexer code.

rnpnr:
> @rnpnr Should we merge individual lexer changes?

I think we want to avoid it and instead just bulk update whenever there is a new scintillua release. Its easier to manage that way (also if we squash all the scintillua changes into a single commit).

eworm-de:
BTW... There was a release three weeks ago.
https://github.com/orbitalquark/scintillua/releases/tag/scintillua_6.7

So how do we handle this now?

rnpnr:
We can pull in the changes in a update to scintillua 6.7 commit. We need to be a little careful though because a worse partial version of #1264 was added to upstream. @Nomarian would you be able to help? Or at least comment on if its useful in vis or if we should just ignore that and keep the work you did unchanged?

fischerling:
> We can pull in the changes in a update to scintillua 6.7 commit. We need to be a little careful though because a worse partial version of #1264 was added to upstream. @Nomarian would you be able to help? Or at least comment on if its useful in vis or if we should just ignore that and keep the work you did unchanged?

Sorry @rnpnr I read your comment to late for #1354.

PR #1345, opened 2026-05-02
ui: implement numberwidth option

### LINKED ISSUE #1261: [Feature Request]: implement 'numberwidth' option from Vim
### What feature would you like to see?

Vim has an option called [`numberwidth`](https://vimhelp.org/options.txt.html#%27numberwidth%27) which allows you to set how many columns to use for the line number. numberwidth is a value that can range from 1-20.

nvim with numberwidth equal to 8

rnpnr:
This is an option from vim which is quite nice for preventing the source from shifting when you jump across power of 10 line count boundaries. Unlike the implementation in vim we don't clamp the user provided value to 20 because that seems very arbitrary.

closes: #1261

### DIFF
--- a/man/vis.1
+++ b/man/vis.1
@@ -1,4 +1,4 @@
-.Dd January 14, 2017
+.Dd May 14, 2026
 .Dt VIS 1
 .Os Vis VERSION
 .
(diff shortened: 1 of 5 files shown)


fischerling:
This should probably added to `man/vis.1`.

Otherwise lgtm :)

PR #1349, opened 2026-05-08
core: unify lua and :set option handling

rnpnr:
The only thing that needs to be different about these paths is how the value that will be assigned to option is obtained. When the value comes from lua we can ask lua directly for the expected type but when it comes from :set we need to parse it ourselves.

Besides that, lookup of the option, and setting of the option should not be using different codepaths. By combining them we no longer need worry about modifying multiple places in the code to add an option or update its handling.

The addition of a vis_option_get() helper should also facilitate easy building of a full table of option values to pass to lua but that is not implemented here. Furthermore the C-API was updated to add a VisOptionGetFunction pointer to the option registration which would allow custom options to be included in the table but since addition to lua would require breaking the existing API it needs to be further discussed.

### DIFF
--- a/vis.h
+++ b/vis.h
@@ -52,6 +52,23 @@ typedef union {
 	void (*f)(Vis*);
 } Arg;
 
+typedef enum {
+	VisValueKind_Nil,
+	VisValueKind_String,
+	VisValueKind_Integer,
+	VisValueKind_Boolean,
+	VisValueKind_Count,
+} VisValueKind;
+
+typedef struct {
+	VisValueKind kind;
+	union {
+		int64_t     integer;
+		bool        boolean;
+		const char *string;
+	} u;
+} VisValue;
+
 /**
  * Key action handling function.
  * @param vis The editor instance.
@@ -1152,42 +1169,49 @@ VIS_EXPORT bool vis_cmd_unregister(Vis *vis, const char *name);
  * @{
  */
 /** Option properties. */
-enum VisOption {
-	VIS_OPTION_TYPE_BOOL = 1 << 0,
+typedef enum {
+	VIS_OPTION_TYPE_BOOL   = 1 << 0,
 	VIS_OPTION_TYPE_STRING = 1 << 1,
 	VIS_OPTION_TYPE_NUMBER = 1 << 2,
-	VIS_OPTION_VALUE_OPTIONAL = 1 << 3,
-	VIS_OPTION_NEED_WINDOW = 1 << 4,
-	VIS_OPTION_DEPRECATED = 1 << 5,
-};
+	VIS_OPTION_NEED_WINDOW = 1 << 3,
+	VIS_OPTION_DEPRECATED  = 1 << 4,
+} VisOptionFlags;
 
 /**
  * Option handler function.
  * @param vis The editor instance.
  * @param win The window to which option should apply, might be ``NULL``.
  * @param context User provided context pointer as given to `vis_option_register`.
- * @param force Whether the option was specified with a bang ``!``.
+ * @param toggle Whether the option was specified with a bang ``!``.
  * @param option_flags The applicable option flags.
  * @param name Name of option which was set.
  * @param value The new option value.
  */
-typedef bool (VisOptionFunction)(Vis *vis, Win *win, void *context, bool force,
-                                 enum VisOption option_flags, const char *name, Arg *value);
+#define VIS_OPTION_SET_FUNCTION(fname) bool fname(Vis *vis, Win *win, const char *name, \
+                                                  void *context, VisOptionFlags flags, \
+                                                  VisValue value, bool toggle)
+typedef VIS_OPTION_SET_FUNCTION(VisOptionSetFunction);
+#define VIS_OPTION_GET_FUNCTION(fname) VisValue fname(Vis *vis, Win *win, const char *name, \
+                                                      void *context, VisOptionFlags flags)
+typedef VIS_OPTION_GET_FUNCTION(VisOptionGetFunction);
 
 /**
  * Register a new ``:set`` option.
  * @param vis The editor instance.
  * @param names A ``NULL`` terminated array of option names.
  * @param option_flags The applicable option flags.
- * @param func The function handling the option.
- * @param context User supplied context pointer passed to the handler function.
+ * @param set The function which handles a set operation for the option.
+ * @param get The function which handles a get operation for the option.
+ * @param set_context User supplied context pointer passed to the set function.
+ * @param get_context User supplied context pointer passed to the get function.
  * @param help Optional single line help text.
  * @rst
  * .. note:: Fails if any of the given option names is already registered.
  * @endrst
  */
-VIS_EXPORT bool vis_option_register(Vis *vis, const char *names[], enum VisOption option_flags,
-                                    VisOptionFunction *func, void *context, const char *help);
+VIS_EXPORT bool vis_option_register(Vis *vis, const char *names[], VisOptionFlags flags,
+                                    VisOptionSetFunction *set, VisOptionGetFunction *get,
+                                    void *set_context, void *get_context, const char *help);
 /**
  * Unregister an existing ``:set`` option.
  * @param vis The editor instance.
--- a/sam.c
+++ b/sam.c
@@ -258,147 +258,6 @@ static const CommandDef cmddef_select = {
 	NULL, VIS_HELP(NULL) CMD_NONE, NULL, cmd_select
 };
 
-/* :set command options */
-typedef struct {
-	const char *names[3];            /* name and optional alias */
-	enum VisOption flags;            /* option type, etc. */
-	VIS_HELP_DECL(const char *help;) /* short, one line help text */
-	VisOptionFunction *func;         /* option handler, NULL for builtins */
-	void *context;                   /* context passed to option handler function */
-} OptionDef;
-
-enum {
-	OPTION_SHELL,
-	OPTION_ESCDELAY,
-	OPTION_AUTOINDENT,
-	OPTION_EXPANDTAB,
-	OPTION_TABWIDTH,
-	OPTION_SHOW_SPACES,
-	OPTION_SHOW_TABS,
-	OPTION_SHOW_NEWLINES,
-	OPTION_SHOW_EOF,
-	OPTION_STATUSBAR,
-	OPTION_NUMBER,
-	OPTION_NUMBER_RELATIVE,
-	OPTION_CURSOR_LINE,
-	OPTION_COLOR_COLUMN,
-	OPTION_SAVE_METHOD,
-	OPTION_LOAD_METHOD,
-	OPTION_CHANGE_256COLORS,
-	OPTION_LAYOUT,
-	OPTION_IGNORECASE,
-	OPTION_BREAKAT,
-	OPTION_WRAP_COLUMN,
-};
-
-static const OptionDef options[] = {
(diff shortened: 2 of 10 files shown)


rnpnr:
With respect to having a `get` function pointer for custom options I would like feedback from plugin authors who register custom options.

@fischerling I know you register a `spelllang` option in [vis-spellcheck](https://gitlab.com/muhq/vis-spellcheck/-/blob/moved/spellcheck.lua?ref_type=heads#L479) (and I'm sure [vis-lspc](https://gitlab.com/muhq/vis-lspc) has some as well). If we were to modify the way `win.options` works so that it actually pushes a table with all the current options and values already filled a get callback would allow for the value of `spelllang` to be included in the table. What are your thoughts on this?

I do believe the additional function pointer is valuable on the C side because it would be required for a possible C plugin interaction (the rest of vis is already kind of set up for this even though its not implemented).

fischerling:
> With respect to having a `get` function pointer for custom options I would like feedback from plugin authors who register custom options.

Currently the `get` function pointer is included in `VisOption` and only used if the option is unknown in `vis-options.c:429`?

> 
> @fischerling I know you register a `spelllang` option in [vis-spellcheck](https://gitlab.com/muhq/vis-spellcheck/-/blob/moved/spellcheck.lua?ref_type=heads#L479) (and I'm sure [vis-lspc](https://gitlab.com/muhq/vis-lspc) has some as well). If we were to modify the way `win.options` works so that it actually pushes a table with all the current options and values already filled a get callback would allow for the value of `spelllang` to be included in the table. What are your thoughts on this?

What are the benefits of having an actual lua table of options?
The only benefit I see is to include the actual option values in the `:help` output or in a separate `:options` window.

> 
> I do believe the additional function pointer is valuable on the C side because it would be required for a possible C plugin interaction (the rest of vis is already kind of set up for this even though its not implemented).

I see no reason why we should not allow to pass a get function pointer during `vis:options_register`.

However, I think  we should not make it mandatory and force all plugins to adapt.
This are the calls to `option_register` in my setup.

```
vis-spellcheck/spellcheck.lua:1
vis-editorconfig/edconf.lua:1
vis-ctags/init.lua:1
vis-lspc/init.lua:8
```

rnpnr:
> Currently the `get` function pointer is included in `VisOption` and only used if the option is unknown in `vis-options.c:429`?

Yes, more specifically it is called for user registered options. This is the same as the original `func` pointer (renamed to `set`). 

Note as a consequence of this commit custom options will also be accessible through the `vis.options` table. In it's current state it should be possible to do `vis.options.spelllang = 'foo'` if you have `vis-spellcheck` (untested). Part of the reason I'm requesting feedback is to determine if this is a good idea or not.

> What are the benefits of having an actual lua table of options? The only benefit I see is to include the actual option values in the `:help` output or in a separate `:options` window.

Exactly. Furthermore, we allow you to set the options via a table (eg. `win.options = { relativenumbers = true }`) but there is no way to know, without looking through the documentation, exactly which members are in the `win.options`/`vis.options` tables. If we push an actual table then the user can just iterate over the keys/values in lua and do whatever they want.

> I see no reason why we should not allow to pass a get function pointer during `vis:options_register`.
> 
> However, I think we should not make it mandatory and force all plugins to adapt. This are the calls to `option_register` in my setup.

I agree that it should not be mandatory but the reason I suggested breaking the API is so that we can avoid another `vis:pipe` situation. The ergonomics are not great with the optional help string currently coming at the end. Granted `vis:option_register(name, type, set [, get] [, help])` wouldn't be that bad. Since `get` has to be a function there shouldn't be any ambiguity.

fischerling:
> > Currently the `get` function pointer is included in `VisOption` and only used if the option is unknown in `vis-options.c:429`?
> 
> Yes, more specifically it is called for user registered options. This is the same as the original `func` pointer (renamed to `set`).
> 
> Note as a consequence of this commit custom options will also be accessible through the `vis.options` table. In it's current state it should be possible to do `vis.options.spelllang = 'foo'` if you have `vis-spellcheck` (untested). 

Yes this works indeed.

> Part of the reason I'm requesting feedback is to determine if this is a good idea or not.

This was possible before by using `vis:command(':set spelllang foo')`.
So there is no difference in functionality.
Maybe it is easier to do mistakes when assigning directly to an options.

However, for the plugins I maintain and their options I see no reason to forbid lua code to directly assign values via the options table.

> > I see no reason why we should not allow to pass a get function pointer during `vis:options_register`.
> > However, I think we should not make it mandatory and force all plugins to adapt. This are the calls to `option_register` in my setup.
> 
> I agree that it should not be mandatory but the reason I suggested breaking the API is so that we can avoid another `vis:pipe` situation. The ergonomics are not great with the optional help string currently coming at the end. Granted `vis:option_register(name, type, set [, get] [, help])` wouldn't be that bad. Since `get` has to be a function there shouldn't be any ambiguity.

Yeah preventing a second `vis:pipe` is a good idea!
So plugins have to test with `pcall` if the old or new API is present?

Should we post a notification about this on the mailing list?

rnpnr:
> > I agree that it should not be mandatory but the reason I suggested breaking the API is so that we can avoid another `vis:pipe` situation. The ergonomics are not great with the optional help string currently coming at the end. Granted `vis:option_register(name, type, set [, get] [, help])` wouldn't be that bad. Since `get` has to be a function there shouldn't be any ambiguity.
> 
> Yeah preventing a second `vis:pipe` is a good idea! So plugins have to test with `pcall` if the old or new API is present?
> 
> Should we post a notification about this on the mailing list?

I think that having to test with `pcall` is even worse for plugin authors. Honestly I didn't think very hard about it to start. I just knew I didn't want `vis:option_register(name, type, set [, help] [, get])` because randomly having the help string in the middle is very weird. I don' think that `vis:option_register(name, type, set [, get] [, help])` will have any issues. We can just check if the 4th argument is a string or a function. The main issue with `vis:pipe` is that we have a bunch of arguments that are all strings but they mean different things depending on what else is passed.

fischerling:
> I don' think that `vis:option_register(name, type, set [, get] [, help])` will have any issues. We can just check if the 4th argument is a string or a function.

Did not think about this. So we can we have a non breaking API change :)

rnpnr:
Ok, so I added the optional get method. I also added the option name as a third parameter to the set method, omitting this was obviously an oversight in the original interface. It is backwards compatible but the lua code will get a `nil` if the running version of vis doesn't support it.

I did the following to test:
```lua
local local_options = {}
local option_set_handler = function(value, toggle, name)
	if name == "test1" then
		local_options[name] = toggle and not local_options[name] or value
	else
		local_options[name] = value
	end
end
local option_get_handler = function(name)
	return local_options[name]
end
vis:option_register("test1", "bool",   option_set_handler, option_get_handler, "Test1 Option")
vis:option_register("test2", "string", option_set_handler, option_get_handler, "Test2 Option")
vis:option_register("test3", "string", option_set_handler, "Test3 Option")
```

Then I did:
```
:set test1 true
:set test2 foo
:set test3 bar
```

Then in lua code I tried to read back the values (through `vis.options.testx`) and printed with `vis:message`
```
test1 = true
test2 = foo
test3 = nil
```

fischerling:
> Ok, so I added the optional get method. I also added the option name as a third parameter to the set method, omitting this was obviously an oversight in the original interface. It is backwards compatible but the lua code will get a `nil` if the running version of vis doesn't support it.

LGTM. Why did you force push? Instead of adding a second commit?
This makes it harder for me to look through your additional changes.

> I did the following to test:
> 
> ```lua
> local local_options = {}
> local option_set_handler = function(value, toggle, name)
> 	if name == "test1" then
> 		local_options[name] = toggle and not local_options[name] or value
> 	else
> 		local_options[name] = value
> 	end
> end
> local option_get_handler = function(name)
> 	return local_options[name]
> end
> vis:option_register("test1", "bool",   option_set_handler, option_get_handler, "Test1 Option")
> vis:option_register("test2", "string", option_set_handler, option_get_handler, "Test2 Option")
> vis:option_register("test3", "string", option_set_handler, "Test3 Option")
> ```
> 
> Then I did:
> 
> ```
> :set test1 true
> :set test2 foo
> :set test3 bar
> ```
> 
> Then in lua code I tried to read back the values (through `vis.options.testx`) and printed with `vis:message`
> 
> ```
> test1 = true
> test2 = foo
> test3 = nil
> ```

This should probably be included in our tests.

fischerling:
Unfortunately, plugins still have to use `pcall` to detect the new API if they want to use it.
Running `vis:option_register('name', 'type', set_func, get_func, 'help')` will fail on older version of vis without this patch.

Therefore, to safely use the new API a `pcall` check is required.

```lua
if not pcall(vis.option_register, vis, 'name', 'type', set_func, get_func, 'help'))
then
  vis:option_register('name', 'type', set_func, get_func, 'help')
end

File: vis.h [hunk 1]
```diff
@@ -1171,23 +1187,31 @@ enum VisOption {
  * @param name Name of option which was set.
  * @param value The new option value.
  */
-typedef bool (VisOptionFunction)(Vis *vis, Win *win, void *context, bool force,
-                                 enum VisOption option_flags, const char *name, Arg *value);
+#define VIS_OPTION_SET_FUNCTION(fname) bool fname(Vis *vis, Win *win, const char *name, \
+                                                  void *context, VisOptionFlags flags, \
+                                                  VisValue value, bool force)
```
fischerling: The parameter should be named toggle instead of force.

vis.h
fischerling: This should be changed to the consistent use of `toggle`.

fischerling (changes requested):
In some places the presence of '!' is called force but in the documentation and most other places it is called toggle.

`:set bool-option!` should be called toggle since there is nothing to force.

rnpnr:
> Unfortunately, plugins still have to use `pcall` to detect the new API if they want to use it. Running `vis:option_register('name', 'type', set_func, get_func, 'help')` will fail on older version of vis without this patch.

You are right, I realized this as well later on. I wonder if we can make this easier going forward. Right now there is `vis.VERSION` in lua but it is kind of useless because its a string. I wonder if we should add a some sort of numeric version. For example the integer commit counter (e.g. `git rev-list --count HEAD`). Then plugin authors can just check against a known point in time rather then having to guess if a workaround is needed or not. Obviously this would only be useful past whatever commit adds such a feature.

> LGTM. Why did you force push? Instead of adding a second commit? This makes it harder for me to look through your additional changes.

Yes I agree. Its a bad habit that I should try to stop.

> This should probably be included in our tests.

Sure, I can probably make a version that works there.

vis.h
rnpnr: Yes I can probably change that. I think everywhere it is used just treats it as a toggle. The 'force' meaning doesn't seem to be useful.

fischerling (approved):

rnpnr:
One thing I'm not entirely satisfied with is that there is no way through the lua interface to specify that an option applies to a window (`VIS_OPTION_NEEDS_WINDOW`) even though the C interface for registering an option can. I think its out of scope for this PR but this deficiency did become much more apparent with these other updates.

fischerling:
> One thing I'm not entirely satisfied with is that there is no way through the lua interface to specify that an option applies to a window (`VIS_OPTION_NEEDS_WINDOW`) even though the C interface for registering an option can. I think its out of scope for this PR but this deficiency did become much more apparent with these other updates.

Yes I agree!
I stumbled upon this while updating `vis-spellcheck`, because the spelllang option is window specific.
Currently is exists in the gloabl scope but does return different values depending on the active window.

PR #1352, opened 2026-05-12
Feature/docs and nix

micke754:


micke754:
Tweaking readme and updating the flake for reproducible builds.

micke754:
Apologies, accidently pushed this upstream rather than to my fork.

PR #1241, opened 2025-05-17
build: allow vendored termkey

sewnie:
As discussed in #79, martanne expressed his refusal for a vendored libtermkey, however, there are distributions that do not want to package an EOL or deprecated library, hence, this patch, similar to Neovim, vendors libtermkey into the source tree, while still allowing distributors to use the system libtermkey by default, if detected.

Changed of note to libtermkey:
- Removed unibilium support

PR #1353, opened 2026-05-17
vendor termkey

### LINKED ISSUE #1071: libtermkey status & curses cannot be disabled
> Note: Use of this library for new programs is discouraged. It is being maintained for legacy purposes only. For writing new programs, please use [libtickit](https://www.leonerd.org.uk/code/libtickit/) instead.

> All of the code in this library will be migrated into libtickit in due course; currently it is used via an internal dependency.

`libtermkey` is a required dependency, and it depends on curses, which will cause a build failure, since curses is needed by a require dependency.

rnpnr:
This is a continuation of #1241 which fixes the issue of always needing curses for embedded termkey by using unibilium (also included) to provide standalone terminfo parsing. On its own unibilium is quite small but I removed roughly 50% since it was not needed by termkey. I then made a number of changes to both libraries to generally improve them for their use here (reduce memory allocations, reduce pointer chasing, etc.).

Since this is a continuation I kept the ability for vis to be built against the system's termkey but I kind of want to remove that and make this the only option. I have a few changes in mind to further reduce memory allocations and make them more suitable for vis but I don't think I can do so while maintaining the backwards compatibility with their existing APIs. Since neither of these libraries are receiving updates I feel like its better to go all in on the included versions and change them in whatever way is most optimal for vis. What are other peoples thoughts on this?

Note I have tested this on a couple different terminal emulators (st, alacritty) and the CI is handling it fine but testing on more terminal emulators wouldn't hurt. If you run a different terminal emulator please test and let me know if it is working. I'm particular interested in wayland only emulators (like foot) and whatever people are using on macOS nowadays since these are the ones that tend to cause issues.


rnpnr:
So having thought about it quite a lot I think we should just remove support for system termkey. It no longer receives updates and I could not find any relevant patches to the source code after looking at a number of different libtermkey packages in various distos/BSDs. In Gentoo Linux the only package with a dependency on libtermkey is vis, it seems to be the same in Arch.

Along with the fact that this gets optimized inline with the rest of vis, the other changes I have made will already offer a better experience than the normal termkey. If I don't need to bother with API compatibility I can make further improvements.

Once I am done removing system termkey support and making one immediate change to avoid mallocing a termkey struct I will test for a few more days and then I will merge this as I have not received any feedback. I am fully confident that this is the right move going forward.

PR #1354, opened 2026-05-19
lua/lexers: update to scintillua 6.7

fischerling:
Bugfixes:
	- Fixed PKGBUILD lexer to not tag keywords after '.' and numbers that are parts of words. Changes:
	- Updated RouterOS lexer keywords.
	- Updated PKGBUILD lexer keywords, functions, and constants.
	- Added Swift lexer.
	- Added Arduino lexer.
	- Updated Go lexer with new functions.
	- Various Perl lexer improvements.
	- Recognize more extensions for C++, diff, Javascript, Makefile, Powershell, and Python lexers.
	- Added lexer.ignore_extensions and lexer.ignore_patterns for ignoring extensions and filename parts, respectively, in lexer.detect().
	- Added more LaTeX math environment keywords.
	- Updated Lua lexer to reflect changes in Lua 5.5.
	- Dart lexer recognizes capitalized words as types.

### DIFF
--- a/lua/lexers/actionscript.lua
+++ b/lua/lexers/actionscript.lua
@@ -1,4 +1,4 @@
--- Copyright 2006-2025 Mitchell. See LICENSE.
+-- Copyright 2006-2026 Mitchell. See LICENSE.
 -- Actionscript LPeg lexer.
 
 local lexer = require('lexer')
--- a/lua/lexers/ada.lua
+++ b/lua/lexers/ada.lua
@@ -1,4 +1,4 @@
--- Copyright 2006-2025 Mitchell. See LICENSE.
+-- Copyright 2006-2026 Mitchell. See LICENSE.
 -- Ada LPeg lexer.
 
 local lexer = require('lexer')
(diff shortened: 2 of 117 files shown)


fischerling:
@rnpnr @Nomarian What do you think about the filetype changes?

rnpnr:
I guess I was mistaken, this doesn't really effect much about what we are doing in `filetype.lua`. I think there is a little work to do to remove our `M.ignoresuffixes` since I think it is just a duplicate of what is now in `lexer.lua`, but I will have to look at it in a little more detail later.

Nomarian:
Overall I don't mess with lexing stuff, only on filetypes.lua

There are some behavioral changes, Vis implementation identifies multiple junk extensions like .bak.bak or .old.old but lexer.lua only identifies a single extension. which... maybe the time of dealing with multiple .bak.bak files is over, (someone might use it maybe)

Identifying utility directly via lookup table was rejected. https://github.com/orbitalquark/scintillua/pull/188

Multiple standard unix filenames like .login and .profile where also rejected https://github.com/orbitalquark/scintillua/pull/187

So filetypes should remain mostly as is, except if the ignore_extensions behavior is not acceptable.

fischerling:
I have no strong opinion about this.

Maybe we could keep both our filetype detection because it does more and the new scintillua bits just to not modify the upstream code.
This means, similar things are done twice, which I think is ok because I consider the file type detection not performance critical.

rnpnr (approved):
Yes I think its fine to keep what we have in `filetype.lua` for now. Feel free to apply!

Issue #1356, opened 2026-05-23
[RFC] Restructing Theme Handling


rnpnr:
Hi,

(TLDR follows)

Currently theme handling in vis is a giant mess. In theory the code is written such that every window can have its own theme but in practice that is not really possible. The problem is that when you do `:set theme mytheme` the expectation is that `mytheme.lua` will modify the internal `vis.lexers` table and overwrite all the styles in there. Then all open windows are looped over and their styles are updated to the new definitions in `vis.lexers`. What this means practically is that we just copy the exact same style definitions into all open windows which is obviously a complete waste of time and memory.

This makes no sense. Personally I have never wanted to have two different themes used in the same editor instance. Even if you wanted to you could just open a second vis and set the theme differently in that instance. This is the first issue to address.

Next, a user may want to define their own styles to use with `win:style()`. This is done with `win:style_define(id, ...)`. How does a user obtain a new id to use? Currently you have to look into the internals of the lexers, understand that they expect to assign a bunch ids themselves in a very specific way, then try to pick one after these predefined ones and hope that they haven't all been used up (hint: currently the HTML lexer uses them all up). Furthermore, if the lexer needs to change to one that has more tags defined it will overwrite your existing styles and you will have to redefine them using new ids.

Again this is completely broken. What I think should happen is that we should a function `ui:style_push()` which gives you a new id that you are free to use. The lexers should also use this function the first time they see a new tag and not just assume they can arbitrarily use an index of their choosing. Practically speaking the vt100 backend can have unlimited styles defined, and the curses backend is limited because we use some indexed table. We can still choose some reasonable limit on the number of styles but we should probably stop using an indexed table unnecessarily in curses (all reasonable terminal emulators support arbitrary 8bit per channel color on every cell).

## Summary

What I would like to do is:

- [ ] Remove per window style definition. Its not really even possible to use it currently.
 * A new `ui:style_define()` function should be added making it clear that styles are defined globally.
 * `win:style_define()`  could be kept around for backwards compatibility even though it will just reroute to `ui:style_define()`.
- [ ] Add a new `ui:style_push()` function for getting a new id which the caller can freely use as they wish.
 * This should probably optionally take the style definition so you can get an id and define the style in one call.
 * This will also fix the current issue of the HTML lexer trying to use more ids than are available.
[...]

PR #1358, opened 2026-05-29
Rework Style Handling

rnpnr:
Details in #1356 and in the commit messages.

Testing would be appreciated especially if you were experiencing the some of the bugs mentioned in the 3rd commit.

Also any comments on the lua API should be mentioned now, it is preferable to not have to change it later. The `keep_non_default` argument is no long supported anywhere. While it has not made it into a release it is still backwards compatible to remove it because the C code won't look for an argument in it's position. The following functions were added:

```lua
-- ui:style_push([style])
-- Obtain a new style id from style pool
-- style: (optional) the initial style definition to fill the slot
-- returns the new style id, nil if none are available
local red_fg_id   = ui:style_push("fore:red")
local green_bg_id = ui:style_push()
ui:style_define(green_bg_id, "back:green")
```

```lua
-- ui:style_define(style_id, style)
-- Set parameters for style (same as old win:style_define())
-- style_id: the style slot to fill
-- style: the style definition to fill the slot
-- returns (bool) whether the operation was successful or not
ui:style_define(style_id, "underlined,keep_attribute")
```

The `keep_attribute` keyword was added to indicate that the underlying cell's attributes should be maintained in addition to the specified attributes when the style is used. This is useful for adding the `underlined` attribute to the `CURSORLINE` for example.


fischerling:
Somehow with this PR the name of a non-existing file is not displayed in the statusbar.

`vis non-existing` results in`[No Name]` being displayed in the status bar.

While briefly looking at the first commit e2caf0ae26b5eee5ef5ab01e41c88fea0aadda92 introducing this bug, if have no idea why this happens.

rnpnr:
Are you testing with the lastest version? I also noticed that at some point and fixed it.

~~EDIT: oh you mean something else. I will have to look into it.~~
EDIT EDIT: no that seems to work for me

fischerling:
> Are you testing with the lastest version? I also noticed that at some point and fixed it.
> 
> ~EDIT: oh you mean something else. I will have to look into it.~ EDIT EDIT: no that seems to work for me

Yes I can reproduce this with e92c74454707abb308825bffffe198d99513a614 installed system wide.

rnpnr:
Are you sure this is from this patchset? Nothing related to filename or display is touched here. Can you confirm if this happens on master? In particular e9791c2c could have caused it.

fischerling:
> Are you sure this is from this patchset? Nothing related to filename or display is touched here. Can you confirm if this happens on master? In particular [e9791c2](https://github.com/martanne/vis/commit/e9791c2c70c54daa708ef3a3081402ff55d54faf) could have caused it.

I tried this, because I could not see a reason why this should happen for this patchset.

On the current master including e9791c2 the status bar correctly shows the name of a non-existing file.

`git bisect` and checking with `./configure && make clean && make && ./vis non-existing` reports e2caf0ae26b5eee5ef5ab01e41c88fea0aadda92 as first bad commit.

With this commit and a system wide install to ensure the lua file changes are in use, the status bar does not show the name of a non-existing file.

rnpnr:
> `git bisect` and checking with `./configure && make clean && make && ./vis non-existing` reports [e2caf0a](https://github.com/martanne/vis/commit/e2caf0ae26b5eee5ef5ab01e41c88fea0aadda92) as first bad commit.
> 
> With this commit and a system wide install to ensure the lua file changes are in use, the status bar does not show the name of a non-existing file.

Thanks, I will try and get a local reproducer later.

fischerling:
Now I am really confused.

On the current master (eacbb21731f7119e0dd6e0d0c8ab355b171f5b52) the status bar works properly if building an optimized binary.
Using a debug binary produced by `make debug` the same bug occures.

So it seems to be an optimization problem.

Using `git bisect` with a debug build it reports e9791c2c70c54daa708ef3a3081402ff55d54faf as the first bad commit.
Which makes sense because it touches the file name logic.

rnpnr:
> Now I am really confused.

I wasn't able to fully figure it out either. It only happens when compiling with GCC with optimizations disabled. Since I usually use Clang I didn't notice this sooner. It should be fixed with 84ee9d1. The logic error should have manifested in the other configurations as well though so I'm not sure whats up with that.

fischerling:
> > Now I am really confused.
> 
> I wasn't able to fully figure it out either. It only happens when compiling with GCC with optimizations disabled. Since I usually use Clang I didn't notice this sooner. It should be fixed with [84ee9d1](https://github.com/martanne/vis/commit/84ee9d1f3cf28c2a234db71fb721e6f7816fcba3). The logic error should have manifested in the other configurations as well though so I'm not sure whats up with that.

Thanks Randy! 84ee9d1 fixes the bug for optimized builds with gcc.

micampe:
@rnpnr I think this is because new is left uninitialized when stat() fails, so you have UB.

rnpnr:
> @rnpnr I think this is because new is left uninitialized when stat() fails, so you have UB.

It's not UB because the compiler no longer knows it is uninitialized after the syscall. I also don't think its that likely that `new` has bogus values. From what I have seen in Linux, which is where I am testing this, the passed in buffer is zeroed even on error. I am more confused as to why the logic error (matching internal files) didn't show up as an issue previously.

PR #1359, opened 2026-05-30
ui: implement title option

ghost:
This brings the ability to set the title into the core editor from the vis-title plugin. The order of the title string is based on vim's implementation (file name, directory, program name).

### DIFF
--- a/sam.c
+++ b/sam.c
@@ -16,6 +16,7 @@
  * REPRESENTATION OR WARRANTY OF ANY KIND CONCERNING THE MERCHANTABILITY
  * OF THIS SOFTWARE OR ITS FITNESS FOR ANY PARTICULAR PURPOSE.
  */
+#include "ui.h"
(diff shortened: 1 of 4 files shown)


ghost:
This is a work in progress, so feel free to offer thoughts on the PR's direction, but the basics should work.
I intend to:
- [ ] Add a title option that is off by default
- [ ] Fix the title string order to have the file name before the directory
- [ ] Re-set the title when switching windows

I will probably leave a titlestring option (similar to vim) to a future PR.

rnpnr (commented):
Here are some thoughts:
* I don't see a point of making such a behavior configurable. What benefit would the user get by turning it off?
* You probably want to set this in `vis_window_focus()` so that the title bar changes to whatever the current window is.
* This should go to `STDERR` like everything else in vis.
* The title should be reset on exit.

PR #1360, opened 2026-06-03
build: do not install libtermkey(-dev) for docker

eworm-de:
This is a follow-up on vendored termkey, see
commit 5cd42390aca3db022efc3441bba0c77deb07b6db.


rnpnr:
Oops missed that file, thanks!

PR #1362, opened 2026-06-03
build: remove vis from .PHONY

eworm-de:
With `vis` in `.PHONY` the executable was rebuilt on every call of `make`. The same was true for `make install`, causing trouble with the debug package when building a distribution package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,4 +172,4 @@ uninstall:
 	@echo removing support files from ${DESTDIR}${SHAREPREFIX}/vis
 	@rm -rf ${DESTDIR}${SHAREPREFIX}/vis
 
-.PHONY: all clean vis dist distclean install install-strip uninstall debug profile coverage test test-update luadoc luadoc-all luacheck man docker-kill docker docker-clean
+.PHONY: all clean dist distclean install install-strip uninstall debug profile coverage test test-update luadoc luadoc-all luacheck man docker-kill docker docker-clean


eworm-de:
Or is there a good reason to have it there?

eworm-de:
Hmm, it was added there in 5abe94420528d3e107eceb0f3df073e8a1270191...

rnpnr:
Some one opened a similar PR in the past and I rejected it because I don't understand why you want to run make and not get a new vis. If you can explain what it is exactly that you are trying to do I can see if it can be supported.

From my perspective vis depends on nearly every C and H file in tree so if you edit one you want vis to be rebuilt. I do not want to list every single file as a dependency of vis in the Makefile, that is already clear by looking at vis.c. For the record I'm usually going even simpler and just putting this in a build.sh:

```sh
#!/bin/sh

CFLAGS="-O0 -gdwarf-4"
#CFLAGS="-O3"
#CFLAGS="${CFLAGS} -fsanitize=address,undefined"
#CFLAGS="${CFLAGS} -fproc-stat-report"

#CFLAGS="${CFLAGS} -DCONFIG_CURSES=1"
#LDFLAGS="-ltinfow -lncursesw"

clang -march=native ${CFLAGS} -Wall -std=c99 -pedantic \
      -Wno-initializer-overrides \
      -DVIS_EXPORT=static \
      -DVIS_PATH=\"/usr/alt/share/vis\"\
      -DCONFIG_LUA=1 \
      -I/usr/include/lua5.3 \
      main.c -o vis -llua5.3 ${LDFLAGS}
```

I would make something like this the default but I don't want to spend time thinking about all the random things the Makefile has in it.

eworm-de:
The potential issue is that the debug package is made from the first build, then another build is finally packaged. Actually the build is reproducible here, generating exactly the same bits. Let's hope is stays this way.

Another reason could be some extra cpu cycle spent... But that does not matter much for a small project (compared to others taking hours to build).

So feel free to close this. But possibly a comment with explanation could make sense?

Finally I can just adjust the `Makefile` before building. 😉️

eworm-de:
[do not build twice...](https://gitlab.archlinux.org/archlinux/packaging/packages/vis/-/commit/93585fface62c8ebb07c31e5bebb57a15de5c629) 😉️

rnpnr:
> [do not build twice...](https://gitlab.archlinux.org/archlinux/packaging/packages/vis/-/commit/93585fface62c8ebb07c31e5bebb57a15de5c629) 😉️

If you are happy with this then we can close this. I know it is quite common for distros to modify the build process because they want some specific behaviour. In the future we might have something simpler like the script I posted above so I'm not really that concerned about what currently happens.

eworm-de:
Not absolutely, but it's ok. 😉️

On the other hand we have targets that do generate `vis`, but on a completely different way (like `make docker`, which I use regularly). The current situation makes sure a plain `make` generates a clean default build. Let's just keep it for now.

PR #1368, opened 2026-06-22
Fix completion of words without submitted characters in prefix

### LINKED ISSUE #1361: C-n completes candidates without the first symbol
### Problem

I'm not sure if it's a bug or intended behavior. I can activate buffer completion (C-n) after a dot, which will show all words in a buffer. If i try to complete any of such candidates, they will be completed without the first symbol of the candidate.

https://github.com/user-attachments/assets/98caefe8-de30-4e4d-8401-1e7a0a569350

### Steps to reproduce

1. Open a file with this content:
```
vim.keymap.set
vim.
```
2. Put the cursor after `vim.` on the second row
3. Press `C-n`
4. Select any completion candidate and press `Enter`

### vis version (vis -v)

vis v0.9-438-g3ca47592 +curses +lua +selinux

### Terminal name/version

foot version: 1.27.0

### TERM environment variable

foot

tassmjau:
When using `<C-n>` to do auto-completion without having typed the beginning of a word, `prefix` is whatever character came before the cursor position. In this case `#prefix` is `1`.

When appending the selected completion match, `#prefix` is subtracted from the match before the remaining characters of the match are written to `pos`.

When `prefix` is not part of the match, `#prefix` should not be taken into account for character subractiion, since that results in chars being lost from the result.

This fixes that.

Resolves https://github.com/martanne/vis/issues/1361

### DIFF
--- a/lua/plugins/complete-word.lua
+++ b/lua/plugins/complete-word.lua
@@ -26,7 +26,8 @@ vis:map(vis.modes.INSERT, "<C-n>", function()
 		candidates = table.concat(candidates, "\n")
 		local status, out, err = vis:pipe(candidates, "sort -u | vis-menu -b")
 		if status == 0 and out then
-			out = out:sub(#prefix + 1, #out - 1)
+			local start = out:sub(1, #prefix) == prefix and #prefix + 1 or 1
+			out = out:sub(start, -2)
 			file:insert(pos, out)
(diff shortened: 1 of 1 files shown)


rnpnr:
Applied with minor style tweak. Thanks!

tassmjau:
thanks!

PR #1371, opened 2026-06-26
lua: add vis.API to make feature checks easier

rnpnr:
Ignoring a few times in the early git history we now always maintain a linear revision list. This means we can use a revision counter like SVN to provide a method of doing an ordered version comparison. This allows plugin authors who care about backwards compatibility with older versions of vis to simply check the revision counter to know if a feature is available. This can allow authors to enable and disable features during ~~vis.INIT~~ plugin load instead of having to do runtime checks with pcall().

@fischerling as someone who has lots of backwards compatibility checks in your plugins, is something like this useful? Can you think of any improvements? Note I included an example in the documentation for what to do when `vis.API` is `nil` (i.e. vis was built prior to the commit in this PR).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,10 @@ MANUALS = $(EXECUTABLES:=.1)
 DOCUMENTATION = LICENSE README.md
 
 VERSION = $(shell git describe --always --dirty 2>/dev/null || echo "v0.9-git")
+API     = $(shell git rev-list --count HEAD 2>/dev/null || echo "0")
 
 CFLAGS_STD ?= -std=c99 -DNDEBUG
-CFLAGS_STD += -DVERSION=\"${VERSION}\"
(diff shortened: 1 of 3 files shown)


fischerling:
LGTM.
Thanks for working on this.

It does not solve my current backward compatibility checks, but it is definitly an improvement for the future.

rnpnr:
> It does not solve my current backward compatibility checks, but it is definitly an improvement for the future.

The idea is that after the next release you would have the option of using either the course grained releases or this counter if you want to support new features that are not yet in a release. If you want something finer grained between v0.9 and this commit you would still need the checks you have. Ultimately its up to the plugin authors how fine grained they want to be.

All we can do is try to be more careful about getting new lua functions right the first time so that we don't end up with more weird `vis:pipe` amalgamations.

PR #1375, opened 2026-07-08
Fix: pasting the dot-register (last inserted text) verbally pastes symbolic keys

jorbakk:
When pasting the dot-register, symbolic keys like `<Escape>, <Enter>, <Left>, <Right>`, ... are inserted as text and not executed. Thus, the text `<Escape>` is always appended, and if multiple lines were inserted or a cursor key was used while inserting, the corresponding text for these keys is inserted. For example, when inserting two lines and then pasting the last insert with `".p`, the pasted text looks like:
```
line1<Enter>line2<Enter><Escape>
```
Instead of pasting the contents of the dot-register verbally, this fix switches to insert mode and then replays the content of the dot-register like a macro. The resulting text from the example then looks like:
```
line1
line2
```

### DIFF
--- a/vis-operators.c
+++ b/vis-operators.c
@@ -37,7 +37,24 @@ static size_t op_put(Vis *vis, Text *txt, OperatorContext *c) {
 	size_t pos = c->pos;
 	bool sel = text_range_size(c->range) > 0;
 	bool sel_linewise = sel && text_range_is_linewise(txt, c->range);
-	if (sel) {
+	if (vis_register_used(vis) == VIS_REG_DOT) {
+		/* Macro-replaying the dot-register prevents that op_put() is executed multiple times if
+		 * there is more than one selection, so deletion of multiple selections needs to be done
+		 * before replaying the macro. */
+		for (Selection *s = view_selections(&vis->win->view); s; s = view_selections_next(s)) {
+			Filerange r = view_selections_get(s);
+			size_t start = r.start;
+			if (text_range_size(r) > 1) {
+				text_delete_range(txt, r);
+			} else {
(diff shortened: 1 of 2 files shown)


rnpnr (commented):
I agree that this is an issue but I don't think this is the correct way to fix it. Here is an example of something which should work but doesn't with this patch:
```
iabc def<Escape>vb".p
```
what should be in the window is:
```
abc abc def
```
without this patch we get:
```
abc abc def<Escape>
```
with this patch we get:
```
abc 
```
(also note the space after the `c`.)

I assume that at some point `vis_repeat()` also reads from the `VIS_REG_DOT` register but it must do something else with the contents. We will need to do something similar here.

jorbakk:
Here's an update of the patch that should cover all use cases of putting the dot-register:
  - from visual mode
  - with multiple selections
  - with a count argument

`vis_repeat()` works very similar in that the dot-register macro is also replayed if the previous action was inserting some text.

In contrast to `vis_repeat()`, putting the dot-register needs to work also from visual mode (like in your example), which this patch takes care of.

Please also note that generally replaying a macro does not work from visual mode and probably also should not, though this is stated in the docs (vim doesn't do that either).

Therefore the following steps to replay the dot-register macro are necessary:
- in case of multiple selections in visual mode, remove the selection text in all selections. This needs to be done explicitely, because `op_put()` is executed only once when replaying a macro in `op_put()` due to resetting the action after replaying the macro.
- take care of positioning the cursor correctly for insertion and not disposing it in some edge cases
- switch to insert mode before replaying the dot-register macro
- actually replay the dot-register macro

I left some comments in the changes for each of the steps (which don't need to be kept eventually).

PR #1380, opened 2026-07-29
Disable now useless 8-bit mode

### LINKED ISSUE #1252: When using foot's official terminfo, alt+[key] outputs wrong characters
### System details 
- Distro: Void Linux.
- Window Manager: tested on both River and Sway. Bug happens on both.
- [Foot package on Void](https://github.com/void-linux/void-packages/blob/4047da148697d610a59f970662ca39e3fc15507e/srcpkgs/foot/template#L4).
- Void [patches out ncurses' terminfo entry for foot](https://github.com/void-linux/void-packages/commit/b1b3c4dcfd908777b3ddf065e175d086ebebc8f9), so it doesn't factor in. Real terminfo comes from foot's repo.
- Tested with both `us` and `eu` keyboard layouts.
- Tested with both Ghostty and Kitty on River WM. Both work as expected; their TERMs are `xterm-ghostty` and `xterm-kitty`, respectively.

### ncurses and foot
[The foot terminfo definitions in ncurses' lack the non-standard capabilities](https://codeberg.org/dnkl/foot/src/branch/master/INSTALL.md#other) [sic]. 

Some distros ship a separate `foot-terminfo` package with a `foot-extra` terminfo file.

Void patches out de definition from ncurses and ships the `foot-terminfo` package instead, built from foot's sources.

- Foot's source terminfo file: [foot.info](https://codeberg.org/dnkl/foot/src/branch/releases/1.22/foot.info). 
- [Terminfo build instructions](https://codeberg.org/dnkl/foot/src/branch/releases/1.22/INSTALL.md#terminfo).
- [IRC logs with dnkl](https://libera.catirclogs.org/foot/2025-06-20#38315066;)  (foot's maintainer). Note: press "Live updates" for better contrast on the chat.

### PD
visidata has this same bug. alt+[key] sends the same wrong characters. Seems related to the [handling of meta by foot](https://codeberg.org/dnkl/foot#alt-meta).

### Steps
1. Open `vis` in `foot`.
2. Enter `INSERT` mode, then press `Alt+h`.

### Expected results 
 Vis inserts `<M-h>`.

### What really happens
Vis inserts `è`.

### vis version (vis -v)

vis 0.9 +curses +lua +acl

### Terminal name/version

foot version: 1.22.3 -pgo +ime +graphemes -assertions

### $TERM environment variable

foot

z3bra:
Enabling 8-bit mode was rendered useless when switching to libtermkey as the input handler. This also causes issues with terminals honoring the `smm` property (metaSendsEscape) by stripping the Alt sequence and should thus be removed.

Fixes #1252

### DIFF
--- a/ui-terminal-curses.c
+++ b/ui-terminal-curses.c
@@ -264,7 +264,6 @@ ui_backend_init(Ui *ui, char *term)
 		noecho();
 		nonl();
 		keypad(stdscr, TRUE);
-		meta(stdscr, TRUE);
 		curs_set(0);
 
 		pair_content(0, &ui->curses.default_fg, &ui->curses.default_bg);


rnpnr (commented):
Thanks for the patch!

Can some other people confirm if this doesn't cause issues?

I never use the curses backend (and would like to remove it after fixing some lingering bugs) so I am unlikely to notice any issues even if I switch just for the purpose of testing.

z3bra:
I've been using this patch everyday for a week now and haven't encountered a single issue.
I also tried different terminfo (`screen`, `xterm`, `dvtm` and their 256color counterparts), and everything works as expected.

Issue #1320, opened 2026-03-02
Case changing commands over multi-line ranges lose newlines


donaldcallen:
### Problem

Assume these lines are present
````
now is the time
for all good men
````
If I position the cursor on the first character of the first line and then invoke 'shift-v' to go into VISUAL-LINE mode and then invoke 'j' to extend the VISUAL-LINE range to both lines and then issue the command 'gU', I end up with this:
````
NOW IS THE TIMEFOR ALL GOOD MEN
````
Something similar happens if use the 'gu' command. So whether converting to upper or lower case, this sequence of commands results in the loss of the newlines.

This works correctly in vim, nvim, kakoune, etc.

### vis version (vis -v)

vis v0.9-384-g93f2c297 +curses +lua +tre +acl

donaldcallen:
````
dca@giovanni:/home/dca$ echo -e "now is the time\nfor all good men" | tr '[:lower:]' '[:upper:]'
NOW IS THE TIME
FOR ALL GOOD MEN
````

````
dca@giovanni:/tmp$ tr --version
tr (GNU coreutils) 9.10
````

My initial post was based on the behavior of vis as compiled from the sources a few days ago. So I installed the released version on my system and tested that. Same problem.

donaldcallen:
Now I am really confused. Why are you fixated on the fact that awk print always appends a newline to its output  when the change you cited  that broke things uses awk printf, not print, which does not unconditionally append newlines to its output?

If you were referring to my use of print in my tests, those tests were intended to see if (GNU) awk damaged an **interior** newline, which was the original problem I raised? And it turns out it doesn't, so, as I said previously, something more complex than my simple tests is happening between vis and awk.

fischerling:
> Now I am really confused. Why are you fixated on the fact that awk print always appends a newline to its output when the change you cited that broke things uses awk printf, not print, which does not unconditionally append newlines to its output?


Both versions are broken.
We can not use `print` because in vis the selection we use `gU` on can be arbitrary.

f**oo**  
**ba**r

In this example the selected text is `"oo\nba"` if we replace this with the output of awk which appends a newline `"OO\nBA\n" the resulting file content would be:

fOO  
BA  
r

Both versions are broken. The current version swallows interior newlines. All other versions using POSIX awk will append a new line or will not preserve a leading one if present.

I have no good solution for this problem. Maybe we should implement this in vis.

TwoF1nger:
Related: #1238

[...]

PR #1381, opened 2026-07-31
Fixes case switching keybinds gu and gU so they dont chew up newlines

### LINKED ISSUE #1238: V+u removes new line
### Problem

When selecting a line in visual line mode with `V` and then converting to lowercase with `u`, the selected line is joined with the following line. This differs from Vim's behavior, where u only changes the case and preserves line structure. I believe this is a bug or unintended behavior.

### Steps to reproduce

Steps to reproduce:

1. Open a file in Vis.
2. Move to any non-empty line followed by another non-empty line.
3. Press `V` to select the line in visual line move.
4. Press `u`.

Expected behavior: 

- Line case is changed to lowercase; line structure remains unchanged.

Actual behavior: 

- Line is lowercased and joined with the next line.

### vis version (vis -v)

vis v0.9-308-g286fd30-dirty +curses +lua +tre

### Terminal name/version

_No response_

### $TERM environment variable

_No response_

Nomarian:
Fixes #1238 #1320 

| tr does the appropriate behavior but gawk supports other languages. so gawk is used

This makes it unportable, but if portability is desired, then | tr should be used and if which gawk returns true then gawk is used.


rnpnr (commented):
I don't think you need something so complicated:
```
awk '{print tolower($0)}'
```
should do the trick. You don't need to mess with the field separator or force `gawk`. 

Personally I think we should just axe this whole passing off text editing functionality to external tools but that's a different story.

Nomarian:
It is necessary because print adds a newline. (so selections are not usable) So you must use printf or ORS="". But then you have a problem with RS="", which is that awk has this weird thing that eats away the last newline before EOF. This is an all AWKs thing


Now that I think about it, gawk can use \0 in RS, so if its that.

rnpnr:
Frankly I think it's a problem with the way vis handles the result. But we can't really fix here without breaking every other use of `vis_pipe` (including from the user perspective).

Like I said, this whole passing off text editing tasks to external programs needs to be thrown away. It's not bloat for a text editor to perform text edits.

Nomarian:
No no, its fine, its awk, awk is the problem. Both sed and tr work perfectly well to transform text, take a bunch of text, transform it, print it out, its a simple gsub() call.

```
; echo this | tr -s '[:lower:]' '[:upper:]' | wc
      1       1       5

; echo -n this | tr -s '[:lower:]' '[:upper:]' | wc
      0       1       4

; echo -n this | sed 's/this/that/' | wc
      0       1       4

; echo this | sed 's/this/that/' | wc
      1       1       5
```

But awk does not work this way, awk reads a chunk of text until it hits the the record separator (RS), and it throws away the separator itself, you don't know if there is a RS, which means this is tricky (and wrong)

```
; echo this | awk 'BEGIN {ORS=""} {print toupper($0)}' | wc
      0       1       4
; echo -n this | awk 'BEGIN {ORS=""} {print toupper($0)}' | wc
      0       1       4
```

gawk is the only one that tells if there is an RS by saving it in RT (record terminator), but this is gawk specific.

```
; echo -n this | gawk 'BEGIN {ORS=""} {print toupper($0) RT}' | wc
      0       1       4
; echo this | gawk 'BEGIN {ORS=""} {print toupper($0) RT}' | wc
      1       1       5
```

if RS is multiple characters, it can be a regex, which means this is also possible

```
; echo -n '^$hi^$' | awk 'BEGIN {RS="^$";ORS=""} {print toupper($0) RT}' | wc
      0       1       6
; echo '^$hi^$' | awk 'BEGIN {RS="^$";ORS=""} {print toupper($0) RT}' | wc
      1       1       7
```

This is by far the best, most portable solution, there is one awk that does not support RS to be a regex (goawk), but that's also not the default anywhere.

I will keep using this last one for a while and see if there's a pitfall.

PR #1382, opened 2026-08-05
Issue 1320 fix

donaldcallen:
Simple but not perfect fix (it requires GNU Awk, necessitating that BSD and MacOS users install it) for Issue 1320 and 1238.

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -160,9 +160,14 @@ static const KeyBinding bindings_operators[] = {
 	{ "'",                  ACTION(MARK)                                },
 	{ "c",                  ACTION(OPERATOR_CHANGE)                     },
 	{ "d",                  ACTION(OPERATOR_DELETE)                     },
-	{ "g~",                 ALIAS(":|tr '[:lower:][:upper:]' '[:upper:][:lower:]'<Enter>") },
-	{ "gu",                 ALIAS(":|tr '[:upper:]' '[:lower:]'<Enter>")},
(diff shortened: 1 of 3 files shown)


Nomarian:
I recently wrote this up


https://github.com/martanne/vis/wiki/Tips-and-Tricks#using-and-piping-to-awk

donaldcallen:
On Wednesday, August 5th, 2026 at 19:30, lan Nomar ***@***.***> wrote:

> Nomarian left a comment [(martanne/vis#1382)](https://github.com/martanne/vis/pull/1382#issuecomment-5198560422)
>
> I recently wrote this up
>
> https://github.com/martanne/vis/wiki/Tips-and-Tricks#using-and-piping-to-awk
>
> Also I may be wrong about this, but if linux is defined then its gawk not awk?

No. It's awk.

> I think if its unix then it should be | tr -s a-z A-Z

FreeBSD tr has been extended to support UTF-8, I believe, but not Unicode in general. Gnu Awk is a better choice.

The code I submitted has been tested on FreeBSD 15.1 and Arch Linux and works correctly on both systems..

fischerling:
@donaldcallen Thanks! Could you verify that 7d53cd6fa07cce43b2ba6c33d04c8d27bb50a794 fixes the issue on FreeBSD?

rnpnr:
Closing as this was superseded by 7d53cd6.

Issue #1385, opened 2026-08-16
force highlight


JustTaken:
### Problem

The `filetype.lua` plugin is forcing syntax highlight (through the `FILE_SAVE_POST` event) despite `syntax off` being set in my configuration. I usually use text editors without syntax highlight, i guess there aren't a lot of users that prefer this way, but it's kind of a use case.

### Steps to reproduce

1. Open a file with a known file extension (e. g. `.lua`).
2. `:set syntax off`.
3. `:w`, Save it.

The save will override the `syntax off`.

### vis version (vis -v)

vis v0.9-git +curses +lua +tre +acl +selinux

### Terminal name/version

foot version: 1.27.0 +pgo +ime +graphemes +toplevel-tag +blur -assertions

### TERM environment variable

foot

JustTaken:
Sorry, by vis0.9-git i mean the latest commit as of today: 7d53cd6f.

In respect to the solution, i think it might be reverting b42b82d2f and introducing a `WIN_OPEN` event on the first save of the newly created window (with `:new`), so that we can check the issued filepath to `:w <filepath>` and set the syntax accordingly to the file extension before the `WIN_OPEN` event.

I guess i could work around it with a counter `FILE_SAVE_POST` that sets the syntax off in my configuration, instead of `win:set_syntax(nil)` inside a `WIN_OPEN` event subscription. Or perhaps there is another solution that is common knowledge that I'm not aware of.

Either way I'm willing to make a pull request regarding this issue, or accept defeat because of the niche use case I'm presenting.

rnpnr:
I don't think your use case is niche. I also disable syntax highlighting sometimes for various reasons and it is annoying having it re-enabled when I save the file. b42b82d also breaks diff highlighting in git `COMMIT_EDITMSG` files after writing the file.

I am also thinking of reverting b42b82d for now.

> In respect to the solution, i think it might be reverting [b42b82d](https://github.com/martanne/vis/commit/b42b82d2f57981978a1ee390f4c497958474922c) and introducing a `WIN_OPEN` event on the first save of the newly created window (with `:new`), so that we can check the issued filepath to `:w <filepath>` and set the syntax accordingly to the file extension before the `WIN_OPEN` event.

[...]

PR #1388, opened 2026-08-17
Revert "plugins: filetype: set syntax on FILE_SAVE_POST event"

JustTaken:
This reverts commit b42b82d2f57981978a1ee390f4c497958474922c.

Now syntax highlight is no longer forcibly enabled on file save. Related to #1385.

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -327,8 +327,9 @@ end
 -- utility -> datap -> filename -> extension -> L.detect() -> mime
 -- mime is supposed to be the most correct, but it execs file
 -- Which is slow and semi non-portable, so its left for last
-local function Detect(file)
+local function Detect(win)
 	local R
+	local file = win.file
 
 	-- pass first few bytes of file to custom file type detector functions
 	local data = file:content(0, 256)
@@ -420,7 +421,7 @@ local function Detect(file)
 end
 
 vis.events.subscribe(vis.events.WIN_OPEN, function(win)
-	local syntax = Detect(win.file) -- syntax/lexer/filetype
+	local syntax = Detect(win) -- syntax/lexer/filetype
 	local filetype = rawget(filetypes, syntax) -- avoid backwards compatibility mt
 	if filetype then
 		for _, Action in ipairs(filetype.actions or {}) do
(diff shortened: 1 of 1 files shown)


lua/plugins/filetype.lua
```diff
@@ -327,8 +327,9 @@ end
 -- utility -> datap -> filename -> extension -> L.detect() -> mime
 -- mime is supposed to be the most correct, but it execs file
 -- Which is slow and semi non-portable, so its left for last
-local function Detect(file)
+local function Detect(win)
 	local R
+	local file = win.file
```
Nomarian, line 332: Unnecessary, Detect does not need to know the window, only the file, so creating an extra variable for detect is redundant

lua/plugins/filetype.lua
```diff
@@ -420,7 +421,7 @@ local function Detect(file)
 end
 
 vis.events.subscribe(vis.events.WIN_OPEN, function(win)
-	local syntax = Detect(win.file) -- syntax/lexer/filetype
+	local syntax = Detect(win) -- syntax/lexer/filetype
```
Nomarian, line 424: unnecessary.

Nomarian:
It seems you're just erasing the event, which is fine with me but that's still going to undo the previous solution which might make some people mad at, at least its good for your use case.


The good thing is lua allows you to use your version of filetype.lua in $HOME/.config/vis/plugins/filetype.lua

JustTaken:
This pull request was intended to be a pure revert, but i can reproduce the api change in `Detect` if wanted.

> The good thing is lua allows you to use your version of filetype.lua in $HOME/.config/vis/plugins/filetype.lua

Yes, i can do that indeed, the problem is that this event subscription in mainstream affect everyone that is trying to block syntax highlight on any given window. The motivation of the first commit that introduced the `FILE_SAVE_POST` event subscription was for newly created windows (`<C-w>n`) which does not have any filetype and therefore has no file extension. I don't know if there is a deeper reason to reset syntax highlight on save.

I guess the detection of the first save of newly spawned windows can happen in core, even because we have to change the status line indication of the filepath issued with `:w <filepath>` on these new windows, alongside that we can check the file extension and ask for syntax highlight based on the user configuration (i. e. emitting WIN_OPEN event on the first save?).

You can still argue that syntax highlighting is enabled by default and users who don't want it should explicitly disable them, both on window creation and file save. That is a valid point, i was just confused on where it happened and figured other people would wanted the option to disable it without the trouble to trace down where to change to make it work. Because, to me, it feels more natural to detect file types when the file is first opened or when its extension is first provided.

JustTaken:
already fixed #1390

PR #1390, opened 2026-08-18
no syntax highlighting option

Nomarian:
Changes:

Environmental Variables:
 Identifies `$NO_COLOR` and `$VIS_NOCOLOR`

Adds `set syntax_highlight BOOLEAN`

Changes functionality of syntax option

`set syntax on` -- sets syntax on of the window
`set syntax off` -- sets syntax off of the window
`set syntax X` -- changes syntax of current window to X

FILE_SAVE_POST only triggers once

### DIFF
--- a/lua/plugins/filetype.lua
+++ b/lua/plugins/filetype.lua
@@ -322,6 +322,7 @@ local function GetHashBang(data)
 	return hb, utility
 end
 
+M.default_syntax = "text" -- should never be nil
 
 -- Returns syntax/filetype
 -- utility -> datap -> filename -> extension -> L.detect() -> mime
@@ -365,7 +366,7 @@ local function Detect(file)
 		end
 	end
 
-	local path = file.name -- filepath
+	local path = file.path or file.name -- filepath
 	local name
 	if path and path~="" then
 		name = path and path:match("[^/]+$") -- filename
@@ -416,12 +417,30 @@ local function Detect(file)
 		end
 	end
 
-	return 'text'
-end
+	return M.default_syntax
+end M.Detect = Detect
 
-vis.events.subscribe(vis.events.WIN_OPEN, function(win)
-	local syntax = Detect(win.file) -- syntax/lexer/filetype
-	local filetype = rawget(filetypes, syntax) -- avoid backwards compatibility mt
+local FindSyntax
+= function --> syntax: string --
+(
+	win -- vis.win|nil
+)
+	if win then
+		local syntax = win.file.path==nil -- empty window
+			and M.default_syntax -- should be "text"
+			or Detect(win.file) -- syntax/lexer/filetype
+		return syntax
+	end
+	return M.default_syntax
+end
+M.FindSyntax = FindSyntax
+
+local SetActions
+= function --> syntax: string
+(
+	win -- vis.win
+)
+	local filetype = rawget(filetypes, win.SYNTAX) -- avoid backwards compatibility mt
 	if filetype then
 		for _, Action in ipairs(filetype.actions or {}) do
 			Action(win, filetype)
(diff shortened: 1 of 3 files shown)


Nomarian:
Wiped the FILE_SAVE_POST event...

A couple of thing need to be done, :w sets it as default, but the file itself does not have a name, so the status bar does not detect it, this detects it, overriding the name and using path, but there's still needs to be an event for it.

But for now, this is a big enough PR

Also why can't you set win.options.key? it just disappears, you can set win.key but options is a no no.

Nomarian:
This also adds the feature of #1311

Nomarian:
@rnpnr This is ready for merging, functionality didn't change either so nothing should break... I hope

lua/plugins/filetype.lua
```diff
@@ -416,42 +417,82 @@ local function Detect(file)
 		end
 	end
 
-	return 'text'
+	return M.default_syntax
+end M.Detect = Detect
```
rnpnr, line 421: I don't really get the point of this over `M.Detect = function(...) ... end`. Is there a reason?

lua/vis-std.lua
```diff
@@ -54,14 +54,53 @@ vis:option_register("theme", "string", function(name)
...
 end, "Color theme to use, filename without extension")
 
-vis:option_register("syntax", "string", function(name)
-	if not vis.win then return false end
-	if not vis.win:set_syntax(name) then
-		vis:info(string.format("Unknown syntax definition: `%s'", name))
-		return false
+
+-- syntax_highlight is on by default, unless $NO_COLOR is set
+vis.options.SYNTAX_HIGHLIGHT = not(os.getenv"NO_COLOR" or os.getenv"VIS_NOCOLOR")
```
rnpnr: Please just set the default to true. I don't want any more vis or platform specific environment variable usage.

lua/vis-std.lua
```diff
@@ -54,14 +54,53 @@ vis:option_register("theme", "string", function(name)
...
+-- NOTE: .options.SYNTAX_HIGHLIGHT is used as the real option because modifying
+-- .options.syntax_highlight leads to a C stack overflow, possible BUG?
```
rnpnr, line 61: This needs to be investigated before this can be merged. We don't want extra and inconsistent public API just to workaround an unexplored issue.

lua/plugins/filetype.lua
```diff
@@ -416,42 +417,82 @@ local function Detect(file)
...
+
+local FindSyntax
+= function --> syntax: string --
+(
+	win -- vis.win|nil
+)
```
rnpnr, line 427: This is really ugly and difficult to read. If you need to include these put it in a comment above the function declaration.

rnpnr (changes requested):
Hi thank you for working on this! Overall I think the idea is sound but there are few things that should be changed.

Also can you address the case of the user manually specifying a syntax and then that getting overwritten in `FILE_SAVE_POST`. Basically if the user ever enters `:set syntax off` (doesn't have to be `off`), `FILE_SAVE_POST` should not set the syntax again.

PS: sorry to be a little bit more nitpicky about style then usual but I don't anything near the horrendous style I have to look at every day at work making its way into vis.

lua/plugins/filetype.lua
Nomarian: So when localizing variables, the linter I use (luacheck) picks it up in the file, so if I make a typo like DEtect(), it will warn me there is no such thing as DEtect(), but it wont pick up typos inside tables like M.DEtect, probably because tables could be a metatable so they could return anything.


This has a drawback, both a bug and also a feature depending on how you see it, if anybody decides to replace vis.ftdetect.Detect = theirfunction then all the other functions that use localized Detect will still use it and not the other version.


it also saves a GETTABLE call which is... honestly just a non noticeable 1 ns speedup.

lua/vis-std.lua
Nomarian: Right, no VIS_NOCOLOR, I do think supporting [NO_COLOR](https://no-color.org/) is good, for the colorblind at least, but I'm not colorblind, and I never disable syntax highlighting anyway.


It's also a good workaround for now for the lexer recursivity limit, so vis doesn't crash.

lua/plugins/filetype.lua
Nomarian: How should it look? I want it to keep parameter and return type information

```
-- (win: window|nil): syntax: string
local FindSyntax = function(win)
```

I should probably switch to ldoc, but I'll have to read the docs.

lua/vis-std.lua
Nomarian: its not just vis.options but win.options as well, win.options doesn't even exist sometimes.

PR #1397, opened 2026-08-28
vis: add smartcase option

amrfti:
Add smartcase option, mimicing the option of the same name in vim. smartcase implies ignorecase, except that it disables itself when there are uppercase characters are in the search pattern.


amrfti:
Uff, I didn't look hard enough to see someone's already done this.
https://repo.or.cz/vis/gkirilov.git/patch/patch-search-options

amrfti:
On second, thought the above patch is much more general than mine, I'll leave this PR open at the discretion of maintainers.

